PR #12, opened 2021-11-30
Finalize v1.0 with @component tag

FredKSchott:
- Start Date: 11/29/2021
- Status: Abandoned

## Summary

Abandoned.

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/FredKSchott-patch-1/active-rfcs/0000-component-tag.md)


jonathantneal:
I like the concept. I am naming TBD on `@component`.

I think there is also a use-case for hoisting just one global `<style>` from a component. The workaround being `:global`.

FredKSchott:
Updated with feedback from RFC call, mainly:

1. Added a callout for Alpine.js `@event` support, which would not overlap in practice but may be overlap enough in concept that it would be annoying for those users. Looking for feedback from any current Alpine.js + Astro users.
1. Relaxed the "top-level only" rules so that this could work inside of pages and layouts, where nesting inside of `<head>` and `<body>` would be common.
1. Added more detailed description of how `<style>`, `<style @component>`, `<script>`, and `<script @component>` would all work.

FredKSchott:
Updated to remove conflict with #26 RFC proposed behavior around `head/html/body`

jonathantneal:
I searched the page for `media` and saw 0 results, so just adding these use-cases here.

```html
<link rel="stylesheet" href media="list-of-media-queries" />
```

```html
<style media="list-of-media-queries"></style>
```

```css
@import url list-of-media-queries;
```

jonathantneal:
If '_`@import` will be resolved using Vite's `@import` resolve logic_', and Vite uses `exports` then how does one author an export for CSS?

```json
{
  "exports": {
    ".": {
      "import": "./dist/open-props.module.js",
      "style": "./open-props.min.css"
    }
  }
}
```

Or is that out of scope?

FredKSchott:
I'm not sure I follow, but the idea would be that we would support whatever Vite supports. So if you do `@import 'open-props';` I believe Vite resolves that correctly to `open-props/open-props.min.css`. 

The goal here would be to match what would happen in any other CSS file that Vite is bundling today. For example, if you `import './foo.css'` in a React component and then `foo.css` has an `@import` inside of it, Vite should already be doing some resolving and bundling. This proposal says that we would match that behavior in `<style @component>`

jonathantneal:
> So if you do `@import 'open-props';` I believe Vite resolves that correctly to `open-props/open-props.min.css`. 

Literally the `style` export in my hypothetical example? Or did you test this another way with the `open-props` package?

It did not work for me when I tested it in the latest Astro, because Vite uses `exports`, not `main`, `module`, `style`, etc. <sup>[reference](https://unpkg.com/browse/open-props@1.0.2/package.json#L22)</sup>

Another workaround export had to be added. <sup>[reference](https://unpkg.com/browse/open-props@1.0.8/package.json#L29)</sup>

FredKSchott:
Ah gotcha, no I just assumed Vite still fell back to `style` if there was no style entrypoint in the `exports` map, but didn't run it myself to confirm (I should have!).


Looking at [that package.json](https://unpkg.com/browse/open-props@1.0.8/package.json) more, and I think the author actually intentionally wants you to do `@import 'open-props/style';` (which works in Vite) and intentionally wants `@import 'open-props';` to fail. It's hard to say for sure with such a minimal README but that seems to be the intention, in which case, Vite is right to fail on `@import 'open-props';`.

But even if we assume this to be incorrect behavior of Vite, this specific case is intentional in this RFC: Using Vite to handle resolution in all cases is proposed as objectively better than using Vite to handle resolution in some cases AND using some custom implementation that we've built ourselves in others.

jonathantneal:
Correct. That was the workaround. The two links are to different versions. I was asking the question because CSS import resolution via exports seemed like undefined behavior. Maybe it’s just something to call out. Vite can resolve, but resolving styles as a _kind_ of export is undefined behavior? I would be very comfortable with something like that.

active-rfcs/0000-component-tag.md
```diff
@@ -0,0 +1,242 @@
...
+- Not available on all types of elements: `<script @component>` and `<style @component>` are the only two supported.
+- Not a new type of directive: `@component` is the only `@` rule that we want to supported.
+- More aggressive than your normal directive: for example, no other attributes are allowed when the `@component` tag is used.
+- Easily statically analyzable by our parser. 
+- Must be easy to statically analyze. Must be top-level or nested inside of a `<head>`/`<body>` element
+  - Compiler error if conditional or nested in any way.
+- Must be static and cannot rely on runtime behavior in any way, so that it can be bundled once.
+  - Compiler error if dynamic in any way (ex: uses `define:vars`).
+
+## `<style>` (no tag)
```
matthewp, line 66: No attr you mean?

active-rfcs/0000-component-tag.md
```diff
@@ -0,0 +1,242 @@
...
+<style>h1 { color: red; }</style>
+
+<!-- OUTPUT: -->
+<style>h1 { color: red; }</style>
+```
+
+- Style content is always unscoped/global.
+- Style content is escaped as CSS but otherwise untouched by Astro. ex: Sass not supported.
+- `@import` will not be resolved or bundled, sent untouched/raw to the browser.
+- Will be output in the final component render(), useful for adding global CSS in `<head>`.
```
matthewp, line 79: Not explicitly said here, but I assume these styles will not be bundled/hoisted, right?

active-rfcs/0000-component-tag.md
```diff
@@ -0,0 +1,242 @@
...
+```astro
+---
+import {Style, Script} from 'astro/components';
+---
+<Style>...</Style>
+<Script>...</Script>
+```
+
+- Pro: No user has an expectation about how this would work. That means less conflict with how `<style>` and `<script>` are expected to work.
+- Con: Requires extra import boilerplate in every component. We could make these globally available to mitigate, but we've never done that before.
```
matthewp, line 233: We have one global component, `Fragment`. I think the primary problem with a Component is that it can be renamed (`const MyStyle = Style`) and its hard to warn against doing that.

matthewp:
Not much to add, 👍 overall, would love to bikeshed on the name but that can come later.

active-rfcs/0000-component-tag.md
FredKSchott: thanks! clarified to add this point explicitly

FredKSchott:
> Correct. That was the workaround. The two links are to different versions. I was asking the question because CSS import resolution via exports seemed like undefined behavior. Maybe it’s just something to call out. Vite can resolve, but resolving styles as a kind of export is undefined behavior. I would be very comfortable with something like that. What do you think?

I'd love to continue this convo outside of this RFC, since that is already the behavior of a CSS import in Astro today, and this RFC isn't meaning to change anything about that behavior (just clarifying how resolved CSS imports work today to the reader).

tony-sull:
The distinction between layouts and components really feels like an important one here with regards to the default behavior

I could see it making sense to say layouts would default to not scoping, for example, but components really probably want scoping by default.  In either case, default behavior for preprocessing like SASS may be a project-wide default where if I use SASS I'd expect both layouts and components to be run through SASS

If we end up with special behavior for the `src/layouts` directory, would it make sense to assume scoping by default in `src/pages` and `src/components`?

tony-sull:
One other thought here, the idea of an all-or-nothing approach to the style/script processing magic could be either a bit limiting or cumbersome if we end up with a long list of custom attributes to disable each step of the magic

i.e. the RFC allows for `<style @component is:global lang='scss'>` to get global, bundled, SASS-compiled styles.  That's already a bit of a heavy lift as far as DX goes with so many non-standard attributes there, and we could end up needing more down the road if more style processing magic is added to the `@component` functionality

I know the idea of something similar to `svelte-preprocess` came up before and was punted, but maybe we'd be better off here including the current default preprocessing behavior with as few attribute flags as possible and know that down the road this behavior would likely just become a default config that could be overridden with something along the lines of `astro-preprocess`?

jasikpark:
Thoughts about https://github.com/withastro/rfcs/pull/12#

- I want to have default scoping for components b/c I don't want their styles to leak into other components unless intentionally, it's rare that I would be using `<style>` inside a component w/o `@component` added in this RFC
- I've generally used styles as scoped generally + <link rel="stylesheet" src={Astro.resolve("../theme.css")}/> or via `import "@theme/theme.css"` when I wanted global css to be included in my HeadCommon.astro component, ex: [HeadCommon.astro](https://github.com/jasikpark/unofficial-distractible-scoreboard/blob/dd3bbceaa7750bdadb2e8e43200154bc694bea15/src/components/HeadCommon.astro#L2)
- Would it be possible to add a `please:don't:preprocess:me` directive to style and script tags that enables this no-magic functionality? So you add an attribute to these elements and they become fully static?
- A use case that I would not want to lose is scoped styles + define:vars, which it sounds like this RFC removes w/ scoped styles being behind `@component` + `define:vars` only supported via static markup. Is this not a use case that's supported today?
- Would it be possible to create this RFC in a way that isn't a binary on-off of no processing to all the processing but rather having a default of no processing + incrementally adding processing as you add directives? I suppose that makes `@import` resolving ambiguous about whether it's resolved by Vite or the browser, but that could be solved by something like `hoist` to opt-in to bundling of the css.
- Is it a requirement that all attributes are dropped from a tag with `@component`? Could those attributes not be copied over to the browser?

matthewp:
@jasikpark 

> A use case that I would not want to lose is scoped styles + define:vars, which it sounds like this RFC removes w/ scoped styles being behind @component + define:vars only supported via static markup. Is this not a use case that's supported today?

This can't be supported in SSR due to the fact that it makes those style tags be runtime dynamic. So we could never optimize these styles and bundle them.

If we had a lighter weight way to do scoping (rather than running PostCSS through Vite) so that it *could* be done at runtime then perhaps define:vars + scoping could be done together. For that I would propose an opt-in way to do scoping (`astro:scope` or some such).

FredKSchott:
Hey all, thanks again for the feedback! I'm going to go ahead and split this into two RFCs which should hopefully make it easier to discuss. I'll make sure all feedback here is carried into those PRs so that nothing is lost.

FredKSchott:
New RFC, focused only on behavior: https://github.com/withastro/rfcs/pull/41

PR #26, opened 2021-12-01
Finalize <head>, <body>, <html>

FredKSchott:
- Start Date: 2021-11-30
- Status: Draft

## Summary

Finalize behavior around `<head>`, `<body>`, `<html>`, and `<!DOCTYPE html>` elements in Astro component templates.

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/FredKSchott-patch-2/active-rfcs/0000-finalize-head-body.md)

### DIFF
--- a/active-rfcs/0000-finalize-head-body.md
+++ b/active-rfcs/0000-finalize-head-body.md
@@ -9,8 +9,11 @@ Finalize behavior around `<head>`, `<body>`, `<html>`, and `<!DOCTYPE html>` ele
 # Motivation
 
 in Astro v0.21, we have had user reports around `<head>` not acting as expected, breaking some projects from v0.20:
+- **Summary:** https://github.com/withastro/astro/issues/2128
 - https://github.com/withastro/astro/issues/2046
 - https://github.com/withastro/astro/issues/2022
+- https://github.com/withastro/astro/issues/2132
+- https://github.com/withastro/astro/issues/2151
 
 Some of these issues stem from undocumented or undefined behaviors in v0.20: was it allowed for `<head>` be conditionally added in a nested expression? 
 
--- a/active-rfcs/0000-finalize-head-body.md
+++ b/active-rfcs/0000-finalize-head-body.md
@@ -15,7 +15,7 @@ in Astro v0.21, we have had user reports around `<head>` not acting as expected,
 Some of these issues stem from undocumented or undefined behaviors in v0.20: was it allowed for `<head>` be conditionally added in a nested expression? 
 
 Other issues stem from our attempt to provide implicit `<head>` and `<body>` support in pages and layouts. There was a quick workaround added for v0.21 where `src/pages` and `src/layouts` folder locations must be defined in config so that our parser knows how to parse them as full HTML documents and not smaller HTML fragments. This unblocked users but at the expense of breaking two larger design goals of the project:
-  - Added `src/layouts` as a new special folder location, when only `src/pages` was intented to be "special".
+  - Added `src/layouts` as a new special folder location, when only `src/pages` was intended to be "special".
   - Caused Astro to create different output for a component based on the file path of that component.
 
 ## Goals
--- a/active-rfcs/0000-finalize-head-body.md
+++ b/active-rfcs/0000-finalize-head-body.md
@@ -8,7 +8,13 @@ Finalize behavior around `<head>`, `<body>`, `<html>`, and `<!DOCTYPE html>` ele
 
 # Motivation
 
-@jonathantneal and I were chatting about [RFC: Finalize v1.0 style and script API for components](https://github.com/withastro/rfcs/pull/12) and discussing `<head>` behavior & bugs in Astro. We came to the realization that only supporting `<head>` in your top-level pages (and not in components) doesn't make sense for anyone using [layout components](https://docs.astro.build/core-concepts/layouts/), which can act as a component but often defines the entire `<head>`. This is a very common use-case for Astro users.
+in Astro v0.21, we have had user reports around `<head>` not acting as expected, breaking some projects from v0.20:
+- https://github.com/withastro/astro/issues/2046
+- https://github.com/withastro/astro/issues/2022
+
+Some of these issues stem from undocumented or undefined behaviors in v0.20, like the fact that `<head>` could be conditional.
+
+This RFC is an attempt to agree on and document the behavior of the `<head>`, `<body>`, and `<html>` tags in a way that makes sense, and can be communicated to users. It should support most common use-cases, such as layouts:
 
 ```astro
 ---
@@ -19,22 +25,33 @@ import CommonLayout from '../layouts/CommonLayout.astro';
 <CommonLayout someProp="..." />
 ```
 
-This was something that we had supported in v0.20, which was accidentally removed as a part of the new compiler. Adding back this behavior could be considered a regression bug fix and not necessarily requiring an RFC (and I think Nate is already working on a fix for this to add back previous behavior), but this feels like a good time to commit to a larger plan around `<head>` and other similar HTML elements.
+## Goals
 
-This same `<head>` problem applies to the `<body>` element, and may also apply to the `<html>` and `<!DOCTYPE html>` elements.
+- Agree on and document the rules and behaviors of the `<head>`, `<body>`, and `<html>` tags.
+- Support any component in the tree adding to the page `<head>`, similar to `<svelte:head>` or head manager libraries in React/Vue. 
+  - (Note: see "Alternatives" for an alternative design if this is decided to be a non-goal).
+
+## Non-Goals
+
+- **Intentionally not implemented:** The ability or option to omit `<body>`, `<head>`, or `<html>` in the final output. This is intentionally not supported here in an effort to reduce complexity around when these tags might or might not be output in your final page HTML. As an Astro user, we assume that you are are comfortable with Astro controlling/optimizing your final HTML output.
+  - If this is sticking point, we can reword this from "intentionally not supported" to "out of scope / possible to implement in the future", but I think it's useful to commit to this behavior as a team if there is no objection.
+- **Out of scope / future:** The ability to author without `<body>`, `<head>`, or `<html>`, and relying on the compiler to handle them like the browser would.  This makes some sense because this is valid in HTML and we try to match HTML semantics whenever possible.  This is currently treated as out-of-scope, but there is nothing here that blocks our ability to do this in the future.
+  - For example, our compiler could automatically detect some elements in the component template (for example, a `<title>`) and auto-hoist it out of the component body and into the `<head>` of the final compiler result to support this.
+- **Out of scope / future:** Intelligent merging of repeated head elements. This is common when a top-level layout `<head>` might define a title, and then a more specific component might overwrite with a more specific `<title>` of its own. This is considered out of scope but not blocked to be implemented in the future.
 
 # Detailed design
 
 - `<head>`
   - A `<head>` element can appear in any component, including pages, layouts, and simple UI components.
+  - The contents of a `<head>` element can be dynamic.
   - Astro will collapse (flatten) all `<head>` elements together from the top-level page to the lowest component into the final page `<head>`.
-  - `<head>` elements are collected and printed in breadth-first order.
-  - If no `<head>` elements are found in the final output, none will be output to the page.
+  - `<head>` elements are collected and and printed in the order that the components appear on the page.
+  - If no `<head>` elements are used by the author, no head will be output to the page.
   - This solves for the page layout problem AND gives us new head management support, similar to [`<svelte:head>`](https://svelte.dev/docs#svelte_head)
 - `<body>`
   - completely ignored in a component template
   - a `<body>` tag is always included in the final output
-  - everything in the component template is assumed to be the body
+  - everything in the component template is assumed to be the body, other than the other elements listed here
 - `<html>`
   - completely ignored in a component template
   - an `<html>` tag is always included in the final output
@@ -42,19 +59,27 @@ This same `<head>` problem applies to the `<body>` element, and may also apply t
   - completely ignored in a component template
   - a `<!DOCTYPE html>` tag is always included in the final output
 
+## Implementation
+
+Similar to how the Svelte compiler works, `head` elements would be collected in a `result.head` property by the compiler, seperate from the body of the template. 
+
+The runtime would collect these across all rendered components on the page, and at the end of the page render would have the final collapsed `head` to print to the page. 
+
 # Drawbacks
 
 - `<head>` ordering - I'm not sure if this is easy to implement. Could use a prototype or expert set of eyes :)
-- There's a small amount of interest in supporting developers who do not want their output to include `<body>`, `<head>`, or `<html>` in the final output. This is intentionally not supported by this proposal in an effort to reduce complexity. This is understood to be mostly for vanity (the bytes saved are minor) and as an Astro user you are already comfortable with Astro controlling/optimizing your final output. 
-  - If this is sticking point, we can reword this from "intentionally not supported" to "out of scope / possible to implement in the future", but I think it's important to commit to this behavior as a team if there is no objection.
-- There's a slightly larger amount of interest in supporting developers who do not want to their output to include `<body>`, `<head>`, or `<html>` in the templates, and instead relying on the compiler to handle this for them.  This makes some sense because this is valid in HTML and we try to match HTML semantics whenever possible.  This is currently treated as out-of-scope, but there is nothing here that blocks our ability to do this in the future. 
-  - For example, our compiler could automatically detect some elements in the component template (for example, a `<title>`) and auto-hoist it out of the component body and into the `<head>` of the final compiler result to support this.
+- Does not support Streaming HTML for SSR. See "No `<head>` Collapsing" for an alternative approach that addresses this drawback. 
 
 # Alternatives
 
-- Do we warn if `<body>` _NEVER_ appears in the page's rendered component tree? appears twice?
-  - @jonathantneal made the good point that `<body>` isn't actually required for valid HTML
-- Do we warn if `<html>` _NEVER_ appears in the page's rendered component tree? appears twice?
+## No `<head>` Collapsing
+
+It is a goal to allow all components to add elements directly to `<head>`. If we remove this goal, then then we could consider a design that only allows one component to act as "page layout" and set the `<html>`, `<head>` and `<body>` tags and then all other components are just fragments, always assumed to be a part of the body. There seems to be interest in this goal from the community (see linked discussion for more) but this could still be up for debate if there is interest.
+
+One reason to consider this is that it would make naive HTML streaming easier once we move to SSR. We could respond with the first component's `<head>` contents as soon as that one component renders, and then we would stream down its body as it rendered. As written, the current proposal does not support this since the head could be modified at any point in the document.
+
+This may be much easier to implement in the current compiler, which would be a factor to consider. Looking for feedback on this.
+
 
 # Adoption strategy
 
@@ -62,4 +87,7 @@ This same `<head>` problem applies to the `<body>` element, and may also apply t
 
 # Unresolved questions
 
-- None yet.
+
+- Do we warn if `<body>` _NEVER_ appears in the page's rendered component tree? appears twice?
+  - @jonathantneal made the good point that `<body>` isn't actually required for valid HTML
+- Do we warn if `<html>` _NEVER_ appears in the page's rendered component tree? appears twice?
--- a/active-rfcs/0000-finalize-head-body.md
+++ b/active-rfcs/0000-finalize-head-body.md
@@ -58,7 +58,7 @@ In this design it is more "on you" to write valid HTML. This comes from the real
 
 # Alternatives
 
-- Forcing that a user includes `<html>`, `<body>` and `<head>` elements in the final page output. This could be explored as an optional lint rule, however there is not currently consensus around this idea and also there is no known objective value in this from a compiler or runtime standpoint. If this is desired, it should be considered a seperate feature request and out of scope from this RFC.
+- Forcing that a user includes `<html>`, `<body>` and `<head>` elements in the final page output. This could be explored more, however there is not currently consensus around this idea and also there is no known reason to tackle this as a part of this RFC. If this is desired, it should be considered a seperate feature request and out of scope from this RFC.
 
 # Adoption strategy
 
--- a/active-rfcs/0000-finalize-head-body.md
+++ b/active-rfcs/0000-finalize-head-body.md
@@ -54,7 +54,7 @@ In this design it is more "on you" to write valid HTML. This comes from the real
 # Drawbacks
 
 - This RFC will allow you to author a page of HTML that does not actually output `<html>`, `<body>` or `<head>` elements. This can be considered an advantage in that it is more flexible than something like Svelte or Vue. However, it means we need to be more diligent with testing this kind of output across our codebase and dependencies. For example, we would need to confirm that Vite does not have trouble injecting things into an HTML document without a head or body.
-  - If we are blocked by this limitation in some way and can't reasonably unblock (for example, a non-trivial Vite limitaiton) this RFC states that it would be acceptable for Astro to add `<head>` or `<body>` elements in a way that keeps the output HTML valid but unblocks Astro.
+  - If we are blocked by this limitation in some way and can't reasonably unblock (for example, a non-trivial Vite limitaiton) this RFC states that it would be acceptable for Astro to add or require a `<head>` or `<body>` element in the final output HTML in a way that keeps the output HTML valid but unblocks Astro/tooling. Reminder that it is not an explicit goal to support authoring with `<head>` or `<body>`, but it is something that we would like to support if we reasonably can.
 
 # Alternatives
 
--- a/active-rfcs/0000-finalize-head-body.md
+++ b/active-rfcs/0000-finalize-head-body.md
@@ -36,11 +36,13 @@ Other issues stem from our attempt to provide implicit `<head>` and `<body>` sup
 - `<!DOCTYPE html>`
   - compiler: completely ignored by the compiler, stripped out from render output.
   - runtime: a `<!DOCTYPE html>` tag is always included in final page HTML output.
-- `<html>`, `<body>`, `<head>`
+- `<html>`, `<body>`
   - compiler: will be left as-is in the component template.
   - runtime: will be left as-is in final page HTML output.
-  - runtime: will warn if no `head` tag is rendered by an Astro component for an entire page.
-  - runtime: may warn if duplicate `html`, `body`, and `head` tags are rendered inside of Astro components for an entire page.
+- `<head>`
+  - runtime: element is always required in final page output for style/script injection.
+  - runtime: will warn if `head` tag is not rendered exactly once for an entire page.
+  - `<head>` not supported inside of a non-Astro components. Non-Astro components could generate `<head>` contents, like `<link>` or `<meta>`, but not the `<head>` element itself.
 
 ## Compiler Changes
 
@@ -53,7 +55,6 @@ Losing `"as": "document"` parse mode will remove some special parser handling, l
 In this design it is more "on you" to write valid HTML. This comes from the reality that an imported component can contain unknown HTML, so the compiler can't implicitly assume anything about what is or is not included included the final parent component template.  See https://github.com/withastro/astro/issues/2022 for examples of when this breaks down today. We can help with some static linting, and runtime warnings if the final HTML output is invalid. However, this RFC acknowledges the reality that already exists in v0.21 that imported components break any assumptions and help that we previously attempted to provide.
 
 ## Head Injection Changes
-
 - runtime: will remove current post-build head injection, which involves a fragile `'</head>'` string find-and-replace.
 - compiler: will add a new `head: string` (or: `injectHead(): string`) property to the compiler transform options, which will inject the given HTML string into the bottom of a `<head>` element, if one is return by the compiler.
 - runtime: will provide this value head injection value to the compiler, and throw an exception if not used/called exactly once during a page render. 
--- a/active-rfcs/0000-finalize-head-body.md
+++ b/active-rfcs/0000-finalize-head-body.md
@@ -12,79 +12,69 @@ in Astro v0.21, we have had user reports around `<head>` not acting as expected,
 - https://github.com/withastro/astro/issues/2046
 - https://github.com/withastro/astro/issues/2022
 
-Some of these issues stem from undocumented or undefined behaviors in v0.20, like the fact that `<head>` could be conditional.
+Some of these issues stem from undocumented or undefined behaviors in v0.20: was it allowed for `<head>` be conditionally added in a nested expression? 
 
-This RFC is an attempt to agree on and document the behavior of the `<head>`, `<body>`, and `<html>` tags in a way that makes sense, and can be communicated to users. It should support most common use-cases, such as layouts:
-
-```astro
----
-// src/pages/index.astro
-import CommonLayout from '../layouts/CommonLayout.astro';
----
-<!-- this is valid, where CommonLayout contains `<head>`, `<body>`, etc.
-<CommonLayout someProp="..." />
-```
+Other issues stem from our attempt to provide implicit `<head>` and `<body>` support in pages and layouts. There was a quick workaround added for v0.21 where `src/pages` and `src/layouts` folder locations must be defined in config so that our parser knows how to parse them as full HTML documents and not smaller HTML fragments. This unblocked users but at the expense of breaking two larger design goals of the project:
+  - Added `src/layouts` as a new special folder location, when only `src/pages` was intented to be "special".
+  - Caused Astrp to create different output for a component based on the file path of that component.
 
 ## Goals
 
-1. Agree on and document the rules and behaviors of the `<head>`, `<body>`, and `<html>` tags.
-2. Support any component in the tree adding to the page `<head>`, similar to `<svelte:head>` or head manager libraries in React/Vue. (Note: see "Alternatives" for an alternative design if this is decided to be a non-goal).
+1. Agree on and document the rules and behaviors of the `<head>`, `<body>`, `<html>`, and `<!DOCTYPE html>` tags.
+2. Remove special compiler behavior based on where a file lives in your src directory. 
+3. Remove `src/layouts` as a "special" folder and any config around this.
 
 ## Non-Goals
 
-1. **Intentionally not implemented:** The ability or option to omit `<body>`, `<head>`, or `<html>` in the final output. This is intentionally not supported here in an effort to reduce complexity around when these tags might or might not be output in your final page HTML. As an Astro user, we assume that you are are comfortable with Astro controlling/optimizing your final HTML output.  
-  • If this is sticking point, we can reword this from "intentionally not supported" to "out of scope / possible to implement in the future", but I think it's useful to commit to this behavior as a team if there is no objection.
-1. **Out of scope / future:** The ability to author without `<body>`, `<head>`, or `<html>`, and relying on the compiler to handle them like the browser would.  This makes some sense because this is valid in HTML and we try to match HTML semantics whenever possible.  This is currently treated as out-of-scope, but there is nothing here that blocks our ability to do this in the future.  
-  • For example, our compiler could automatically detect some elements in the component template (for example, a `<title>`) and auto-hoist it out of the component body and into the `<head>` of the final compiler result to support this.
-1. **Out of scope / future:** Intelligent merging of repeated head elements. This is common when a top-level layout `<head>` might define a title, and then a more specific component might overwrite with a more specific `<title>` of its own. This is considered out of scope but not blocked to be implemented in the future.
+1. **Out of scope:** Support `<head>` Injection inside of individual components. This is a commonly requested user feature, but can be tackled as an additional feature, leaving this RFC focused on clarifying our existing API. This RFC does not impact that feature request.
+
 
 # Detailed design
 
-- `<head>`
-  - A `<head>` element can appear in any component, including pages, layouts, and simple UI components.
-  - The contents of a `<head>` element can be dynamic.
-  - Astro will collapse (flatten) all `<head>` elements together from the top-level page to the lowest component into the final page `<head>`.
-  - `<head>` elements are collected and and printed in the order that the components appear on the page.
-  - If no `<head>` elements are used by the author, no head will be output to the page.
-  - This solves for the page layout problem AND gives us new head management support, similar to [`<svelte:head>`](https://svelte.dev/docs#svelte_head)
-- `<body>`
-  - completely ignored in a component template
-  - a `<body>` tag is always included in the final output
-  - everything in the component template is assumed to be the body, other than the other elements listed here
-- `<html>`
-  - completely ignored in a component template
-  - an `<html>` tag is always included in the final output
+## Template Changes
+
 - `<!DOCTYPE html>`
-  - completely ignored in a component template
-  - a `<!DOCTYPE html>` tag is always included in the final output
+  - compiler: completely ignored by the compiler, stripped out from render output.
+  - runtime: a `<!DOCTYPE html>` tag is always included in final page HTML output.
+- `<html>`, `<body>`, `<head>`
+  - compiler: will be left as-is in the component template.
+  - runtime: will be left as-is in final page HTML output.
+  - runtime: may warn if duplicate `html`, `body`, and `head` tags are included in final page HTML output.
 
-## Implementation
+## Compiler Changes
 
-Similar to how the Svelte compiler works, `head` elements would be collected in a `result.head` property by the compiler, seperate from the body of the template. 
+- `src/pages` and `src/layouts` will no longer be parsed or rendered differently from other components
+- `as?: 'document' | 'fragment';` will be removed from the compiler.
+- All Astro components will now be compiled as fragments.
 
-The runtime would collect these across all rendered components on the page, and at the end of the page render would have the final collapsed `head` to print to the page. 
+Losing `"as": "document"` parse mode will remove some special parser handling, like getting an implicit `<head>` to wrap a standalone `<title>` element. This is intentional to bring us more inline with the RFC, where we respect the HTML authored by the developer as the source of truth and leave mostly as-is in the final output.
+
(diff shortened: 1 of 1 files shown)


active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,65 @@
...
+- `<html>`
+  - completely ignored in a component template
+  - an `<html>` tag is always included in the final output
+- `<!DOCTYPE html>`
+  - completely ignored in a component template
+  - a `<!DOCTYPE html>` tag is always included in the final output
+
+# Drawbacks
+
+- `<head>` ordering - I'm not sure if this is easy to implement. Could use a prototype or expert set of eyes :)
```
FredKSchott, line 47: @natemoo-re would love your feedback on how hard this will be to implement, and whether changes are needed to the compiler or only the runtime

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,65 @@
...
+- There's a small amount of interest in supporting developers who do not want their output to include `<body>`, `<head>`, or `<html>` in the final output. This is intentionally not supported by this proposal in an effort to reduce complexity. This is understood to be mostly for vanity (the bytes saved are minor) and as an Astro user you are already comfortable with Astro controlling/optimizing your final output.
```
jonathantneal, line 48:
```suggestion
```

I can’t speak for whoever brought up saving bytes, but I don’t think this bit is necessary. It’s hard to say what is 'vanity'. One could say _adding_ tags is vanity.

> ... There's a small amount of interest in supporting developers who do not want their output to include `<body>`, `<head>`, or `<html>` in the final output. This is intentionally not supported by this proposal ...

This bit also seems inconsistent.

> If no `<head>` elements are found in the final output, none will be output to the page.

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,65 @@
...
+- `<head>`
+  - A `<head>` element can appear in any component, including pages, layouts, and simple UI components.
+  - Astro will collapse (flatten) all `<head>` elements together from the top-level page to the lowest component into the final page `<head>`.
+  - `<head>` elements are collected and printed in breadth-first order.
+  - If no `<head>` elements are found in the final output, none will be output to the page.
+  - This solves for the page layout problem AND gives us new head management support, similar to [`<svelte:head>`](https://svelte.dev/docs#svelte_head)
+- `<body>`
+  - completely ignored in a component template
+  - a `<body>` tag is always included in the final output
+  - everything in the component template is assumed to be the body
```
jonathantneal, line 37:
```suggestion
  - If no `<body>` elements are found in the final output, none will be output to the page.
```

I suggest removing this to match the behavior of `<head>` in this proposal, and also what the web allows.

If it ain’t broke, don’t fix it.

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,65 @@
...
+```astro
+---
+// src/pages/index.astro
+import CommonLayout from '../layouts/CommonLayout.astro';
+---
+<!-- this is valid, where CommonLayout contains `<head>`, `<body>`, etc.
+<CommonLayout someProp="..." />
+```
+
+This was something that we had supported in v0.20, which was accidentally removed as a part of the new compiler. Adding back this behavior could be considered a regression bug fix and not necessarily requiring an RFC (and I think Nate is already working on a fix for this to add back previous behavior), but this feels like a good time to commit to a larger plan around `<head>` and other similar HTML elements.
```
matthewp, line 22: I think having these paragraphs makes the rest of the RFC confusing, because as you say here, the motivation for an API change wouldn't actually be because of a regression, would it? 

If so the motivation becomes irrelevant as soon as the regression is fixed. I think this motivation section should contain the movations/use-cases assuming that this bug didn't exist.

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,65 @@
...
+- `<body>`
+  - completely ignored in a component template
+  - a `<body>` tag is always included in the final output
+  - everything in the component template is assumed to be the body
+- `<html>`
+  - completely ignored in a component template
+  - an `<html>` tag is always included in the final output
+- `<!DOCTYPE html>`
+  - completely ignored in a component template
+  - a `<!DOCTYPE html>` tag is always included in the final output
```
matthewp, line 43: I'm fine with this personally, we should search discord just to be sure there aren't people out there wanting to use older doctypes though.

active-rfcs/0000-finalize-head-body.md
FredKSchott: This was meant to give @natemoo-re the freedom to keep working on a smaller fix for `<head>` layout behavior as a short-term workaround for everyone blocked by the `v0.20 -> v0.21` behavior change, without feeling blocked by this RFC.

Agreed that this paragraph can be removed if it's distracting. The larger RFC is attempting to make a behavior change explicit across head, body, and `<html>` so that we can document, test, and then commit to it. This is different from what we currently support in v0.21 today which is much less defined (as far as I can tell).

matthewp:
Couple of notes:

1. This is missing use-cases. There are some known use-case from within SPA frameworks, but are those use-cases relevant in a framework where you have control over the entire page template? A comparison of the approaches would be helpful here.
2. This RFC is missing a Drawbacks section. One that comes to mind is that this probably prevents streaming SSR. If we have to wait for all components to render before knowing what to put into the head, we can't stream out the HTML.
  * This is perhaps not true if there was a static requirement place on the component-level head usage (in which case it could be extracted at compile time), but this RFC doesn't talk about any restrictions.

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,65 @@
...
+This was something that we had supported in v0.20, which was accidentally removed as a part of the new compiler. Adding back this behavior could be considered a regression bug fix and not necessarily requiring an RFC (and I think Nate is already working on a fix for this to add back previous behavior), but this feels like a good time to commit to a larger plan around `<head>` and other similar HTML elements.
+
+This same `<head>` problem applies to the `<body>` element, and may also apply to the `<html>` and `<!DOCTYPE html>` elements.
+
+# Detailed design
+
+- `<head>`
+  - A `<head>` element can appear in any component, including pages, layouts, and simple UI components.
+  - Astro will collapse (flatten) all `<head>` elements together from the top-level page to the lowest component into the final page `<head>`.
+  - `<head>` elements are collected and printed in breadth-first order.
```
jonathantneal, line 31:
```suggestion
  - `<head>` elements are collected and printed in the order they are parsed (token-first).
```

- The compiler works off a stream of tokens.
- IIRC, breadth-first means it would need to build a tree first, which is less performant, and wouldn’t match browser behavior.

active-rfcs/0000-finalize-head-body.md
jonathantneal, line 37:
```suggestion
```

For the layout that includes `<html lang={Astro.props.lang}>`.

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,65 @@
...
+  - If no `<head>` elements are found in the final output, none will be output to the page.
+  - This solves for the page layout problem AND gives us new head management support, similar to [`<svelte:head>`](https://svelte.dev/docs#svelte_head)
+- `<body>`
+  - completely ignored in a component template
+  - a `<body>` tag is always included in the final output
+  - everything in the component template is assumed to be the body
+- `<html>`
+  - completely ignored in a component template
+  - an `<html>` tag is always included in the final output
```
jonathantneal, line 40:
```suggestion
  - If no `<html>` elements are found in the final output, none will be output to the page.
```

Small fragment-like pages may turn out to be very useful (someone might come up with  'hotwire' like html injections, because it already works on the web™, so now Astro can make it easier).

active-rfcs/0000-finalize-head-body.md
jonathantneal, line 47: I defer to @natemoo-re, and chime to say that if the compiler reads as a stream, parsing-order seems doable.

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,65 @@
...
+
+# Alternatives
+
+- Do we warn if `<body>` _NEVER_ appears in the page's rendered component tree? appears twice?
+  - @jonathantneal made the good point that `<body>` isn't actually required for valid HTML
+- Do we warn if `<html>` _NEVER_ appears in the page's rendered component tree? appears twice?
+
+# Adoption strategy
+
+- This proposal is more relaxed than our existing rules, so this shouldn't be a breaking change for any existing users.
```
jonathantneal, line 61: Agreed. To me, it feels like an improvement of v0.20 behavior.

active-rfcs/0000-finalize-head-body.md
jonathantneal, line 43: Generally in favor, because:

1. `<!doctype html>` totally affects how a page is handled by a browser.
2. `<!doctype html>` is not an element. It’s a one-off and the only node of its kind.

jonathantneal (commented):
I can dig it. Left some comments.

FredKSchott:
> @matthewp: **This is missing use-cases. There are some known use-case from within SPA frameworks, but are those use-cases relevant in a framework where you have control over the entire page template? A comparison of the approaches would be helpful here.**

Let me see if I can improve the "motivation" section based on your other item of feedback. This section is meant to answer this question.

> @matthewp: **One drawback that is not listed here is streaming SSR. This probably prevents it. If we have to wait for all components to render before knowing what to put into the head, we can't stream out the HTML.**

I believe that this is already not supported based on how layouts work, but I agree that this is making that lack of support explicit. I'll add a note to the Drawbacks section.

active-rfcs/0000-finalize-head-body.md
FredKSchott: yea, I think I was using this term wrong. Reworded!

active-rfcs/0000-finalize-head-body.md
FredKSchott: Yup, this is a departure from what we'd talked about around not printing a body if no explicit `<body>` is found. In this proposal, we always print a `<body>` element in the final output, so unless the page is completely empty of content, you can assume a `<body>` element will exist in the final output.

See new "non-goals" section for summary of those changes and reasoning for this stance.

FredKSchott: clarified to not include `html` and `head` elements, but otherwise this is correct: the proposal is that everything not inside of `head` is assumed to be a body element and output inside the final output's `<body>`.

active-rfcs/0000-finalize-head-body.md
FredKSchott: ah, this would be a good reason to revisit that non-goal around output format. I'll add this as an open question for now.

active-rfcs/0000-finalize-head-body.md
FredKSchott: fair enough, rewritten to be less opinionated and more clear/consistent.

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,93 @@
...
+The runtime would collect these across all rendered components on the page, and at the end of the page render would have the final collapsed `head` to print to the page. 
+
+# Drawbacks
+
+- `<head>` ordering - I'm not sure if this is easy to implement. Could use a prototype or expert set of eyes :)
+- Does not support Streaming HTML for SSR. See "No `<head>` Collapsing" for an alternative approach that addresses this drawback. 
+
+# Alternatives
+
+## No `<head>` Collapsing
```
FredKSchott, line 75: @matthewp outlined a version of this proposal here that would remove the ability of any component to add elements to the head, in favor of supporting streaming HTML. I actually like the simplicity of this, and if this is easier to implement I might be more in favor of this for v1.0 (with the understanding that we could still add dynamic `<head>` injection later, post-v1.0 if there was enough interest)

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,93 @@
...
+<CommonLayout someProp="..." />
+```
+
+## Goals
+
+- Agree on and document the rules and behaviors of the `<head>`, `<body>`, and `<html>` tags.
+- Support any component in the tree adding to the page `<head>`, similar to `<svelte:head>` or head manager libraries in React/Vue. 
+  - (Note: see "Alternatives" for an alternative design if this is decided to be a non-goal).
+
+## Non-Goals
```
FredKSchott, line 34: @jonathantneal reworded this section to make it more clear that this proposal doesn't support optional `<head>` and `<body>` tags in the output, and why. 

Since some of your comments were rewording parts of the proposal to add support for this, it might be better to discuss any non-goals that you disagree with here first, since those changes didn't make sense without first changing the goals/non-goals here.

active-rfcs/0000-finalize-head-body.md
jonathantneal: I prefer this to the current proposal, because it would be more _true_ to HTML, preventing automatic `<html>`, `<head>` or `<body>` generation from doing the wrong thing.

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,93 @@
...
+- **Out of scope / future:** The ability to author without `<body>`, `<head>`, or `<html>`, and relying on the compiler to handle them like the browser would.  This makes some sense because this is valid in HTML and we try to match HTML semantics whenever possible.  This is currently treated as out-of-scope, but there is nothing here that blocks our ability to do this in the future.
+  - For example, our compiler could automatically detect some elements in the component template (for example, a `<title>`) and auto-hoist it out of the component body and into the `<head>` of the final compiler result to support this.
+- **Out of scope / future:** Intelligent merging of repeated head elements. This is common when a top-level layout `<head>` might define a title, and then a more specific component might overwrite with a more specific `<title>` of its own. This is considered out of scope but not blocked to be implemented in the future.
+
+# Detailed design
+
+- `<head>`
```
matthewp, line 44: It would be nice if these points where numbered, so you could reference them elsewhere, acting as ids ("clause 1.5 says....")

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,93 @@
...
+  - A `<head>` element can appear in any component, including pages, layouts, and simple UI components.
+  - The contents of a `<head>` element can be dynamic.
+  - Astro will collapse (flatten) all `<head>` elements together from the top-level page to the lowest component into the final page `<head>`.
+  - `<head>` elements are collected and and printed in the order that the components appear on the page.
+  - If no `<head>` elements are used by the author, no head will be output to the page.
+  - This solves for the page layout problem AND gives us new head management support, similar to [`<svelte:head>`](https://svelte.dev/docs#svelte_head)
+- `<body>`
+  - completely ignored in a component template
+  - a `<body>` tag is always included in the final output
```
matthewp, line 53: Previously it says:

> If no `<head>` elements are used by the author, no head will be output to the page.

It's confusing to me that a head might be omitted but a body tag will always be included. I think it makes more sense to just always print doctype, html, head, and body tags.

active-rfcs/0000-finalize-head-body.md
FredKSchott: Suprisingly, GitHub's nested numbered list support is... non-existent? Had to hack it a bit but added numbers where I could.

active-rfcs/0000-finalize-head-body.md
FredKSchott: I guess this was more of a minor point saying: if the head would be empty in the final output, we just won't print it. But I'm fine removing that for now, we can always add that later as an enhancement/optimization if there's interest.

active-rfcs/0000-finalize-head-body.md
FredKSchott: Got nate's feedback in person, the latest draft incorporates it

active-rfcs/0000-finalize-head-body.md
FredKSchott: revisited this in the new RFC, ptal!

FredKSchott:
@matthewp @jonathantneal thank you for the feedback and comments on this RFC, and @natemoo-re for giving feedback in person.

Based on feedback (and a better understanding of how this all works in the current codebase) I took another pass at this and really re-scoped the RFC:

- `<head>` injection/collapsing has been removed as out-of-scope, to be tackled in a separate RFC. 
- No longer talking about adding or removing `head/body/html` elements on the page automatically, and instead deferring to the component author.

The RFC is now only focused on how we use the compiler and how we treat the `head/body/html` elements when we do see them in your template. 

Hopefully the smaller scope makes this easier to review. PTAL when you can!

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,80 @@
...
+
+in Astro v0.21, we have had user reports around `<head>` not acting as expected, breaking some projects from v0.20:
+- https://github.com/withastro/astro/issues/2046
+- https://github.com/withastro/astro/issues/2022
+
+Some of these issues stem from undocumented or undefined behaviors in v0.20: was it allowed for `<head>` be conditionally added in a nested expression? 
+
+Other issues stem from our attempt to provide implicit `<head>` and `<body>` support in pages and layouts. There was a quick workaround added for v0.21 where `src/pages` and `src/layouts` folder locations must be defined in config so that our parser knows how to parse them as full HTML documents and not smaller HTML fragments. This unblocked users but at the expense of breaking two larger design goals of the project:
+  - Added `src/layouts` as a new special folder location, when only `src/pages` was intented to be "special".
+  - Caused Astrp to create different output for a component based on the file path of that component.
```
jonathantneal, line 19:
```suggestion
  - Caused Astro to create different output for a component based on the file path of that component.
  ```

Comment (matthewp) [inline active-rfcs/0000-finalize-head-body.md:19]: typo `Astrp`

File: active-rfcs/0000-finalize-head-body.md [hunk 14]
```diff
@@ -0,0 +1,80 @@
...
+# Motivation
+
+in Astro v0.21, we have had user reports around `<head>` not acting as expected, breaking some projects from v0.20:
+- https://github.com/withastro/astro/issues/2046
+- https://github.com/withastro/astro/issues/2022
+
+Some of these issues stem from undocumented or undefined behaviors in v0.20: was it allowed for `<head>` be conditionally added in a nested expression? 
+
+Other issues stem from our attempt to provide implicit `<head>` and `<body>` support in pages and layouts. There was a quick workaround added for v0.21 where `src/pages` and `src/layouts` folder locations must be defined in config so that our parser knows how to parse them as full HTML documents and not smaller HTML fragments. This unblocked users but at the expense of breaking two larger design goals of the project:
+  - Added `src/layouts` as a new special folder location, when only `src/pages` was intented to be "special".
```
jonathantneal, line 18:
```suggestion
  - Added `src/layouts` as a new special folder location, when only `src/pages` was intended to be "special".
  ```

File: active-rfcs/0000-finalize-head-body.md [hunk 15]
```diff
@@ -0,0 +1,80 @@
...
+1. **Out of scope:** Support `<head>` Injection inside of individual components. This is a commonly requested user feature, but can be tackled as an additional feature, leaving this RFC focused on clarifying our existing API. This RFC does not impact that feature request.
+
+
+# Detailed design
+
+## Template Changes
+
+- `<!DOCTYPE html>`
+  - compiler: completely ignored by the compiler, stripped out from render output.
+  - runtime: a `<!DOCTYPE html>` tag is always included in final page HTML output.
```
matthewp, line 38: why append this in the runtime rather than the compiler? makes more sense to me that the compiler generates essentially the exact HTML that will be sent to the user.

active-rfcs/0000-finalize-head-body.md
jonathantneal: Quickly registers the astrp package, astrp dot org, begins selling astrp merch, etc.

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,80 @@
...
+- `<html>`, `<body>`, `<head>`
+  - compiler: will be left as-is in the component template.
+  - runtime: will be left as-is in final page HTML output.
+  - runtime: may warn if duplicate `html`, `body`, and `head` tags are included in final page HTML output.
```
matthewp, line 42: this may be difficult, remembering that the runtime only see strings of HTML, so not easy to detect multiple tags.

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Compiler Changes
+
+- `src/pages` and `src/layouts` will no longer be parsed or rendered differently from other components
```
matthewp, line 46: 👍 agree on this

active-rfcs/0000-finalize-head-body.md
FredKSchott: dang, https://github.com/astrp is already taken! :D

active-rfcs/0000-finalize-head-body.md
FredKSchott: The compiler is responsible for compiling every component individually. In this proposal, the compiler has no context for document (would need `<!DOCTYPE html>`) and component (would break if `<!DOCTYPE html>` were added). An Astro component is always considered a fragment of HTML.

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,80 @@
...
+
+- This RFC will allow you to author a page of HTML that does not actually output `<html>`, `<body>` or `<head>` elements. This can be considered an advantage in that it is more flexible than something like Svelte or Vue. However, it means we need to be more diligent with testing this kind of output across our codebase and dependencies. For example, we would need to confirm that Vite does not have trouble injecting things into an HTML document without a head or body.
+  - If we are blocked by this limitation in some way and can't reasonably unblock (for example, a non-trivial Vite limitaiton) this RFC states that it would be acceptable for Astro to add `<head>` or `<body>` elements in a way that keeps the output HTML valid but unblocks Astro. This would not be done unless absolutely necessary. 
+
+# Alternatives
+
+- Forcing that a user includes `<html>`, `<body>` and `<head>` elements in the final page output. This could be explored more, however there is not currently consensus around this idea and also there is no known reason to tackle this as a part of this RFC. If this is desired, it should be considered a seperate feature request and out of scope from this RFC.
```
matthewp, line 61: I think for SSR it makes sense that we require there be a `<head>` in *some* component (could be any component). This is because we will likely need to inject things into the head, and the best way to do that is for the compiler to tell us where the end of the head is. But we cannot know that without there being an ending head tag within some component.

Happy to explain this detail offline if this description doesn't explain it well enough.

active-rfcs/0000-finalize-head-body.md
FredKSchott: I can it make more clear that this isn't required, and is just a suggestion of something that is not impossible/blocked by this RFC.

active-rfcs/0000-finalize-head-body.md
matthewp: Ok, I follow that then.

jonathantneal: Does this mean adding `<!doctype html>` will definitely create a duplicate? I imagine the SSR answer is "yes".

Otherwise, you might need to scan the page, and then if the first thing is a `<Layout>` with `<!doctype html>`, doh!

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,80 @@
...
+- `<html>`, `<body>`, `<head>`
+  - compiler: will be left as-is in the component template.
+  - runtime: will be left as-is in final page HTML output.
+  - runtime: may warn if duplicate `html`, `body`, and `head` tags are included in final page HTML output.
+
+## Compiler Changes
+
+- `src/pages` and `src/layouts` will no longer be parsed or rendered differently from other components
+- `as?: 'document' | 'fragment';` will be removed from the compiler.
+- All Astro components will now be compiled as fragments.
```
jonathantneal, line 48: 👍 Yay!

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,80 @@
...
+
+In this design it is more "on you" to write valid HTML. This comes from the reality that an imported component can contain unknown HTML, so the compiler can't implicitly assume anything about what is or is not included included the final parent component template.  See https://github.com/withastro/astro/issues/2022 for examples of when this breaks down today. We can help with some static linting, and runtime warnings if the final HTML output is invalid. However, this RFC acknowledges the reality that already exists in v0.21 that imported components break any assumptions and help that we previously attempted to provide.
+
+# Drawbacks
+
+- This RFC will allow you to author a page of HTML that does not actually output `<html>`, `<body>` or `<head>` elements. This can be considered an advantage in that it is more flexible than something like Svelte or Vue. However, it means we need to be more diligent with testing this kind of output across our codebase and dependencies. For example, we would need to confirm that Vite does not have trouble injecting things into an HTML document without a head or body.
```
jonathantneal, line 56: I’m ready.

```xml
<!DOCTYPE xml [
	<!ATTLIST xsl:stylesheet
		id			ID	#FIXED	'!'
		xmlns:xsl		CDATA	#FIXED	'http://www.w3.org/1999/XSL/Transform'
		version			CDATA	#FIXED	'1.0'
	>
	<!ATTLIST xsl:output
		doctype-system	CDATA	#FIXED	'about:legacy-compat'
		method			CDATA	#FIXED	'html'
	>
]>
<?xml-stylesheet type="text/xsl" href="#!"?>
```

Not actually, but I like the idea that a side-effect of these SSR changes mean future HTML/XML could breathe here.

active-rfcs/0000-finalize-head-body.md
matthewp: I read it as if you add a doctype as an author the compiler will remove it. Then when we go to render the page it just prepends the doctype always as the first bytes to go out.

FredKSchott: @matthewp is correct, that is the intention

active-rfcs/0000-finalize-head-body.md
FredKSchott: danger

active-rfcs/0000-finalize-head-body.md
FredKSchott: I think that makes sense. If no head exists, is it safe to inject into the top of the `<html>` element instead?

jonathantneal (commented):
I have one lingering concern left. Looking for clarity regarding `<!doctype html>`.

active-rfcs/0000-finalize-head-body.md
jonathantneal: I understand what @matthewp wrote. Maybe that should be added?

I don’t agree with what you wrote, @FredKSchott, if you were saying a component should 'break' if it has a doctype.

```html
<!doctype html><!doctype html>
<style>.a { color: blue }</style>
<style>.b { font-weight: bold }</style><!doctype html>
<body class="a">
<body class="b">
<p>hello<!doctype html>
```

If the compiler does nothing, no harm. If the compiler omits it, no harm. If the compiler breaks? _Cruel_. 😂

active-rfcs/0000-finalize-head-body.md
jonathantneal: If we’re omitting `<!doctype html>`, pretty sure the compiler could inject everything at the top, and it would be considered `<head>`.

Intelligently finding the first `<head>` below some HTML5-Boilerplate-like comments that may exist in a `<Layout>` seems much more complicated.

jonathantneal: An aside about `<head>` —

The browser will generate `<head>` in any current HTML document. The DOM has a built-in shorthand to get it (`document.head`). `<head>` has no role, no attributes, and — unlike `<html>` or `<body>` — there have been no popular inventions that required adding classes or IDs to it.

Crawlers, open-graph see exciting things all look for individual `<meta>` & `<title>` tags anyway.

My two cents: Don’t fuss with `<head>`. The web 'has your back'.

active-rfcs/0000-finalize-head-body.md
FredKSchott: I mean, is it THAT cruel if our compiler can't handle that? :)

But yes, my intention in writing "compiler: completely ignored by the compiler, stripped out" is that it would ignored by the compiler and basically invisible, not included in the compiler render output. Let me know how I could clarify the RFC to make this more clear. This RFC would treat that code snippet as valid (although, arg!)

jonathantneal: Then I love it!

FredKSchott: 🙌 🙌 🙌

FredKSchott:
Resolved in convo above!

active-rfcs/0000-finalize-head-body.md
matthewp: The browser doesn't need head but Astro does. Astro injects things into the head, in particular it injects component styles. So we need to know where the head is. Given an `index.astro` that looks like this:

```astro
<One />
<Two />
<Three />
```

The head could be in One, or One and Two, or One, Two and Three. Any of these components could be a framework component, so we can't inspect them to know if they include head content or not (aside from parsing the generated HTML, which we don't want to do).

Open to suggestions but I just don't see a way to implement head injections without there being a head in one .astro component.

FredKSchott: We are okay injecting into the generated HTML though, correct? As a pass on the final page HTML output by finding the closing `</head>` element? If this happens at runtime, then I don't think that requires any sort of ahead-of-time analysis.

https://github.com/withastro/rfcs/blob/FredKSchott-patch-2/active-rfcs/0000-finalize-head-body.md#drawbacks The drawback section addresses this more directly, especially the  "If we are blocked by this limitation in some way and can't reasonably unblock" point.

The assumption is that we could inject into the top of the page if no `<head />` exists, but I agree that there's a strong chance our tooling & Vite wouldn't support this (Vite tries to inject things into head as well), in which case requiring a head (with a runtime error if none is found) becomes reasonable. I'll update the RFC to make this more clear.

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,80 @@
...
+  - If we are blocked by this limitation in some way and can't reasonably unblock (for example, a non-trivial Vite limitaiton) this RFC states that it would be acceptable for Astro to add `<head>` or `<body>` elements in a way that keeps the output HTML valid but unblocks Astro.
```
FredKSchott: cc @jonathantneal @matthewp making sure you get eyes on this point!

active-rfcs/0000-finalize-head-body.md
```diff
@@ -0,0 +1,80 @@
...
+
+# Drawbacks
+
+- This RFC will allow you to author a page of HTML that does not actually output `<html>`, `<body>` or `<head>` elements. This can be considered an advantage in that it is more flexible than something like Svelte or Vue. However, it means we need to be more diligent with testing this kind of output across our codebase and dependencies. For example, we would need to confirm that Vite does not have trouble injecting things into an HTML document without a head or body.
+  - If we are blocked by this limitation in some way and can't reasonably unblock (for example, a non-trivial Vite limitaiton) this RFC states that it would be acceptable for Astro to add or require a `<head>` or `<body>` element in the final output HTML in a way that keeps the output HTML valid but unblocks Astro/tooling. Reminder that it is not an explicit goal to support authoring with `<head>` or `<body>`, but it is something that we would like to support if we reasonably can.
```
FredKSchott, line 57: cc @jonathantneal @matthewp making sure you get eyes on this point!

active-rfcs/0000-finalize-head-body.md
matthewp: Yes, we inject it post-render by parsing the HTML. In an SSR scenario this seems too costly to continue to do. It also prevents streaming.

active-rfcs/0000-finalize-head-body.md
matthewp: Not a goal to support authoring *without* I think you mean?

active-rfcs/0000-finalize-head-body.md
gabrielgrant: Is the reason for treating `<!doctype>` differently from, say, `<html>` spelled out somewhere? Sorry to drop in without context (thinking this has maybe been discussed elsewhere?), but would be great if there were a bit more explanation here.

Asking because, as a naive user it feels like the least surprising behavior would just be to output the doc as created in the templates I've defined, including the doctype and wrapping tags I've defined (or haven't defined, if I don't want those wrapping tags at all eg: https://github.com/withastro/astro/issues/1305#issuecomment-928219171 )

jonathantneal: > the least surprising behavior would just be to output the doc as created in the templates I've defined

To clarify: I’m open to preserving authored tags, too. I happen to like the convenience of `<!doctype html>`, but I wouldn’t stop using Astro if I had to add them myself. My only issue has been if Astro _injects_ tags that I never authored at all.

active-rfcs/0000-finalize-head-body.md
jonathantneal: Re:
> The head could be in One, or One and Two, or One, Two and Three. Any of these components could be a framework component, so we can't inspect them to know if they include head content or not (aside from parsing the generated HTML, which we don't want to do).

If we can’t inspect those components, then wouldn’t that require at least every `.astro` page to define the `<html lang>` and `<head>` without any conditional surrounding conditional logic?

Re:
> `src/pages` and `src/layouts` will no longer be parsed or rendered differently from other components

If we can’t inspect components, then is this compiler change no longer possible?

active-rfcs/0000-finalize-head-body.md
FredKSchott: I'm not strongly against changing this! This mainly comes from a place of making it easy to do the right thing, and hard to do the wrong thing. I think most developers are at a place where they don't even think about (or even know about) `<!doctype html>` and we should be supporting those users. I am almost positive they are in the majority of web developers these days.

So the main use-case that I want to support is: if I forgot to add `<!doctype html>`, Astro adds it for me. @jonathantneal @gabrielgrant if either of you want to propose a change to this item that you'd be happier with but still keeps that use-case/goal supported, I'm probably +1!

active-rfcs/0000-finalize-head-body.md
jonathantneal: Agreed it’s meant to read as '_without_', and agreed on the sentiment.

I’m unclear on the limitations right now, which IMHO put other parts of this RFC in more uncertainty than this bit. :)

active-rfcs/0000-finalize-head-body.md
jonathantneal: Oh no, my clarification went sideways? Re: `<!doctype html>` I’m fine with injecting it, and I also wouldn’t be mad if we couldn’t for some reason. No strong opinion here. No plans to suggest alternatives. I like the convenience, and I agree with your sentiment about web developers, @FredKSchott.

active-rfcs/0000-finalize-head-body.md
FredKSchott: Maybe this is a change to how we render page HTML then? I don't think there's anything blocking our ability to, when rendering the page, say "ah yes, the head element, great. I've been looking for you. I'll just inject some things here before outputting the final page HTML."

@jonathantneal if you're still -1 on this, lets all talk on Monday. I think we've made our best effort here to support no-`<head>` HTML authoring development. But since we need to inject stuff into the head, I agree with Matthew that that would be significantly more difficult without an actual `<head>` component to inject into.

Happy to talk through alternatives but I think we could benefit from a quicker back-and-forth given the complexity here.

active-rfcs/0000-finalize-head-body.md
FredKSchott: ah gotcha, thanks for the clarification @jonathantneal !

active-rfcs/0000-finalize-head-body.md
jonathantneal: I want to offer some peace of mind that I’m +1 on injection if it’s needed for SSR performance.

> I don't think there's anything blocking our ability to, when rendering the page, say "ah yes, the head element, great. I've been looking for you..."

This is the bit I’d like us to discuss Monday. It definitely aligns with what you mentioned earlier about finding the closing `</head>` element (presumably from the page _or_ a component), but that doesn’t align with what @matthewp wrote about a need to not inspect components to know if they include head content or not (specifically to not parse the generated HTML).

FredKSchott: Yup, I'd like to understand this better as well. 

Oddly enough, you've gotten me excited now about the idea of supporting `<head>`-less HTML authoring -- as long as its not a non-trivial amount of extra work. I feel like I started out this process without caring much about it, and now I'm almost sad at the idea of losing it!

matthewp: My proposal would only require that 1 astro component includes the `<head>` tag. It could be any astro component; pages, layouts, regular components.

My opinion is that rendering should happen like this `render() -> final HTML -> browser`. Currently it works like this `render() -> rendered HTML -> parse -> injections -> finalHTML -> browser`. I think we need a very good reason to continue doing it like this.

FredKSchott: Talked in person, and are all onboard with requiring a single `<head>` element for head injection. RFC updated to include latest thinking and additional details.

active-rfcs/0000-finalize-head-body.md
FredKSchott: rewrote this as a part the latest changes, apologies for the typo!

RafidMuhymin:
I am mentioning a use case for hoisting `<link>` or `<meta>` tags to the `<head>` of the page.

Most of the images on a webpage should be lazily loaded. But some images are like banners, hero images, or logos (it should be inline though) should be *eagerly* loaded. To do this, one should not only tell the browser to eagerly load the image using `loading="eager"` but he should also preload the image. For this, users can pass a format using the `preload` prop to the `<Image>` component to *eagerly preload* the image.

But when I was implementing this feature I found that unlike `v0.20`, `<link>`, `<meta>` or other tags that should be in the `<head>` of the page aren't hoisted.

So, not only the `<Image>` component is affected but any third-party package or even a user wants to implement a similar feature can't do this.

active-rfcs/0000-finalize-head-body.md
```diff
@@ -51,29 +52,43 @@ Losing `"as": "document"` parse mode will remove some special parser handling, l
 
 In this design it is more "on you" to write valid HTML. This comes from the reality that an imported component can contain unknown HTML, so the compiler can't implicitly assume anything about what is or is not included included the final parent component template.  See https://github.com/withastro/astro/issues/2022 for examples of when this breaks down today. We can help with some static linting, and runtime warnings if the final HTML output is invalid. However, this RFC acknowledges the reality that already exists in v0.21 that imported components break any assumptions and help that we previously attempted to provide.
 
+## Head Injection Changes
+
+- runtime: will remove current post-build head injection, which involves a fragile `'</head>'` string find-and-replace.
+- compiler: will add a new `head: string` (or: `injectHead(): string`) property to the compiler transform options, which will inject the given HTML string into the bottom of a `<head>` element, if one is return by the compiler.
+- runtime: will provide this value head injection value to the compiler, and throw an exception if not used/called exactly once during a page render.
```
matthewp: 👍

active-rfcs/0000-finalize-head-body.md
```diff
@@ -51,29 +52,43 @@ Losing `"as": "document"` parse mode will remove some special parser handling, l
 
 In this design it is more "on you" to write valid HTML. This comes from the reality that an imported component can contain unknown HTML, so the compiler can't implicitly assume anything about what is or is not included included the final parent component template.  See https://github.com/withastro/astro/issues/2022 for examples of when this breaks down today. We can help with some static linting, and runtime warnings if the final HTML output is invalid. However, this RFC acknowledges the reality that already exists in v0.21 that imported components break any assumptions and help that we previously attempted to provide.
 
+## Head Injection Changes
+
+- runtime: will remove current post-build head injection, which involves a fragile `'</head>'` string find-and-replace.
+- compiler: will add a new `head: string` (or: `injectHead(): string`) property to the compiler transform options, which will inject the given HTML string into the bottom of a `<head>` element, if one is return by the compiler.
```
matthewp, line 59: I like `injectHead()` rather than `injectStyles()` / `injectLink()` because it gives the runtime the freedom to inject anything it needs into the bottom of the head, in whatever order it decides is best.

matthewp: I don't think this should be part of the compiler options though. I would expect this to be part of the runtime code. Either part of the `$$result` object or an import, but I think this is an implementation detail that doesn't need to be spelled out in the RFC.

FredKSchott: ah, I thought that we needed it to be in the compiler since `transform()` returns a final string, but maybe I misunderstood. Agreed, happy to finalize this in the implementation since this RFC is mainly focused on expected behavior.

FredKSchott:
RFC Notes:
- document this behavior in the docs site and tests
- Question: can we only add a `<!DOCTYPE html>` if an `<html>` element is detected?
  - we need a head already, so "html fragment" is not currently a supported usecase by this RFC
- Status: Accepted with consensus on RFC call

active-rfcs/0000-finalize-head-body.md
jacobrask: I’m looking to build parts of a page with Astro. These parts are assembled at the edge, allowing them to have different caching behavior, not invalidating the edge cache due to only a change in the navigation for instance.
The automatic doctype makes use cases like this quite difficult. I guess the same would be true if you want to output an RSS feed or perhaps a dynamic SVG.

matthewp:
@RafidMuhymin Thank you, this is great feedback and something we should figure out a way to support. If you'd like, we can start a new Discussion in this repo to talk about that use-case more.

RafidMuhymin:
@matthewp Yeah, I have no problem doing that if you tell me what's the way of starting a discussion in this repo

FredKSchott:
yes please! Happy to revisit that part of this, it came up in the convo and the balance we need to strike is supporting users who _are_ building pages and forgot to add the `doctype` (and may not even know what that tag means).

gabrielgrant:
@FredKSchott To confirm/clarify from the [comments buried in the review](https://github.com/withastro/rfcs/pull/26/files#r762315796) and your [notes from the RFC call](https://github.com/withastro/rfcs/pull/26#issuecomment-988205659) -- am I understanding correctly that the decision was to auto-add `doctype` only in the case that an <html> element is detected?

That seems like a sensible compromise to both seamlessly avoid problems when uses forget to add `doctype` themselves, while still allowing for generating non-html content such as [JSON data](https://github.com/withastro/astro/issues/1305#issuecomment-928219171) or @jacobrask 's use case of [page fragments, an RSS feed, or dynamic SVG](https://github.com/withastro/rfcs/pull/26/files#r767166065)

PR #44, opened 2021-12-13
Adds an RFC on enforcing static usage of client directives

matthewp:
- Start Date: 2021-12-13
- Status: Draft

## Summary

Enforce that `client:` directives not be added dynamically, to enable better performant builds.

__Not allowed__

```astro
---
import Clock from '../components/Clock.jsx';

const { shouldHydrate } = Astro.props;

const attrs = {
  'client:idle': shouldHydrate
};
---

<Clock {...attrs} />
```

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/static-directives/active-rfcs/0000-static-client-directives.md)


matthewp:
Closing this in favor of an larger build performance RFC.

PR #46, opened 2021-12-13
Improved Build Performance

matthewp:
- Start Date: 2021-12-13
- Status: Draft

## Summary

Deprecate features of Astro that are not compatible with SSR and also degrade build performance on larger sites. Replace these features with static alternatives.

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/build-performance-rfc/active-rfcs/0000-build-performance.md)

### DIFF
--- a/active-rfcs/0000-build-performance.md
+++ b/active-rfcs/0000-build-performance.md
@@ -15,7 +15,6 @@ The following features are bundled as part of this RFC:
 
 - Enforced static use of `client:` directives.
 - Deprecation of `Astro.resolve`.
-- Introduction of `local:` directive to act as a static replacement for Astro.resolve.
 
 ## Static client hydration directives
 
@@ -33,26 +32,6 @@ const { shouldHydrate } = Astro.props;
 
 Astro will be able to build the above correctly, but only JavaScript will be added to the page if the `shouldHydrate` condition is truthy.
 
-## Local directive
-
-In order to replace `Astro.resolve`, which can't be used to build things like styles and images, this RFC proposes a new sugar syntax:
-
-```astro
-<img local:src="../images/penguin.png" />
-```
-
-Note that this can already be achieved without sugar syntax by using an import statement:
-
-```astro
----
-const imgUrl = '../images/penguin.png';
----
-
-<img src={imgUrl} />
-```
-
-The sugar syntax is being introduced because we feel the above is harder to learn and less obvious of a solution.
-
 ## Deprecation of Astro.resolve
 
 The `Astro.resolve` method was added to enable adding relative links to assets in the `src/` folder. However, since this is a function it can take dynamic values that are not compatible with a static build. For example:
@@ -68,13 +47,10 @@ Instead, you can use a dynamic import to dynamically add a URL:
 
 ```astro
 ---
-const images = import.meta.glob('../images/*.png');
-
 const { animal } = Astro.props;
-const { default: animalUrl } = await import(`../images/${animal}.png`);
 ---
 
-<img src={animalUrl} />
+<img src={await import(`../images/${animal}.png`)} />
 ```
 
 The above will result in *all* of the images in `../images/` getting built, but only the one you select will be used.
@@ -113,38 +89,11 @@ The implementation will be:
 3. When rendering, if a component contains a client directive, make sure the directive is matched by the marking in __(2)__.
 4. If the marking is not there, we know that the directive must have been added statically. Throw an `Error` message for this, letting the user know that the directive must be added statically.
 
-## The `local:` directive
-
-A new `local:` directive will be introduced that can be used to set relative links on any element. The syntax looks like this:
-
-```astro
-<img local:src="../images/penguin.png" />
-```
-
-The implementation is:
-
-1. The compiler will see the `local:` directives. They must be used *statically* and not added dynamically (they will be ignored if added dynamically).
-1. The compiler will convert the value of the directive into an import statement. For example the above becomes:
-
-    ```astro
-    ---
-    import imgUrl from "../images/penguin.png";
-    ---
-    <img local:src={imgUrl} />
-    ```
-1. If the value of the directive is *not* a static string, the compiler should ignore that usage and not add an import statement.
-1. The compiler include a warning about the non-static usage of the `local:` directive.
-
-There are a couple of special cases that need to be accounted for:
-
-- `<img srcset>` value is more complex than just a single URL. We need to parse the value and add multiple import statements.
-- A `<link rel=stylesheet>` tag will resolve in an import statement that is side-effectual. As such, we should *remove* the actual link tag in the compiler to prevent it from being rendered.
-
 ## Deprecate Astro.resolve
 
 To deprecate `Astro.resolve` we should:
 
-- Add a warning to the `Astro.resolve` method that says that it is deprecated and links to documentation on alternatives such as the `local:` directive and `import.meta.glob`.
+- Add a warning to the `Astro.resolve` method that says that it is deprecated and links to documentation on alternatives such as the [local: proposal](https://github.com/withastro/rfcs/pull/59) and `import.meta.glob`.
 - While the feature is deprecated but not removed update the build process to *directly* copy anything that's resolved by `Astro.resolve`, by overloading the usage of the method during the build.
   - This will allow users who have build sites dependant on the feature to migrate slowly.
 - After one major version of Astro, replace the warning with an error, preventing its usage in dev or the build.
@@ -152,8 +101,6 @@ To deprecate `Astro.resolve` we should:
 # Drawbacks
 
 - We have heavily promoted the usage of `Astro.resolve` since its inception, as a way to build assets contained in `src/`. This will be a big departure.
-- Having `client:` and `local:` directives both be static might be unintuitive to some, since regular attributes can be added dynamically.
-  - In a future RFC it might make sense to have a blanket requirement that *all* directives be statically added to the template. This would teach developers to expect that requirement.
 
 # Alternatives
 
@@ -162,7 +109,7 @@ No other alternatives have been designed at this time. I do not believe it will 
 # Adoption strategy
 
 - Add the behaviors described in this PR behind a flag, `--experimental-static-build`.  A PR that brings partial support for this [already exists](https://github.com/withastro/astro/pull/2168).
-- Promote the usage of `local:` over `Astro.resolve` in documentation and on Discord.
+- Promote the usage of the [local: directive](https://github.com/withastro/rfcs/pull/59), if that RFC passes, over `Astro.resolve` in documentation and on Discord.
 - Add a deprecation warning to `Astro.resolve` that exists for at least 1 major version.
 - Remove `Astro.resolve` and fully enforce static directives when this feature becomes unflagged.
 
--- a/active-rfcs/0000-build-performance.md
+++ b/active-rfcs/0000-build-performance.md
@@ -1,0 +1,187 @@
+- Start Date: 2021-12-13
+- Reference Issues:
+  - Previous: https://github.com/withastro/rfcs/pull/44
+- Implementation PR: https://github.com/withastro/astro/pull/2168
+
+# Summary
+
+In order to improve build performance and scale to sites that build thousands of pages, deprecate dynamic features that forces our build to scan rendered HTML and provide static replacements.
+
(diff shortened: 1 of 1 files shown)


active-rfcs/0000-build-performance.md
```diff
@@ -0,0 +1,187 @@
...
+
+These features will be required for SSR support as well, so this RFC seeks to unify how Astro sites are built in a way that is workable both for static-site generation (SSG) and server-side rendering (SSR).
+
+# Example
+
+The following features are bundled as part of this RFC:
+
+- Enforced static use of `client:` directives.
+- Deprecation of `Astro.resolve`.
+- Introduction of `local:` directive to act as a static replacement for Astro.resolve.
```
FredKSchott, line 18: Have you checked with @jonathantneal that he's +1 on merging that idea with this PR? I know he did a lot of thinking in that original RFC and I'm sure would have good thoughts/feedback on this.

active-rfcs/0000-build-performance.md
```diff
@@ -0,0 +1,187 @@
...
+The directives used for client hydration are already static by implementation; if you attempt to add `client:load`, etc, during rendering Astro will not build that component's client-side JavaScript.
+
+This change will make this behavior defined and enforced. You can achieve dynamic usage by making the attribute static, but dynamically render a hydrated component like so:
+
+```astro
+---
+import Clock from '../components/Clock.jsx';
+const { shouldHydrate } = Astro.props;
+---
+{ shouldHydrate ? <Clock client:idle /> : <Clock /> }
```
FredKSchott, line 31: +1 This feels good for now, and would be an acceptable workaround for a final Astro v1.0 release

active-rfcs/0000-build-performance.md
```diff
@@ -0,0 +1,187 @@
...
+.filter(([path]) => (
+  path.includes(animal)
+))
+.map(([, mod]) => mod.default)[0];
+---
+
+<img src={animalUrl} />
+```
+
+The above will result in *all* of the images in `../images/` getting built, but only the one you select will be used.
```
FredKSchott, line 85: If you're mentioning workarounds here, there are two that I think work today in Vite and are worth calling out here or somewhere else in the doc:

1. `const animalUrl = await import('../images/${Astro.props.animal}.png')`
2. "just put the image in `public/`"

If this is a look ahead at how we will document this, I'm a bigger fan of documenting dynamic import than `import.meta.glob`

active-rfcs/0000-build-performance.md
```diff
@@ -0,0 +1,187 @@
...
+- In order to support SSR in the future we have to move away from page-scanning as the way to find and build assets, since SSR apps by their nature *cannot* be rendered ahead of time.
+
+# Detailed design
+
+## Enforced static use of client directives
+
+The client directives such as `client:load`, `client:idle`, etc will need to be defined in the hydrated component where they are used, and not rendered dynamically. For example the following is __not allowed__:
+
+```astro
+---
```
FredKSchott, line 101: nit: might be worth putting `// not allowed!` in the code snippet below. I know you have it above but I have to admit that i missed it while scanning the diff and began to read the code snippet

active-rfcs/0000-build-performance.md
```diff
@@ -0,0 +1,187 @@
...
+    import imgUrl from "../images/penguin.png";
+    ---
+    <img local:src={imgUrl} />
+    ```
+1. If the value of the directive is *not* a static string, the compiler should ignore that usage and not add an import statement.
+1. The compiler include a warning about the non-static usage of the `local:` directive.
+
+There are a couple of special cases that need to be accounted for:
+
+- `<img srcset>` value is more complex than just a single URL. We need to parse the value and add multiple import statements.
```
FredKSchott, line 145: can we leave this as post-v1.0? Just making sure that "to be accounted for" doesn't mean that is being marked as "will implement" by this RFC and before v1.0.

active-rfcs/0000-build-performance.md
```diff
@@ -0,0 +1,187 @@
...
+- A `<link rel=stylesheet>` tag will resolve in an import statement that is side-effectual. As such, we should *remove* the actual link tag in the compiler to prevent it from being rendered.
```
FredKSchott, line 146: say more about this? https://github.com/withastro/rfcs/discussions/28 is making the case to leave `<link rel=stylesheet>` tags completely as-is, without any additional processing. I want to make sure the two RFCs are aligned.

active-rfcs/0000-build-performance.md
```diff
@@ -0,0 +1,187 @@
...
+- Add a warning to the `Astro.resolve` method that says that it is deprecated and links to documentation on alternatives such as the `local:` directive and `import.meta.glob`.
+- While the feature is deprecated but not removed update the build process to *directly* copy anything that's resolved by `Astro.resolve`, by overloading the usage of the method during the build.
+  - This will allow users who have build sites dependant on the feature to migrate slowly.
+- After one major version of Astro, replace the warning with an error, preventing its usage in dev or the build.
+
+# Drawbacks
+
+- We have heavily promoted the usage of `Astro.resolve` since its inception, as a way to build assets contained in `src/`. This will be a big departure.
+- Having `client:` and `local:` directives both be static might be unintuitive to some, since regular attributes can be added dynamically.
+  - In a future RFC it might make sense to have a blanket requirement that *all* directives be statically added to the template. This would teach developers to expect that requirement.
```
FredKSchott, line 161: big +1 for this, I like it as a rule since directives are explicitly "a thing for the Astro compiler/runtime to handle" so it makes sense that they are as easy to detect as possible.

active-rfcs/0000-build-performance.md
```diff
@@ -0,0 +1,187 @@
...
+
+Why should we *not* do this? Please consider:
```
FredKSchott, line 163: safe to remove this section from the template

active-rfcs/0000-build-performance.md
```diff
@@ -0,0 +1,187 @@
...
+- whether the proposed feature can be implemented in user space
+- the impact on teaching people Astro
+- integration of this feature with other existing and planned features
+- cost of migrating existing Astro applications (is it a breaking change?)
+
+There are tradeoffs to choosing any path. Attempt to identify them here.
+
+# Alternatives
+
+What other designs have been considered? What is the impact of not doing this?
```
FredKSchott, line 175: needs fleshing out (or safe to say none if there truely is no other alternative)

active-rfcs/0000-build-performance.md
```diff
@@ -0,0 +1,187 @@
...
+# Adoption strategy
+
+- Add the behaviors described in this PR behind a flag, `--experimental-static-build`.  A PR that brings partial support for this [already exists](https://github.com/withastro/astro/pull/2168).
+- Promote the usage of `local:` over `Astro.resolve` in documentation and on Discord.
+- Add a deprecation warning to `Astro.resolve` that exists for at least 1 major version.
+- Remove `Astro.resolve` and fully enforce static directives when this feature becomes unflagged.
+
+# Unresolved questions
+
+- Conceptually we expect these changes to have a dramatic effect on build performance, particularly on very large sites. The slowness of Astro's current build can be felt on sites such as docs.astro.build already.
```
FredKSchott, line 186: nit: neither of these are questions. Better moved to different sections?

active-rfcs/0000-build-performance.md
matthewp: No, I didn't know the status of that RFC. I think the deprecations in this RFC are an easier sell when including this, that's why I added it.

active-rfcs/0000-build-performance.md
matthewp: Didn't realize the dynamic import one could be built. Agree that's better to document and is less code, I'll replace the example with that.

And also note that `public/` is always an option.

FredKSchott: Double check the dynamic import if you can! I know Vite documents this support but it would be great to double check

active-rfcs/0000-build-performance.md
matthewp: This was meant for `local:href`. Like if you do:

```astro
<link rel="stylesheet" local:href="../styles/main.css">
```

It would be equivalent to:

```astro
---
import '../styles/main.css';
---
```

Sound reasonable?

active-rfcs/0000-build-performance.md
matthewp: These are really 1 point. The question is, what is the effect on performance? We don't know, yet. I'll redo this section to make that more clear.

active-rfcs/0000-build-performance.md
matthewp: We should talk about this because I can see 3 options here:

1. Bring `local:` into this RFC.
1. Make `local:` its own RFC and have that block this one.
1. Remove `local:` from this RFC and go ahead with it.

A little concerned that __(3)__ puts users into an uncommon workflow (using JavaScript imports).

active-rfcs/0000-build-performance.md
matthewp: For what reason? My thinking is that `local:` should work on any element and might be unintuitive if it doesn't. For the most part the compiler doesn't need to know/care what type of element it is. This is one exception, however.

active-rfcs/0000-build-performance.md
matthewp: woops

active-rfcs/0000-build-performance.md
matthewp: I've switched to dynamic import in the example.

active-rfcs/0000-build-performance.md
FredKSchott: Because this is being adopted via an opt-in flag, I don't think the risk of 3 is too high while we're in the opt-in period. I agree that shipping it on for everyone without something _like_ `local:X` is a little more tricky and I'd share the same concern.

If you think that this RFC can push for `local:` without being a distraction the, that's still fine by me! Just making sure that in that case, you and @jonathantneal have a chance to sync anytime before the RFC call since I can't imagine that he wouldn't have valuable feedback.

active-rfcs/0000-build-performance.md
FredKSchott: YES! I really like that. We could even go one step further and translate to:

```
<!-- INPUT -->
<img local:src={`../images/${animal}.png`} />

<!-- OUTPUT -->
<img src={await import(`../images/${animal}.png`)} />
```

Also it would be nice to flesh out this behavior in a bit more detail. For example, in that case I assume we still couldn't support `local:src={x}` or `local:src={'../' + 'image.png'}` and would warn on those in the compiler? Is that feasible based on how the compiler works? 

Also, what would trigger a dynamic import vs. a static ESM import? Or, would we always do dynamic import? If we always did a dynamic import, could we then rely on Vite to handle the "this `local:src` value isn't supported" behavior by warning/erroring when it sees the final `import()` expression?

Just some thoughts!

active-rfcs/0000-build-performance.md
FredKSchott: `srcset` is more complex than `src` right? I guess more details about how this would work would address this concern, if you think it's reasonable to tackle in this RFC. I'm just unclear how our compiler would parse and handle something like:

```
<img local:srcset="./elva-fairy-480w.jpg 480w,
             ./elva-fairy-800w.jpg 800w"
     sizes="(max-width: 600px) 480px,
            800px"
     local:src="./elva-fairy-800w.jpg"
     alt="Elva dressed as a fairy">
```

active-rfcs/0000-build-performance.md
FredKSchott: ah yea, I like that a lot especially in the context of the style unification RFC!

Maybe rewrite this line to make it more clear that this is what you mean?

active-rfcs/0000-build-performance.md
matthewp: Yeah, I think translating to always using the dynamic import is cleaner as far as implementation. That way we don't have to care or inspect what the value is, and you'd get an error if the usage is too dynamic for Vite to handle.

aFuzzyBear: 🥳  yes!!! this one small feature would really powerful to include into astro, 
oh the possibilities......

matthewp:
Notes from call:

- Instead of changing how Astro.resolve works, just don't support it in the experimental build flag.
- Allow legacy build throw a flag in 1 major version instead of dropping right away.

PR #51, opened 2021-12-15
Introduce `set:html` and `set:text` directives

natemoo-re:
- Start Date: 2021-12-15
- Status: Draft

## Summary

Currently, Astro does not have an explicit API for setting the `innerHTML` or `textContent` of an element. Since a primary use case for Astro is fetching and displaying content from remote sources, we need to support this in a way that is familiar to other frameworks.

This RFC proposed the following `set:innerHTML` and `set:textContent` directives.

```astro
<!-- Injects unescaped HTML (for trusted content) -->
<article set:innerHTML={content} />

<!-- Injects escaped text for untrusted content -->
<article set:textContent={content} />
```

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/set-innerhtml/active-rfcs/0000-set-innerhtml.md)


jonathantneal:
My first thought is, can we use an HTML-like directive instead of a JS/DOM-like directive?

```astro
<!-- Injects unescaped HTML (for trusted content) -->
<article set:html={content} />

<!-- Injects escaped text for untrusted content -->
<article set:text={content} />
```

My second thought is, do we also plan to [support raw values for attributes](https://github.com/withastro/astro/issues/2095)?

```astro
<!-- Injects unescaped text for untrusted content -->
<a set:href="mailto:test@example.com">Email</a>
```

If so, is it relevant to include those as part of this RFC? If not, should this be called out as out-of-scope?

active-rfcs/0000-set-innerhtml.md
```diff
@@ -0,0 +1,196 @@
...
+
+Other frameworks have chosen not to use convention to discourage this pattern.
+
+### ⛔️ A special `{@html}` directive like Svelte uses
+
+We don't have a similar syntax construct in Astro, so this would feel out of place.
+
+### ⛔️ Do not prefix `innerHTML` or `textContent` OR prefix with a special character (`@`, `:`, `$`, etc)
+
+Per convention, directives that have special meaning in Astro should contain a `namespace:` prefix.
```
jonathantneal, line 181: I like that this RFC implicitly makes this a rule (_"should"_).

sgruenholz2:
You mentioned that these map to the default slot. What happens when named slots are also present like this?:
```
<MyComponent set:innerHtml={content}>
<div slot="special">The secret sauce</div>
</MyComponent>
```
I would expect the compiler to be able to handle this as well.

matthewp:
@sgruenholz2 

> You mentioned that these map to the default slot. What happens when named slots are also present like this?

Wouldn't use use `set:innerHTML` on the `div` in this case instead? It should maybe be an error/warning when using `set:innerHTML` on any element or component that also contains children.

matthewp:
No thoughts other than bikeshedding on the name, which can happen in the PR, so 👍 from me.

natemoo-re:
> You mentioned that these map to the default slot. What happens when named slots are also present like this?:
> 
> ```
> <MyComponent set:innerHtml={content}>
> <div slot="special">The secret sauce</div>
> </MyComponent>
> ```
> 
> I would expect the compiler to be able to handle this as well.

I think @matthewp is correct that this case should error! I would be pretty confused by this as a user. We should enforce that you either use `set:innerHTML` or pass `children`, but cannot use both. 

That's not to say it wouldn't be possible, your example would just need to be a little more explicit. Since you _can_ use `set:innerHTML` with any Component, there's no reason it couldn't be used with the built-in `Fragment` component (which already supports `slot`).

```astro
<MyComponent>
  <Fragment set:innerHTML={content} />
  <div slot="special">The secret sauce</div>
</MyComponent>
```

natemoo-re:
> My second thought is, do we also plan to [support raw values for attributes](https://github.com/withastro/astro/issues/2095)?
```html
<!-- Injects unescaped text for untrusted content -->
<a set:href="mailto:test@example.com">Email</a>
```

@jonathantneal I really like this idea but I'd hesitate to use `set:` for this use case. In this RFC, `set:innerHTML` and `set:textContent` are special props that can project their values into the parent node's `children`, but your example generates an attribute in the output. Users might be confused by the overlap in syntax but difference in behavior.

I'd love to revisit this in a follow-up RFC, potentially with a different prefix like `raw:`

jonathantneal:
> No thoughts other than bikeshedding on the name, which can happen in the PR, so 👍 from me.

Pointing out that this _is_ the PR, @matthewp. Previously, we’ve required the consensus on the naming before accepting a proposal, so @natemoo-re, would you be willing to reconsider the name? The main issues are that Astro attributes are not DOM properties and HTML attributes are case-insensitive.

natemoo-re:
> [W]ould you be willing to reconsider the name? The main issues are that Astro attributes are not DOM properties and HTML attributes are case-insensitive.

I'm not 100% sold that the naming outlined in this RFC is flawed. Familiarity and framework-parity can go a long way towards keeping Astro's learning curve small. Astro also does _not_ normalize attribute case during parsing.

That being said, I'm open to alternative suggestions @jonathantneal! 👍

I like `set:html` and `set:text`. I appreciate how terse both are. My only concern is that using the `set:html` directive with a `style` or `script` tag could be confusing, though I suppose it's not _more confusing_ than `set:innerHTML` would be. Thoughts?

jonathantneal:
> Familiarity and framework-parity can go a long way towards keeping Astro's learning curve small.

[`Element`​`.`**`set`​`HTML`**](https://developer.mozilla.org/en-US/docs/Web/API/Element/setHTML) 😄😄😄

> My only concern is that using the `set:html` directive with a style or script tag could be confusing, though I suppose it's not more confusing than `set:innerHTML` would be. Thoughts?

I agree. 😄😄😄

natemoo-re:
> [`Element`​`.`**`set`​`HTML`**](https://developer.mozilla.org/en-US/docs/Web/API/Element/setHTML) 😄😄😄

Cool, I'm sold! This one was news to me! 😄

matthewp:
> Pointing out that this is the PR, @matthewp. Previously, we’ve required the consensus on the naming before accepting a proposal, so @natemoo-re, would you be willing to reconsider the name?

I meant the PR to implement the feature. Previously we have deferred bikeshedding until after an RFC passed to avoid blocking on naming and to keep focus on the idea. This doesn't mean that a name doesn't need consensus; just that it occurs when reviewing the PR feature. So if you feel strongly enough you can Request Changes in the PR and it can be discussed there until resolved.

At least this is how we did it in the previous RFC process. If something changed on that when we moved to this repo I wasn't aware.

FredKSchott:
@matthewp's comment is how I'd remembered this as well: it's on the RFC creator to pitch a certain name, and then it's on a maintainer to object to consensus if they feels strongly enough to block progress.

The idea here is to open the door for a 3rd option of a reviewer being able to say "I'm not blocking consensus, but I'm not 100% agreed on this either. As you implement and in the final PR I'd love to revisit the naming on last time before merging."

FredKSchott:
> I think @matthewp is correct that this case should error! I would be pretty confused by your example as a user. We should enforce that you either use set:innerHTML or pass children, but cannot use both.

+1 for this!

re naming: I'm fine with `set:html` or `set:innerhtml`, I think they're both clear on what they accomplish.

jonathantneal:
> Previously we have deferred bikeshedding until after an RFC passed to avoid blocking on naming and to keep focus on the idea.

I’m glad it still works this way. My experience with `@component` seemed very different. I vividly remember asking many times if we could bike-shed on the name. I must have misunderstood.

matthewp:
We have consensus on this proposal!

PR #54, opened 2021-12-15
Recursive Components

sgruenholz2:
- Start Date: YYYY-MM-DD
- Status: Draft

## Links

- [Full Rendered Proposal]()


sgruenholz2:
Official pull request for recursive component rendering.

sgruenholz2:
And I already screwed it up. Re-opening.

active-rfcs/0000-recursive-components.md
```diff
@@ -0,0 +1,150 @@
+-   Start Date: Date: 2021-12-15
+-   Reference Issues: https://github.com/withastro/rfcs/discussions/45
+-   Implementation PR:
+
+# Summary
+
+Allow a way for astro components to render themselves recursively.
```
matthewp, line 7: There's a lot of good content in this section but I think most of it belongs in the __Motiviation__ section. This sentence is good, and maybe add one or two explaining that `Astro.self` is the proposed solution.

active-rfcs/0000-recursive-components.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+Note that in the above example, the `<Self />` component provides
+access to this component's render function, allowing it to
+reference itself.
+
+# Motivation
+
+Handling this use case lets .astro components do things that other component
+frameworks can do, making it a 1st class component framework in its own right.
```
matthewp, line 91: As said above, moving some of the paragraphs from the Summary to here would help this section.

Also, I would write up an example of how you have to do this today (probably by adding a second component).

active-rfcs/0000-recursive-components.md
```diff
@@ -0,0 +1,150 @@
...
+                <MyRecursiveItem itemId={childId} items={items} />
+            ))}
+        </ol>
+    ) : ""}
+</>
+```
+
+# Alternatives
+
+The other design pattern considered was to mimic the Svelte syntax, using
```
matthewp, line 120: Referring to Svelte's solution here helps the proposal, good job! I would be curious to hear what solutions are in other similar frameworks, if any. For example does Vue try to solve this problem, and if so how?

active-rfcs/0000-recursive-components.md
```diff
@@ -0,0 +1,150 @@
...
+reference itself.
+
+# Motivation
+
+Handling this use case lets .astro components do things that other component
+frameworks can do, making it a 1st class component framework in its own right.
+
+# Detailed design
+
+The `Astro.self` property exposes the render function of the component.
```
matthewp, line 95: I think the implementation here is that the runtime `$$result.createAstro` function that is created inside of render gets passed the `$$Component` variable (which is the Astro component) and sets it to `Astro.self`.

What this means is that `Astro.self` is likely not going to be available in `getStaticPaths`. I presume that's ok. If not we should discuss it.

active-rfcs/0000-recursive-components.md
```diff
@@ -0,0 +1,150 @@
...
+referencing things like `Astro.props` and `Astro.request.params` in the frontmatter.
+This is just one more `Astro` property that becomes available there.
+
+# Adoption strategy
+
+- No backwards compatibility concerns
+- No potential for breaking changes
+- Worth noting that recursive rendering can be already achieved with any number of other
+platforms: React, Vue, Svelte, etc., so we can approach this at our leisure with
+the short term answer to this being: "Do that in your own platform for now".
```
matthewp, line 143: 👍

matthewp:
Left some comments, but overall this sounds like a good approach to me. Mostly curious how this problem is handled in other frameworks than just Svelte, as that would strengthen the proposal I believe.

active-rfcs/0000-recursive-components.md
sgruenholz2: Yes, exactly. I can't think of a case where this would be needed within getStaticPaths.

PR #58, opened 2021-12-20
Recursive components

sgruenholz2:
- Start Date: 2021-12-20
- Status: Draft

## Links

- [Full Rendered Proposal]()


sgruenholz2:
Updated with how other platforms handle recursion, and the limitations of the single file per component design pattern.

active-rfcs/0000-recursive-components.md
```diff
@@ -0,0 +1,160 @@
...
+We want to use a recursive component to render the entire tree,
+(or one of its branches) by passing in both the entire tree (the items object)
+and an itemId like this:
+
+```
+// src/components/RecursiveItem.astro
+---
+const {itemId, items} = Astro.props;
+const {childIds} = items[itemId];
+const Self = Astro.self;
```
jonathantneal, line 56: Can this be removed? I really like how you’ve used `<Astro.self>` later.

```suggestion
```

active-rfcs/0000-recursive-components.md
```diff
@@ -0,0 +1,160 @@
...
+        <ol>
+            {childIds.map(childId => (
+                <Astro.self itemId={childId} items={items} />
+            ))}
+        </ol>
+    ) : ""}
+</>
+```
+
+Note that in the above example, the `<Self />` component provides
```
jonathantneal, line 70:
```suggestion
Note that in the above example, the `<Astro.self />` component provides
```

active-rfcs/0000-recursive-components.md
```diff
@@ -0,0 +1,160 @@
...
+# Alternatives
+
+The other design pattern considered was to mimic the Svelte syntax, using
+`<astro:self />`
+
+This follows an already established precedent (for those familiar with Svelte),
+but there's nothing else in Astro yet that leverages `astro:` prefixes as
+components.
+
+By contrast, using `const Self = Astro.self` is more explicit,
```
jonathantneal, line 137:
```suggestion
By contrast, using `Astro.self` is more explicit,
```

natemoo-re:
This is great, excited to discuss it tomorrow!

@sgruenholz2 would you mind updating the "Fully rendered proposal" link to point to the following URL?

```
https://github.com/sgruenholz2/rfcs/blob/recursive-components-patch-1/active-rfcs/0000-recursive-components.md
```

active-rfcs/0000-recursive-components.md
```diff
@@ -0,0 +1,159 @@
...
+- Worth noting that recursive rendering can be already achieved with any number of other
+platforms: React, Vue, Svelte, etc., so we can approach this at our leisure with
+the short term answer to this being: "Do that in your own platform for now".
+- Need to update Documentation
+
+# Unresolved questions
+
+Svelte has some [additional restrictions](https://svelte.dev/docs#svelte_self)
+about how/where theirs can be used. It must be within a conditional or an "each" loop.
+They are attempting to protect against infinite loops. Should we build in similar protections?
```
natemoo-re, line 159: The answer to this question might push us to resolve the specific syntax for this feature.

From the Svelte docs:

> It cannot appear at the top level of your markup; it must be inside an if or each block or passed to a component's slot to prevent an infinite loop.

One potential problem with `Astro.self` is that we're unable to statically detect usage unless it is literally authored as `<Astro.self />` because reassignment would confuse the compiler.

On the other hand, `astro:self` would always be statically analyzable.

Alternatively, we could exclusively document the feature using `Astro.self` and warn when encountering any literal `<Astro.self />` usage. That would likely cover 80% of the uses.

matthewp:
Vue does it by referring to itself by filename: https://v3.vuejs.org/api/sfc-script-setup.html#using-components

matthewp:
We have reached consensus on this.

PR #59, opened 2021-12-20
Proposal: Support a `local:` attribute prefix

jonathantneal:
- Start Date: 2021-12-20
- Status: Draft

## Summary

This proposal seeks to improve source-relative URL support in Astro files.

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/jn.relative-url-scheme-proposal/active-rfcs/0000-relative-url-scheme.md)

### DIFF
--- a/active-rfcs/0000-relative-url-scheme.md
+++ b/active-rfcs/0000-relative-url-scheme.md
@@ -1,0 +1,95 @@
+- Start Date: 2021-12-20
+- Reference Issues: [#46](https://github.com/withastro/rfcs/pull/46)
+- Implementation PR: <!-- leave this empty -->
+
+# Summary
+
+Astro should support source-relative URL resolution in the HTML block of Astro files.
+
+# Example
+
+To resolve the source-relative URL of an image, use a `local:` prefixed URL.
+
+```html
+<img src="local:kitten.avif" alt="Kitten" />
+```
+
+# Motivation
+
+In `.astro` files, it’s not always easy to know what a relative URL is relative to.
+
+```astro
+<img src="kitten.avif" alt="kitty" />
+```
+
+In this example, is `kitten.avif` relative to the output `.html` or the source `.astro`? Or, if this example is found within a component `.astro` file, is it relative to that component file?
+
+This ambiguity raises a need for:
+
+- An intuitive way to author URLs relative to a source file.
+- An intuitive way to author URLs within the HTML of an `.astro` file.
+
+# Detailed design
+
+A `local` scheme instructs Astro to resolve the URL relative to the source file.
+
+```astro
+<img srcset="local:kitten.avif 1x, local:kitten@2x.avif" alt="Kitten" />
+```
+
+```astro
+<picture>
+  <source srcset="local:media/kitten@2x.avif" media="(min-width: 800px)" />
+  <img src="local:media/kitten.jpg" alt="" />
+</picture>
+```
+
+```astro
+<div style="background-image: url(local:kitten-background.avif);"></div>
+```
+
+```astro
+<div data-my-framework-attr-for-image="local:kitten-background.avif"></div>
+```
+
+# Drawbacks
+
+All attribute values need to be checked for `local:`. See alternatives below
+
+The provided examples are intuitive and convenient for authors, but not immediately obvious to an implementer.
+
+# Alternatives
+
+- Limit support to `local:` prefixed attributes
+  ```html
+  <img local:src="kitten.avif" alt="Kitten" />
+  ```
+  - It’s not obvious if this would be more performant than checking every attribute for a `local:` prefix.
+  - It’s even less obvious if this would be more performant when the attribute value may include multiple URLs (the `srcset`, `style`, or `data-` attributes), or wraps the URL in `url()`, `url("")`, `url('')`, etc.
+- Limit support to `import` statements and do not support source-relative URLs in the HTML of `.astro` files.
+  ```astro
+  ---
+  import kittenImage from './kitten.avif?url'
+  ---
+  <img src={kittenImage} alt="Kitten" />
+  ```
+  - All relative URLs in a `.astro` file would need to be added to the top of the file as imports.
+  - All path imports would need to use a special directive, like `?url`. 
+  - It is likely actual usage would be far less simple or visually co-located as in this example.
+  - This would force authors to name imports, which would fatigue authors. <sup>[1](https://hilton.org.uk/blog/why-naming-things-is-hard)</sup>
+
+# Adoption strategy
+
+This proposal seeks to improve and replace `Astro.resolve`, but it is not necessary to drop `Astro.resolve` upon accepting this proposal.
+
+# Unresolved questions
+
+Will authors reasonably expect `local:` to work in other environments than HTML attribute values?
+
+```html
+<style>
+:host {
+  background-image: url(local:kitten.avif);
+}
+</style>
+```
--- a/active-rfcs/0000-relative-url-scheme.md
+++ b/active-rfcs/0000-relative-url-scheme.md
@@ -10,7 +10,7 @@ Astro should support source-relative URL resolution in the HTML block of Astro f
 
 To resolve the source-relative URL of an image, use a `local:` prefixed URL.
 
-```html
+```astro
 <img src="local:kitten.avif" alt="Kitten" />
 ```
 
@@ -39,7 +39,7 @@ A `local` scheme instructs Astro to resolve the URL relative to the source file.
 <img srcset="local:kitten.avif 1x, local:kitten@2x.avif" alt="Kitten" />
 ```
 
-That example could be transformed into or handled as tho it were written like this:
+That example would be handled as tho it were written like this:
 
 ```astro
 <img srcset=`${import('./kitten.avif?url')} 1x, ${import('./kitten@2x.avif?url')}` alt="Kitten" />
@@ -53,6 +53,8 @@ That example could be transformed into or handled as tho it were written like th
   <img src="local:media/kitten.jpg" alt="" />
 </picture>
 ```
+
+Similarly, that example would be handled as tho it were written like this:
 
 ```astro
 <picture>
@@ -82,7 +84,7 @@ The provided examples are intuitive and convenient for authors, but it may not b
 # Alternatives
 
 1. Limit support to `local:` prefixed attributes
-  ```html
+  ```astro
   <img local:src="kitten.avif" alt="Kitten" />
   ```
   - A namespaced attribute would be more complex for both the author and the implementer than a [scheme](https://developer.mozilla.org/en-US/docs/Learn/Common_questions/What_is_a_URL#scheme) when assets are referenced in `style`, `srcset`, or `data-` attributes with `url()`. Those are all common techniques used on the web. A namespaced attribute would also necessitate a more careful parsing of `srcset` and `url()`, or otherwise the loss of that functionality entirely if it were deemed too complex. By limiting `local:` to URLs, we reduce this complexity and give this functionality the freedom to be applied to other usecases in the future.
@@ -98,18 +100,30 @@ The provided examples are intuitive and convenient for authors, but it may not b
   - It is likely actual usage would be far less simple or visually co-located as in this example.
   - This would force authors to name imports, which would fatigue authors. <sup>[1](https://hilton.org.uk/blog/why-naming-things-is-hard)</sup>
 
+
+
 # Adoption strategy
 
-This proposal seeks to improve and replace `Astro.resolve`, but it is not necessary to drop `Astro.resolve` upon accepting this proposal.
+The `local` scheme is intended to improve and replace `Astro.resolve`, but its addition does not require the removal of `Astro.resolve`.
+
+This feature can be initially limited to attributes, as we consider whether or how the `local` scheme could be used in other areas.
 
 # Unresolved questions
 
-Will authors reasonably expect `local:` to work in other environments than HTML attribute values?
+Would authors reasonably expect `local:` to work in other environments than HTML attribute values? If so, should this proposal additionally unlock that ability?
 
-```html
+```astro
 <style>
 :host {
   background-image: url(local:kitten.avif);
 }
 </style>
+```
+
+That example would be handled as tho it were written like this:
+
+```astro
+<style set:html=`:host {
+  background-image: url(${import('./kitten.avif')});
+}` />
 ```
--- a/active-rfcs/0000-relative-url-scheme.md
+++ b/active-rfcs/0000-relative-url-scheme.md
@@ -2,17 +2,23 @@
 - Reference Issues: [#46](https://github.com/withastro/rfcs/pull/46)
 - Implementation PR: <!-- leave this empty -->
 
+
+
 # Summary
 
 Astro should support source-relative URL resolution in the HTML block of Astro files.
 
+
+
 # Example
 
-To resolve the source-relative URL of an image, use a `local:` prefixed URL.
+To resolve the source-relative URL of an image, use a `local:` prefixed attribute.
 
 ```astro
-<img src="local:kitten.avif" alt="Kitten" />
+<img local:src="kitten.avif" alt="Kitten" />
 ```
+
+
 
 # Motivation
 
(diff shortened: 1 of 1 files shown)


active-rfcs/0000-relative-url-scheme.md
```diff
@@ -0,0 +1,95 @@
...
+
+The provided examples are intuitive and convenient for authors, but not immediately obvious to an implementer.
+
+# Alternatives
+
+- Limit support to `local:` prefixed attributes
+  ```html
+  <img local:src="kitten.avif" alt="Kitten" />
+  ```
+  - It’s not obvious if this would be more performant than checking every attribute for a `local:` prefix.
```
matthewp, line 67: I think it probably is, you only have to check the first 6 characters of each attribute rather than checking for 6 characters anywhere in an N character string.

active-rfcs/0000-relative-url-scheme.md
```diff
@@ -0,0 +1,95 @@
...
+
+```astro
+<picture>
+  <source srcset="local:media/kitten@2x.avif" media="(min-width: 800px)" />
+  <img src="local:media/kitten.jpg" alt="" />
+</picture>
+```
+
+```astro
+<div style="background-image: url(local:kitten-background.avif);"></div>
```
matthewp, line 48: I like how this idea, rather than `local:` as an attribute prefix, allows you to use it anywhere and Astro doesn't need to special-case certain usage.

active-rfcs/0000-relative-url-scheme.md
matthewp: I'm not worried about the performance of doing it as a URL scheme, to be clear, just nothing that I think it isn't as performant as an attribute would be.

matthewp:
- I like the change to a URL scheme rather than an attibute value prefix because it allows the user the flexibility to use this anywhere inside of any attribute value.
- However this might go against the idea that Astro doesn't touch your code unless you use a directive. What if the user intended the `local:` string to be inside of an attribute value?
  - I think it's still worth doing, just calling out that it contradicts with what we had been saying we would do.
  - This concern could be alleviated by allowing a way to escape `local:` when it's intended to be used as a string and not a URL scheme.

active-rfcs/0000-relative-url-scheme.md
```diff
@@ -0,0 +1,95 @@
...
+In this example, is `kitten.avif` relative to the output `.html` or the source `.astro`? Or, if this example is found within a component `.astro` file, is it relative to that component file?
+
+This ambiguity raises a need for:
+
+- An intuitive way to author URLs relative to a source file.
+- An intuitive way to author URLs within the HTML of an `.astro` file.
+
+# Detailed design
+
+A `local` scheme instructs Astro to resolve the URL relative to the source file.
```
matthewp, line 34: I would expect this section to get into the *how* Astro will resolve.

For example, we had talked about offline that these would be translate to dynamic imports. If that's the case it should be explained in this section.

sgruenholz2:
I may be missing some context, but coming in fresh to Astro here are my initial thoughts/expectations:

- I expect anything in an image src to always be relative to the static compiled html and where the page is compared to the public dir. So, in practice, I would never use something like `Kitten`. I would only ever use `Kitten` or `Kitten` or roll my own custom src helper function if I need to conditionally rebase.
- If you need a way to determine which src image files should be copied into the public dir, I would expect to set that up as part of the config via glob patterns, or sure, do that automatically if an import is present in the frontmatter that then gets referenced as a src in the html, or if there's an Astro.resolve() helper present. Or you could have a standard src/assets dir where everything in there always gets copied into the public dir.
- It feels like there's probably a better approach to solve the underlying problems here.
- If we do go with a `local:` prefix as part of the src value, then yes, I'd expect it to also work for styles.

active-rfcs/0000-relative-url-scheme.md
natemoo-re: I appreciate this as well, but it also exposes a concern. Would this be supported in other files within an `.astro` project? If so, how could we do that without scanning the final HTML output?

natemoo-re:
Even though we should prioritize DX over what is easier to implement, I think it’s worth considering that there isn’t a mechanism to support a feature like this in Rollup/Vite at the moment and we’d be forging our own path here. 

I have open questions about how we handle this URL scheme in CSS, things like the `style` property (especially custom properties), and if we support this in non-Astro files.

active-rfcs/0000-relative-url-scheme.md
jonathantneal: I would expect any code that is currently run thru our resolver to get picked up by this, with the addition of attributes by this RFC.

Perhaps I can flesh out a `resolveId` prototype for Vite, to demonstrate how the functionality could be paired in other compiled files.

Related: Does `resolveId` handle `<script>import X from '@some/npm-module'</script>`? These imports are currently resolved in client-side code, but I wasn’t sure by what mechanism.

jonathantneal:
Hey @sgruenholz2, I appreciate the feedback. I’ll do my best to address your points.

> I expect anything in an image src to always be relative to the static compiled html.

I feel the same way. I believe others do, as well. However, I’m convinced after many conversations I’m convinced we are, at the very least, not a super-majority. 😅

> ... you could have a standard src/assets dir where everything in there always gets copied into the public dir.

Well, that kinda seems like another `public` directory. What we’ve seen before Astro and now within Astro is that developers continually reject separation by type and prefer colocation by context. They want to keep files close together that go together and change together. It’s the "colocation principle" <sup>[1](https://kentcdodds.com/blog/colocation)</sup>.

> It feels like there's probably a better approach to solve the underlying problems here.

The prior approach has been `Astro.resolve`, and this approach has been problematic. I say that as someone who [helped implement it](https://github.com/withastro/astro/pull/1072). Our repo has quite a few "relative" and "resolve" issues <sup>[1](https://github.com/withastro/astro/pull/1481)</sup>, and I think the same is true within Discord. I’m very okay with this. I like that folks like Astro best when they can avoid JavaScript expressions. 😄

In a [prior proposal](https://github.com/withastro/astro/issues/1842) I did suggest something like this:

```astro
---
import Kitten1x from './my-kitten.avif' assert { type: "url" }
import Kitten2x from './my-kitten@2x.avif' assert { type: "url" }
import linkToScript from './script.js' assert { type: "url" }
---
```

However, there was still a significant desire to keep links within HTML, which in fairness is how HTML intended it <sup>[1](https://github.com/withastro/rfcs/pull/46#pullrequestreview-830828859)</sup>. I also bring up "_fatigue_" in this RFC, because import statements introduce this by requiring authors to come up with names for any asset they’ll want to use, and possibly yet another naming scheme for assets in general.

active-rfcs/0000-relative-url-scheme.md
matthewp: I think this needs to be scoped to just .astro files. It should happen in the compiler by replacing `'local:./path/to/something.js'` with `{await import('./path/to/something.js?url')}`

jonathantneal: > I think this needs to be scoped to just `.astro` files.

Thank you for the design detail. I was expecting this to work within any file within `src`.

Related question: Would you expect `local:` to work within `<script></script>` or `<style><style>` within `.astro`?

```astro
<style>
:host {
  background: url(local:kitten.jpg);
}
</style>
```

FredKSchott: Is it possible to keep the original scope of only introducing a `local:src` directive? That would seem to meet the goals of the motivation section, so it's not clear to me why we would need to increase the scope to support `url(...)` as well.

Also I don't believe that this is required: Vite says that it resolves `url()` already (src-relative), similar to how CSS `@import` and ESM `import` work.

matthewp: I can see how that would be useful, but a couple of problems:

- This overlaps with usecases for `define:vars`
- Static styles and hoisted scripts have to be static when .astro files are compiled, this would break that.

I would probably recommend scoping this proposal to attribute values to avoid hitting these types of issues.

This is a good reason to prefer `local:src` attribute prefix instead, as well. Although that has the downsides already discussed.

active-rfcs/0000-relative-url-scheme.md
```diff
@@ -0,0 +1,95 @@
...
+The provided examples are intuitive and convenient for authors, but not immediately obvious to an implementer.
+
+# Alternatives
+
+- Limit support to `local:` prefixed attributes
+  ```html
+  <img local:src="kitten.avif" alt="Kitten" />
+  ```
+  - It’s not obvious if this would be more performant than checking every attribute for a `local:` prefix.
+  - It’s even less obvious if this would be more performant when the attribute value may include multiple URLs (the `srcset`, `style`, or `data-` attributes), or wraps the URL in `url()`, `url("")`, `url('')`, etc.
```
FredKSchott, line 68: Is performance the biggest issue here? I feel like complexity is a big difference between these two that I don't see mentioned here. 
- Implementation complexity: basically the convo above about how we could support this across all use-cases, when it's needed at compile time. 
- User complexity: How can we document this well to developers? ex: you couldn't do `src={someVar}` because it needs to be scanned at compile-time, right?

FredKSchott:
Love the exploration happening here, but agreed with other comments here that this takes a step away from previous iterations of this feature in ways that are definitely concerning from a complexity perspective. I'm a bigger fan of the smaller-scoped, more explicit `local:src` directive that was proposed in the last iteration.

active-rfcs/0000-relative-url-scheme.md
jonathantneal: I agree. Plus, if it were `local:src` then we still have to check the value, unless we drop support for several of the common use cases (`srcset`, `data-`).

jonathantneal: This note has been removed from the RFC.

active-rfcs/0000-relative-url-scheme.md
jonathantneal: I’ve added those details to the section. Let me know if they address this well.

active-rfcs/0000-relative-url-scheme.md
jonathantneal: Yea, the focus of this section should not have been performance, but complexity.

I’ve updated this section to focus on the complexity of a namespaced attribute versus a scheme. Would you check it out and let me know if you feel it better addresses your issues?

jonathantneal:
This RFC will be updated with the following changes:

- Use the alternative design (`local:` attribute prefix).
- Ensure the following details are already present (otherwise added) from the call:
  - `local:` attribute prefix treats the entire attribute value as the URL.
  - `local:` attribute prefix works on any attribute.
    - limiting the attribute prefix to one attribute name or element type would add complexity.
  - `?url` is presumed to be appended to the value.
  - attribute values with partial sources (`srcset`, `style`, some data attributes) are not handled, but can be handled in a separate, future RFC.

active-rfcs/0000-relative-url-scheme.md
jonathantneal: Hopefully these are resolved by aa6df9d5fa30d5086f5562b205a2015f69939721

active-rfcs/0000-relative-url-scheme.md
jonathantneal: This isn’t necessarily resolved by the updates in aa6df9d5fa30d5086f5562b205a2015f69939721, but it did not come up in the RFC discussion, so I am resolving this thread.

FredKSchott:
LGTM based on notes! For the record books: I'd still love to bikeshed the final client directive name in the implementation RFC (slight preference for  `import:src` or `resolve:src`) but that is in no way blocking this RFC.

Also: is there any reason that you're saying "attribute prefix" instead of "client directive"? Do we need a new term for client directives that can dynamically be applied to any `local:*` attribute?

jonathantneal:
> slight preference for `import:src` or `resolve:src`

The name `local` was chosen to avoid confusion with 'import' resolution in JS. The name `import` or `resolve` may suggest it works like ESM `import` or CJS `resolve`. We experienced this issue previously, due to the naming of `Astro.resolve`.

We didn’t spend much time discussing how the URL is transformed on the call, and we didn’t specifically discuss what kinds of URLs are allowed (outside that it would append `?url`). The examples demonstrate the intended resolution, which is local only, but it’s not actually written down. It should be written down, unless we each think it should work differently.

The intended behavior is to match how one would normally reference the value, so that `src="kitten.avif"` would become `local:src="kitten.avif"`.

If the name were `import` or `resolve`, we might see authors writing something like this and expecting it to work:

```astro
<link resolve:href="normalize-css" type="stylesheet" />
```

jonathantneal:
> Also: is there any reason that you're saying "attribute prefix" instead of "client directive"? Do we need a new term for client directives that can dynamically be applied to any `local:*` attribute?

If we wanted, we could refer to them as **local directives**. Thoughts? If there’s consensus, I can use that term. We have a **class:list** directive. I wouldn’t recommend **client directives**, since I expect those are focused on a set of features implemented with a `client:` attribute prefix.

FredKSchott:
> `<link resolve:href="normalize-css" type="stylesheet" />`

Let me think about this a bit more, but I'm not sure I agree on the risk vs. reward here. Because we're using `import()` in our implementation, talking about this as sugar for `import()`, and would most likely document this in our docs by talking about `import()`, I think that the import/resolve connection might be a feature and not a bug.

`import:src` or `resolve:src` both communicate this more clearly to a user in a self-documenting way. With `local:src`, as a user I'm definitely going to think "wait, what does local mean again?" at least a few times. At least, I've never made that connection for the meaning of `local` until you just explained it to me.

jonathantneal:
> I think that the import/resolve connection might be a feature and not a bug.

To be certain, _would_ you expect this feature to resolve node modules?

Currently, as this transformation is static, `@csstools/normalize.css` would become `./@csstools/normalize.css?url` (and `./kitten.avif` would become `././kitten.avif`, which can still resolve)

aFuzzyBear:
Hi guys, I hope you dont mind me commenting on this, I have bee giving this a fair bit of thought since the call yesterday. 
Reading the comments above, Im very much a proponent for this rfc, its spirit and goals are very much aligned to providing an _easier_  syntax to help _resolve_ files from the `./src/` 
If we are discussing the bikeshedding merits of `local:` I would like to express my own thoughts on this,
The use of the term `local:` confers its purpose very well, applied as a attribute directive, *Tangentially, I feel that it was inevitable that the astro lexicon grew to encompass more terms, "client directive": `client:` has more application and easier separation if there was a counteract or balance to it, with use of attribute directives, `local:src` being an example.
Local is like a "Ronsil" description in my opinion; Ronsil for those unfamiliar was a paint manufacturer whose marketing strapline was "It does what it says on the Tin". 
This sort of simple self explanatory application of adverbs and pronouns must be central to the debate over bikeshedding terms, 
There was however a contention that comes from the actual use of the term `local:`, principle its meant to replace the use of `Astro.resolve()` with a more static, syntactically friendly attribute, that would help both compiler and developer.  If this is the case, then in equal measure we must consider the term `resolve:src` or `resolve:href` to its contemporary counterpart, `local:src` `local:href` .

I must confess I am torn between the two, its not often I find myself being ambivalent about things, I do not wish to be the one who makes the final decision on this, 😅

jonathantneal:
We may not have expected this feature does the same thing, but I think we can reach consensus on this issue concerning resolution.

I recommend that we embrace web platform resolution in Astro HTML, perhaps resembling [Deno](https://deno.land/manual@v1.17.1/linking_to_external_code). I recommend that we **do not** embrace CommonJS resolution.

### CommonJS Resolution

In **NodeJS**, `require` and `import` statements require a leading `./`, and are otherwise considered **CommonJS** modules. **CommonJS** modules (_like `@csstools/normalize.css`_) are resolved against a lengthy algorithm. <sup>[ref](https://nodejs.org/api/modules.html#modules_all_together)</sup>
  - **ComonJS** resolution is not on any standards track.
  - **ComonJS** resolution has bad performance. <sup>[ref](https://github.com/nodejs/modules/issues/477#issuecomment-949322749)</sup>
  - For backwards compatibility, **NodeJS** supports **ComonJS** resolution in `import` statements (and dynamic imports).

---

### Platform Import Resolution

In **JS**, `import` statements (and dynamic imports) require a leading `./` or an absolute URL.
  - There is a **Import Maps** proposal, which would allow imports _similar_ to **CommonJS**.  <sup>[ref](https://wicg.github.io/import-maps/)</sup>
  - **Import Maps** are on a standards track. <sup>[ref](https://chromestatus.com/feature/5315286962012160)</sup>
  - **Import Maps** and absolute URLs are supported in **Deno**. <sup>[ref](https://deno.land/manual/linking_to_external_code/import_maps)</sup>
  - **Import Maps** and absolute URLs are not supported (out-of-the-box) in **NodeJS**.
  - **Import Maps** resolve _differently_ than **CommonJS**.

### Platform URL Resolution

In **HTML** `src`, `href`, etc. & **JS** `location`, `URL`; relative URLs **do not require** a leading `./`.
  - Example, given a location of `https://example.com/cats/`:
    - In HTML, `` resolves to `https://example.com/cats/assets/kitten.avif`.
    - In JS, `new URL("assets/kitten.avif", location)` resolves to `https://example.com/cats/assets/kitten.avif`.
  - This is very convenient for sibling URLs.

---

As Astro blends server-side and client-side functionality, we are caught between these worlds, and others. For example, we support (thru Vite) URLs that are _relative to the source file_, or _relative to the project_, or _relative to the public directory_. Thru Vite and TypeScript, we have also supported aliased URLs.

**Do we need a resolution unification RFC?**

I would love to see Astro stick with web platform resolution, _except_ when aliasing is **explicitly** defined by the project. These would be:
- Import Maps (`astro.config.js#config.imports`)
- Defined dependencies (`package.json#dependencies`)
- TypeScript aliases (`tsconfig.json#compilerOptions.paths`, `jsconfig.json#compilerOptions.paths`)

jonathantneal:
After some discussion, we’ll merge this as-is and leave the details to the point of implementation.

altano:
I don't know if I'm too late to the party, but I believe this RFC doesn't address Markdown since there are no attributes to prefix with `local:`. Is that intended?

The main use case I'm interested in is having a colocated image referenced from Markdown, e.g.:
```
![Save](./save.png)
```

> I feel the same way. I believe others do, too. However, after many conversations I’m convinced we are, at the very least, not a super-majority. 😅

If you have a moment could you provide any pointers, @jonathantneal? I'm having a little trouble finding prior discussion of this and I'm having trouble imagining that anyone _likes_ the current behavior (especially in Markdown)?

jonathantneal:
I don’t have pointers, because I believe this RFC is insufficient for Markdown. I believe we should support `alt` and `` as-is. My opinions about resolution changed, but so did the RFC before it was accepted. I did not want `local:src` as an attribute name in the original proposal, which makes offering pointers for it particularly challenging. As for supporting ``, I am grateful that our community and users convinced me, and I hope changing my mind does not give you a lack of faith in me or the process.

To be specific, I think we should support `` and `alt` out of the box, **as long as scoping is restricted to the project, and root slash is relative to the project root**. Then, unmatched paths would ship as-is (_allowing them to match the structure in `public`, effectively_). For Markdown files, this means a `/docs/file.md` with either `Save` or `Save` or `Save` should _just work™_. I believe all 3 of those examples will currently work in GitHub, too.

And to highlight the restriction, I reject Vite’s current behavior where `/Users/your_name/GitHub/project/docs/file.html` can have `` and Vite is like "_yea sure no problem lets add that to the asset graph_" without requiring any special user permission/flag. I’m not sure if we can disable this, or if the change needs to happen upstream, but that is my only remaining hesitation doing things the 'Vite' way. 😅

altano:
> I am grateful that our community and users convinced me, and I hope changing my mind does not give you a lack of faith in me or the process.

Astro is great (especially given how young the project is) and I have nothing but faith in the process thus far :). I am a little unclear on what you were convinced of and what the future direction is. Is it the [merged rfc](https://github.com/withastro/rfcs/blob/021e99d6fb4b57061e9ea98c954f4a6d16136551/active-rfcs/0000-relative-url-scheme.md)?

> For Markdown files, this means a `/docs/file.md` with either `Save` or `Save` or `Save` should _just work™_. I believe all 3 of those examples will currently work in GitHub, too.

Yeah, that sounds good. I don't even particularly care how we get there or what it looks like exactly as long as the goal of colocation is achieved. Splitting resources between `public/` and `src/` (and replicating directory structure, and hard-coding paths that might break) just feels not super great. Although, I wouldn't even really consider it a dealbreaker.

FWIW the biggest head scratcher with the current behavior is how unpredictable the behavior is. This code:

~/src/pages/articles/my-article/index.md:
```astro
---
setup: |
  import saveReference from './save.png'
---

![Save](./save.png)

<img src={saveReference} />
```

Gives you:

| specification | `astro dev` | `astro build && astro preview` |
|---------------|-------------------|---------------------|
| `Save` | ❌ `Save` | ❌ `Save` |
| `` | ✅ `` | ❌ `` |

PR #63, opened 2022-01-05
Preserve specificity in scoped CSS

jonathantneal:
- Start Date: 2022-01-04
- Status: Draft

## Summary

This proposes Astro preserve specificity in scoped CSS, consistent with the specificity in global CSS.

Astro currently adds an additional class to all elements and selectors, increasing specificity from `O+1` to `O(1)+1`.

Astro can prevent this specificity increase by wrapping the scoping mechanism in a `:where()` CSS pseudo-class.

This RFC was previously submitted in [withastro/astro#1909](https://github.com/withastro/astro/issues/1909), but it was not reviewed before the process was changed. Neither this proposal or the current functionality were covered in the [Style Unification Proposal](https://github.com/withastro/rfcs/blob/main/active-rfcs/0000-style-unification.md). This proposal uses the new format, and is submitted to finalize this behavior.

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/jn.respect-authored-specificity/active-rfcs/0000-scoped-css-with-preserved-specificity.md)

### DIFF
--- a/active-rfcs/0000-scoped-css-with-preserved-specificity.md
+++ b/active-rfcs/0000-scoped-css-with-preserved-specificity.md
@@ -1,0 +1,146 @@
+- Start Date: 2022-01-04
+- Reference Issues: [#1909](https://github.com/withastro/astro/issues/1909), [#1762](https://github.com/withastro/astro/issues/1762)
+- Implementation PR: 
+
+
+
+# Summary
+
+Preserve specificity in scoped CSS, consistent with the specificity in global CSS.
+
+Astro currently adds an additional class to all elements and selectors, increasing specificity from `O+1` to `O(1)+1`.
+
+Astro can prevent this specificity increase by wrapping the scoping mechanism in a `:where()` CSS pseudo-class.
+
+
+
+# Example
+
+```astro
+<style>
+p {
+  color: blueviolet;
+}
+
+aside p {
+  color: darkviolet;
+}
+
+:global(aside) p {
+  text-decoration: underline;
+}
+</style>
+<p>This is a paragraph.</p>
+<aside>
+  <p>This is a paragraph in an aside.</p>
+</aside>
+```
+
+In this example, the HTML and CSS would generally transform into the following code:
+
+```astro
+<style>
+p:where(.astro-OZV3B5RX) {
+  color: blueviolet;
+}
+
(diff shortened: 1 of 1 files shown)


active-rfcs/0000-scoped-css-with-preserved-specificity.md
```diff
@@ -0,0 +1,146 @@
...
+/* Before (specificity of 0.1.1) */
+h1.astro-XXXXXXXX {}
+
+/* After (specificity of 0.0.1) */
+h1:where(.astro-XXXXXXXX) {}
+```
+
+
+
+# Drawbacks
```
matthewp, line 128: I think this needs some drawbacks to strength the case. There are definitely some. Some users might expect the styles they write within a component to "always win" and this would break that expectation.

active-rfcs/0000-scoped-css-with-preserved-specificity.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+
+# Motivation
+
+Astro component styles are scoped by default, meaning they only apply to items in the component.
+These are _exclusive_ styling boundaries that prevent styles from applying to items outside the component or within child components.
+
+This specificity increase is perceptually inconsistent to authors, without a deeper knowledge of the Astro internals and **Type**, **ID**, add **Class** selector specificity.
+This specificity increase makes it hard to combine with other CSS files or other styling libraries (Tailwind, CSS Modules, Styled Components, Stitches), as those systems would not include our additional side-effects.
```
matthewp, line 82: Can you provide some examples or where you might use those types of libraries in conjunction with scoped Astro CSS?  For example, I can definitely see where you are getting at with Tailwind, but within an example its unclear to me why you would expect Tailwind's CSS to take precedent.

matthewp:
We talked about this idea once before (in the old RFC process, if I recall), but the issue was some disagreement about whether component scoped CSS *should* have a lower specificity or not.

I think this proposal is missing the *why*, at present. Some more specific examples would strength the case, I think. One I can think of: if you wanted to implement low-level component primitives, like a `<Card />` component that is *intended* to be overridden, then this feature would be desirable.

Some side-by-side examples of how you might implement something with the current behavior and how that would (potentially) be easier with this proposed change, would strength the case and provide reasons for using `:where()`.

PR #72, opened 2022-01-13
Update file conventions

jonathantneal:
In an effort to improve our RFC process, this PR simplifies the current file structure.

1. Drops the `0000` prefix before proposals, because it was not used meaningfully.
2. Drops the `-rfc` suffix from the `active` directory, because it is redundant.
3. Introduces the `complete` directory for RFCs that have been accepted and implemented.
4. Removes the `.placeholder` file that instructed itself to be removed when items existed in `active`.


tony-sull:
Nice, big fan of this cleanup!

> Drops the 0000 prefix before proposals, because it was not used meaningfully

Was the original idea behind the `0000` to number the RFCs so the files get sorted by acceptance?  If so, I could see keeping that naming convention and updating the existing RFCs so they all have a unique prefix #

matthewp:
lgtm

natemoo-re:
> Was the original idea behind the `0000` to number the RFCs so the files get sorted by acceptance? If so, I could see keeping that naming convention and updating the existing RFCs so they all have a unique prefix

Yes, I think the point was to keep things organized and have a short-hand way to reference RFCs, like by RFC0215.

However, I’m not opposed to dropping it. It’s not super useful now. At a larger scale I can see the utility, though.

jasikpark:
I have yet to make an RFC of the new design, but I would argue that if we think having an RFC # will be good in the future, it's better to be consistent, rather than needing to add numbering later... https://github.com/npm/rfcs/blob/main/accepted/0038-workspaces-run-cmds.md seems like a good example where it's useful

I often find referencing issue #'s and pr #'s to be useful when discussing them, so maybe that could be kept?

PR #81, opened 2022-01-20
cleanup rfcs

boehs:
- removes `.placeholder`
- creates `landed-rfcs`
- moves `0004-support-draft-markdown-posts.md` into it


boehs:
GithHub vscode extension did something silly and did not verify this commit, however rest assured it is me

PR #106, opened 2022-02-09
Asset Primitive

FredKSchott:
- Start Date: 2022-02-08
- Status: Draft

## Summary

A consistent asset strategy and low-level primitive to handle assets (Icons, Fonts, Images, etc.) in Astro. Useful for users, integration authors, and to power future, more high-level features.

```astro
---
import imageBook from '../assets/book.png';
import iconCookie from '../assets/cookie.svg';
import robotoFont400 from '../assets/fonts/roboto-slab-v22-latin-regular.woff2';
import {Image, Icon, Font} from '../src/my-asset-components/index.js';
---
<html>
  <head>
    <Font use={robotoFont400} family="Roboto Slab" />
    <style>
      body { font-family: 'Roboto Slab'; }
    </style>
  </head>
  <body>
    <p>Hello, <strong>world!</strong></p>
    <Icon use={iconCookie} />
    <Image use={imageBook} />
  </body>
</html>
```

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/assets-rfc/proposals/0000-asset-primitive.md)

### DIFF
--- a/proposals/0000-asset-primitive.md
+++ b/proposals/0000-asset-primitive.md
@@ -49,7 +49,9 @@ Why is this is a problem? To be clear, it's not a problem *today* because run-ti
 
 However, with SSR support coming this will no longer always be true. If `<Icon name={fetchedAPIUserData.avatarIcon}>` renders for the first time on the server (ex: on Netlify, on Vercel) then Astro has no idea which icon you need to include in your build at build-time, which is when the decision of "what to include in your build" is made.
 
-This RFC is a priority to support our community members who are building components and integrations that use assets, and who need an asset strategy that we can commit to working in the future.  There have been different ideas thrown around to solve for this, some of which I will touch on in the "Alternatives" section of this RFC.
+Today, to get around this problem users can reach for Vite plugins which leverage build-time analysis for ESM imports. This solves our problem, but introduces a new one: no two Astro users have the same idea of what gets returned when an image is imported. For example, https://github.com/ElMassimo/vite-plugin-image-presets provides great features but defines its own API. This makes it impossible for 3rd-party component authors to assume a standard API and write consistent documentation if an asset reference or import is required.
+
+This RFC is a priority to support our community members who are building components and integrations that use assets, and who need an asset strategy that we can commit to working in the future.  There have been different ideas thrown around to solve for this, some of which I will touch on in the "Alternatives" section of this RFC. 
 
 ### Goals
 
@@ -61,7 +63,7 @@ This proposal outlines an **asset primitive** that accomplishes the following go
 - Works with SSR, where all assets must be known at build-time, with 100% accuracy.
 - Works across frameworks (`.jsx`, `.svelte`, `.vue`). Not an `.astro`-only solution.
 - Introduces an acceptable amount of boilerplate (as little as possible, or none).
-- Can power custom 3rd-party integrations and components.
+- Provide a consistent API for dealing with assets that 3rd-party components can safely build on top of.
 - Can power future, higher-level Astro core features.
 
 # Detailed design
--- a/proposals/0000-asset-primitive.md
+++ b/proposals/0000-asset-primitive.md
@@ -19,7 +19,7 @@ import {Image, Icon, Font} from 'astro/components';
 ---
 <html>
   <head>
-	  <Font use={robotoFont400} family="Roboto Slab" />
+    <Font use={robotoFont400} family="Roboto Slab" />
     <style>
       body { font-family: 'Roboto Slab'; }
     </style>
@@ -49,7 +49,7 @@ Astro.asset = function(name: string) {
 ---
 <html>
   <head>
-	  <Font use={Astro.asset('roboto-slab-v22-latin-regular')} family="Roboto Slab" />
+    <Font use={Astro.asset('roboto-slab-v22-latin-regular')} family="Roboto Slab" />
     <style>
       body { font-family: 'Roboto Slab'; }
     </style>
--- a/proposals/0000-asset-primitive.md
+++ b/proposals/0000-asset-primitive.md
@@ -1,6 +1,6 @@
 - Start Date: 2022-02-08
 - Reference Issues: https://github.com/withastro/rfcs/discussions/90
-- Implementation PR: TODO (proof of concept exists, still need to clean it up and get it into a PR)
+- Implementation PR: 
 
 # Summary
 
@@ -70,13 +70,25 @@ This proposal outlines an **asset primitive** that accomplishes the following go
 
 ## The Asset Primitive
 
+```ts
+interface ImageAsset {
+  url: string;
+  width: number;
+  height: number;
+  blurHashUrl: string;
+}
+
+interface SvgAsset {
+  width: number;
+  height: number;
+  content: string;
+}
+```
 ```js
 import imageBook from '../assets/book.png';
   // imageBook is type: ImageAsset (.png|.jpg|.jpeg|.webp)
-  // {url, width, height, blurHashUrl}
 import iconCookie from '../assets/cookie.svg';
-  // iconCookieis type: SvgAsset (.svg)
-  // {width, height, content}
+  // iconCookie is type: SvgAsset (.svg)
 import robotoFont400 from '../assets/fonts/roboto-slab-v22-latin-regular.woff2';
   // robotoFont400 is type: string
   // [If no custom asset is defined, Astro will default to including the asset
@@ -116,7 +128,7 @@ In the future, these can be extended to add new features. For example, the Image
 
 ## Other Notes
 
-- This is trivial to implement in Astro today. Vite provides hooks for these sorts of imports, so we would just need to create two Vite plugins inside of Astro: one for SVGs and one for images.
+- This is trivial to implement in Astro today. Vite provides hooks for these sorts of imports, so we would just need to create two Vite plugins inside of Astro: one for SVGs and one for images. See `ImageAsset` and `SvgAsset` types above for data interface.
 - This is possible to do in userland today by providing those two Vite plugins yourself. This proposal exists to bless this as the "Astro" way and integrate this into Astro itself so that our community can build on top of it.
 
 
--- a/proposals/0000-asset-primitive.md
+++ b/proposals/0000-asset-primitive.md
@@ -219,7 +219,7 @@ That RFC and this RFC both build on the same core idea: We need to pass assets t
 
 ## A smart HTML scanner
 
-@jonneal has proposed scanning Astro templates for assets directly as a way to avoid ESM import syntax. This is a much less known/certain path (very difficult to scan dynamic templates with 100% accuracy, how to handle `<img src="">` vs. `<Image src="">`, etc) that would also be non-trivial to implement. 
+@jonneal has proposed scanning Astro templates for assets directly as a way to avoid ESM import syntax. This is a much less known/certain path (very difficult to scan dynamic templates with 100% accuracy, how to handle `<img src="">` vs. `<Image src="">`, etc) that would also be non-trivial to implement (unlikely to achieve without delaying v1.0). 
 
 However, if we did want to explore this path there is a good chance that an internal Asset representation could still help to power this feature. For example, the compiler could compile the following to tell Vite that the asset must exist in the final build:
 
@@ -229,6 +229,8 @@ However, if we did want to explore this path there is a good chance that an inte
 // Equivilent to this:
 <img src={await import('../assets/book.png')} />
 ```
+
+This could be tackled as a followup to this RFC.
 
 # Adoption strategy
 
--- a/proposals/0000-asset-primitive.md
+++ b/proposals/0000-asset-primitive.md
@@ -15,7 +15,7 @@ A consistent strategy and low-level primitive to handle assets (Icons, Fonts, Im
 import imageBook from '../assets/book.png';
 import iconCookie from '../assets/cookie.svg';
 import robotoFont400 from '../assets/fonts/roboto-slab-v22-latin-regular.woff2';
-import {Image, Icon, Font} from 'astro/components';
+import {Image, Icon, Font} from '../src/my-asset-components/index.js';
 ---
 <html>
   <head>
@@ -32,36 +32,7 @@ import {Image, Icon, Font} from 'astro/components';
 </html>
 ```
 
-## High-Level Example (Future)
-
-**NOTE: This RFC is not proposing this API!** This is just an example to highlight one kind of high-level API that could be built on top of this primitive in the future.
-
-```astro
----
-import {Image, Icon, Font} from 'astro/components';
-
-/* Possible Astro.asset() Implementation (for illustration only)
-Astro.asset = function(name: string) {
-  const allAssets = import.meta.globEager('/assets/**');
-  const path = getPathFromName(name);
-  return allAssets[path];
-} */
----
-<html>
-  <head>
-    <Font use={Astro.asset('roboto-slab-v22-latin-regular')} family="Roboto Slab" />
-    <style>
-      body { font-family: 'Roboto Slab'; }
-    </style>
-  </head>
-  <body>
-    <p>Hello, <strong>world!</strong></p>
-    <Icon use={Astro.asset('book')} />
-    <Image use={Astro.asset('cookie')} />
-  </body>
-</html>
-```
-
+Several high-level APIs could be built on top of this primitive in the future. See **Alternatives** section below.
 
 
 # Motivation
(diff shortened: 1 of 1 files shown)


tony-sull:
Thanks for pulling together the details and examples for an RFC so quickly @FredKSchott!  Great to see we're getting ahead of the SSR challenges with asset imports with one recommended approach for different file assets :+1:

Would there be any issue with using other Vite plugins that also process imports for specific file extensions, like `vite-imagetools`?

Come to think of it, what would it look like if this was a standalone plugin (either for Astro or Vite) so this was completely opt-in?  Maybe even as a great first use case for a tool similar to `astro add @astrojs/images` 👀?

FredKSchott:
> Would there be any issue with using other Vite plugins that also process imports for specific file extensions, like vite-imagetools?

Great question! This actually moves us towards better 3rd party Vite plugin support, since all Vite plugins work through the Vite asset graph which is similar to what this RFC is proposing. There are some things inside of Astro that mean we can't support _all_ existing Vite plugins, but those are outside the scope / unrelated to this RFC.

> Come to think of it, what would it look like if this was a standalone plugin (either for Astro or Vite) so this was completely opt-in? Maybe even as a great first use case for a tool similar to astro add @astrojs/images 👀?

I think that our job is to create the primitive that all plugins & integrations could build on top of. So to build on this idea, the `ImageAsset` and `SvgAsset` concepts need to be built into Astro core for consistency for everyone to build on top of. BUT, anyone could build their own 3rd party `Icon`, `Font` and `Image` components if they'd like, using the same primitives.

proposals/0000-asset-primitive.md
```diff
@@ -0,0 +1,233 @@
...
+
+Why is this is a problem? To be clear, it's not a problem *today* because run-time and build-time are the same. In Astro today, **run-time** and **build-time** both happen during your build, ahead-of-time, before the site is deployed to the server. Runtime has full access to your machine, and every source file in your repo.
+
+However, with SSR support coming this will no longer always be true. If `<Icon name={fetchedAPIUserData.avatarIcon}>` renders for the first time on the server (ex: on Netlify, on Vercel) then Astro has no idea which icon you need to include in your build at build-time, which is when the decision of "what to include in your build" is made.
+
+This RFC is a priority to support our community members who are building components and integrations that use assets, and who need an asset strategy that we can commit to working in the future.  There have been different ideas thrown around to solve for this, some of which I will touch on in the "Alternatives" section of this RFC.
+
+### Goals
+
+This proposal outlines an **asset primitive** that accomplishes the following goals:
```
matthewp, line 85: What is the specific thing that you would call a primitive? I see two things in this proposal:

- Changing how different asset types are resolved through imports.
- High-level components.

The first is a primitive but the second isn't, imo. If the goal is to outline a primitive maybe the components aren't needed here.

proposals/0000-asset-primitive.md
```diff
@@ -0,0 +1,233 @@
...
+  // {width, height, content}
+import robotoFont400 from '../assets/fonts/roboto-slab-v22-latin-regular.woff2';
+  // robotoFont400 is type: string
+  // [If no custom asset is defined, Astro will default to including the asset
+  // in your build and returning a single URL as the result. This is Vite default behavior.]
+```
+
+- The asset primitive is imported via ESM.
+- Astro will define two custom `Asset` types: `ImageAsset` and `SvgAsset`
+- All other assets will just return a URL path string to the asset, and include that asset in the final build.
```
matthewp, line 115: Is there a definitive list of all things that could be called an asset?

This line says that:

> All other assets will just return a URL path

But it doesn't say what these 2 special asset types will return.

proposals/0000-asset-primitive.md
```diff
@@ -0,0 +1,233 @@
...
+
+This proposal outlines an **asset primitive** that accomplishes the following goals:
+
+- Works for Fonts, Icons, and Images.
+- Can reasonably extend to other asset types in the future.
+- Can reasonably extend to support image optimizations in the future.
+- Works with SSR, where all assets must be known at build-time, with 100% accuracy.
+- Works across frameworks (`.jsx`, `.svelte`, `.vue`). Not an `.astro`-only solution.
+- Introduces an acceptable amount of boilerplate (as little as possible, or none).
+- Can power custom 3rd-party integrations and components.
```
matthewp, line 93: How does this power 3rd-party (or even 1st party) integrations? I think that's what's missing most from this proposal.

The only thing from this proposal that really looks different to me from what can be done today is that you can import an SVG without using `?raw`. Is that the only change here, or am I missing something?

proposals/0000-asset-primitive.md
```diff
@@ -0,0 +1,233 @@
...
+- Astro will ship `Font`, `Icon`, and `Image` components that can take these asset types, and properly use them on the page:
+  - `Font`: Creates a `<style>` tag with a single `@font-family` definition
+  - `Image`: Creates a `<img>` tag with `src=` the asset URL
+  - `Icon`: renders the given `svg` content to HTML.
+- In the future, these can be extended to add new features. For example, the Image component can be extended to support image optimizations and automatic [blurhash](https://blurha.sh/) support.
+- Renders could also provide `Font`, `Icon` and `Image` components that are specific to each framework. This can be fleshed out in more detail if it is a concern, but it is strongly believed that this is possible and fairly trivial.
+
+## Other Notes
+
+- This is trivial to implement in Astro today. Vite provides hooks for these sorts of imports, so we would just need to create two Vite plugins inside of Astro: one for SVGs and one for images.
```
matthewp, line 148: What would these plugins do, specifically? I'm assuming the SVG one returns SVG as raw. If that's the case I think it should be explicit in the proposal. It's unclear to me what the image plugin is intended to do.

proposals/0000-asset-primitive.md
```diff
@@ -0,0 +1,233 @@
...
+  <Image use={imageBook} />
+</body>
+</html>
+```
+
+- Astro will ship `Font`, `Icon`, and `Image` components that can take these asset types, and properly use them on the page:
+  - `Font`: Creates a `<style>` tag with a single `@font-family` definition
+  - `Image`: Creates a `<img>` tag with `src=` the asset URL
+  - `Icon`: renders the given `svg` content to HTML.
+- In the future, these can be extended to add new features. For example, the Image component can be extended to support image optimizations and automatic [blurhash](https://blurha.sh/) support.
```
matthewp, line 143: I think this is the most important part of figuring out assets. I don't think these components *can* be extended to do those kinds of things, as there are missing primitives around transforming files either during the build or at runtime.

If you want to make this a future proposal I think that's fine, but in that case I don't think the builtin components should be added until that part is ready. A 3rd party should have the same ability to do transformations as a builtin component and I think shipping builtins before that sends a signal to the community that the builtins are going to get special APIs that they won't have access to.

We made this mistake with the Markdown component (giving special APIs) and haven't yet been able to undo it. Once something is implemented high-level there is less of an incentive to ever pull out the primitives, so I'd be disappointed if we did this again with asset components.

proposals/0000-asset-primitive.md
```diff
@@ -0,0 +1,233 @@
...
+We already have an accepted RFC for the `local:src` and `local:href` directives, which is sugar for the following:
+
+```
+// Sugar:
+<img local:src="../assets/book.png" />
+// Equivilent to this:
+<img src={await import('../assets/book.png?url')} />
+```
+
+That RFC and this RFC both build on the same core idea: We need to pass assets through Vite for them to be included in the final build. You could imagine that we could extend this `local:src` syntax to handle other assets, building on this same primitive:
```
matthewp, line 217: I don't believe `local:` is scoped to only src and href at the moment. The accepted RFC doesn't mention that: https://github.com/withastro/rfcs/blob/main/proposals/0011-relative-url-scheme.md

My understanding is that the right-hand side is not special-cased in any way, so I would expect a `local:use` to work once that proposal is completed.

matthewp:
I added comments but will summarize my feedback here:

- I think this proposal is mostly about changing how certain file types are resolved, specifically `.svg` and perhaps `.png`, `.jpg`, and others. But the specifics of what will change are missing, so I think those should be given full detail.
- I think it's premature to add builtins that don't do anything extra until the primitives exist that would allow them to do the extra things we might want them to do (like image optimizations).
- I think a big part of this proposal is around the opinion that inlining SVGs is the right thing to do for icons. If this is the case, can you add links to articles that make the case that they should be inlined? I think having some "best practices" research to accompany the proposal would strengthen it.
- What about non-icon SVGs? How does this proposal account for adding an SVG that's just an image? Would you pass it to `<Image>` (i don't think so given the explanation of how Image is set to currently work).

proposals/0000-asset-primitive.md
FredKSchott: Thanks @matthewp, I agree. Removing the components for now to keep scope focused

proposals/0000-asset-primitive.md
FredKSchott: clarified

proposals/0000-asset-primitive.md
FredKSchott: good catch, clarified

proposals/0000-asset-primitive.md
FredKSchott: Clarified that this info exists in the code snippet at the top of the section

FredKSchott:
Thanks for the great feedback @matthewp, this is super helpful! I'm still going through it to make sure everything is addressed.

proposals/0000-asset-primitive.md
FredKSchott: Clarified to be more clear. The "powering" of 3rd party integration was in reference to the fact that this RFC lets an integration author assume a common data structure for certain assets.

For example, if you are build an image component today, Astro provides no standard idea of what an image is. Today, we leave it to the integration to override and dictate what happens when you import an image in an Astro project. For example: https://github.com/ElMassimo/vite-plugin-image-presets

proposals/0000-asset-primitive.md
FredKSchott: The data type is highlighted above, but I'll add a note here to make it more clear

proposals/0000-asset-primitive.md
FredKSchott: I think I'm aligned on this. The built-in components were included to make it clear how you would use these assets, but I can illustrate this without actually proposing that we build them before we are ready.

jonathantneal:
I have 2 bits of thought and a lot of suggestions and concerns. They stem from the purpose of this proposal, which I read to be a "strategy and low-level primitive to handle assets". I’ll include suggestions as often as possible.

Bit of thought 1: The platform offers us great APIs for adding assets:

```html
<style>
@font-face {
  font-family: "Comforter";
  src: url("/src/fonts/Comforter.woff2");
}
</style>
```

```html
<link rel="preload" href="/src/fonts/Comforter.woff2" as="font" />
<link rel="stylesheet" href="./relative-to-component-style.css" />
<script src="./relative-to-component-client-script.js"></script>
```

Bit of thought 2. We are supposed to have a powerful API — [`this.emitFile`](https://rollupjs.org/guide/en/#thisemitfile) — for emitting chunks and assets behind the scenes.

These existing foundations should work, and this RFC should build upon them before attempting to replace them.

We have page authors who would like the former bit to work, and we have plugin authors who would like the later bit to work.

Regarding page authors. My own mind has come around a lot on relative paths in AstroHTML files, especially after watching and listening to users try to use AstroHTML files alongside assets. They intuit `` and `<link href>` as resolving source-relative first, otherwise as-is (meaning it will look up public). This proposal moves further from that intuition and closer to NextJS. Leaning on HTML and CSS is part of what makes Astro a better alternative.

Regarding plugin authors. This solution should embrace a plugin API for attaching assets. This proposal makes no mention of the existing 'Vite way'. I think Astro should embrace and pay closer attention to these plugin primitives. Here are two reasons:

1. Because `this.emitFile` was difficult to work with in Astro before, and is now extremely difficult or impossible to work with behind the experimental static build.
2. Because dog-fooding is important. For instance, `import` still embraces Vite’s pseudo-relative imports, which allows me to do this:
  ```html
  <script type="module">
  import oh_was_that_private from '/etc/hosts?raw'

  console.log(oh_was_that_private)
  </script>
  ```

I suggest we get these issues sorted out first, because I believe they would provide most of what this proposal seeks to provide.

P.S. I really like some of the ideas in this proposal, and I am going to write a separate comment about all of the reasons I like them and how I would like to see them play out. However, none of those reasons are related the goal of providing a API so that people _can_ use assets at all.

Holben888:
> I added a note about the fact that this RFC intentionally prioritizes the inline SVG use-case. Someone else may be able to provide better resources, but here are two that I found useful: https://css-tricks.com/svg-loader-a-different-way-to-work-with-external-svg/ (gets at why usability is better) & https://salience.co.uk/insight/magazine/inline-svg-page-speed-seo/#why-use (gets at why perf is better).

@FredKSchott Understandable! I spent too long side-eyeing the `astro/components` note in particular. Sounds like this RFC is just establishing 1 thing: what the heck to SVGs, PNGs, fonts, etc look like when imported via ESM? I'd love for Astro to define all of these out-of-the-box instead of using scattered Vite plugins.

In fact, could these primitives all be one Vite plugin that I can add and remove from my Astro config? 

```js
import astroAssetsPlugin from 'astro/config'

export default /** @type {import('astro').AstroUserConfig} */ ({
  ...
  vite: {
    plugins: [astroAssetsPlugin()],
  },
});
```

☝️ Might be worth auto-applying for Astro projects specifically. But if this were all bundled into a Vite plugin, it could be used across the open source ecosystem as well 😁

Holben888:
And _yes yes_ I know `astro/components` is out of scope, but I did want to mention: wouldn't inline SVGs in particular be better served by the `render` element I've heard floating around?

I like the idea of a `<render:md>` replacing the existing `<Markdown>` component. Maybe something similar could be done for SVGs?

```html
<render:svg local:src="/path/to/vector.svg" />
```

I could even imagine custom renderers to handle complex tasks. For instance, image optimization:

```html
<render:picture local:src="/path/to/unoptimized.png" />
```

Am I bullying `render` into a sort of [11ty shortcode?](https://www.11ty.dev/docs/shortcodes/)? Maybe. Is this still out of scope for this RFC? Absolutely. But I'm still hitting the "comment" button.

jonathantneal:
We can already statically analyze CSS, so we don’t _need_ a fonts component to add a `@font-face` rule. It might not even be the most intuitive way for authors to automatically attach fonts.

```html
<style>
body {
  background-color: green;
  color: white;
  font: 200%/1.5 'Comforter', monospace;
  margin: 5%;
}
</style>
```

That could be enough. [Try it](https://stackblitz.com/edit/github-xduwnr?file=demo%2Fsrc%2Fpages%2Findex.astro).

---

It would be nice to have a low-level API for adding Assets in plugins, and for telling Astro when it gets to statically inject things into `<head>`.

```js
// PostCSS Plugin with its own `getFontFamilyName`, `getPathToFont`, & `generateFontFaceRule` methods.
{
  Declaration(decl) {
    if (isFontDeclaration(decl.prop)) {
      const fontFamily = getFontFamilyName(decl.value)
      const src = getPathToFont(fontFamily)
      const fontFaceRule = generateFontFaceRule({ fontFamily, src })

      low_level_asset_api.add(fontSrc, { injectCSS: fontFaceRule })
    }
  }
}
```

With something like `injectCSS`, then we can override the current behavior in Astro.
```js
import style from './style.css'

console.log(style) // prints the CSS and changes the page HTML, either of which may not be intuitive to everyone
```

jonathantneal:
For HTML, if we can support `` while knowing how React _reacts_ to `class`, then we should support `` while knowing React JSX files would also handle this differently. No namespace attributes necessary.

Similarly, we should support `friend`. In fact, if we limited this behavior to the project directory (_[as I mentioned previously](https://github.com/withastro/rfcs/pull/106#issuecomment-1039704770)_ then writing `friend` would work the same in Markdown files rendered in both Astro or GitHub.

jonathantneal:
What I love about this proposal is the idea that `import` would return objects representing their entries, rather than as strings of the import path. While the current proposal restricts objects in `woff2`, I would suggest rather that this work _consistently_ across media assets.

Since this RFC does not include a detailed design of `ImageAsset`, I propose the RFC be updated to return objects for all asset types as follows:

```js
interface MediaEntry {
  /** Last modified time of the file in milliseconds. */
  readonly lastModified: string

  /** URL of the file as an href. */
  readonly url: string

  /** Path of the file in POSIX. */
  readonly fullPath: string

  /** Subresource integrity value of the file. */
  readonly integrity: string

  /** Size of the file in bytes. */
  readonly size: number

  /** Media type of the file. If the type is unknown, this string is empty. */
  readonly type: string

  /** Returns a promise that resolves with an ArrayBuffer representation of the file. */
  arrayBuffer(): Promise<ArrayBuffer>

  /** Returns a ReadableStream that upon reading returns the data contained within the file. */
  stream(): ReadableStream

  /** Returns a promise that resolves with a text representation of the file. */
  text(): Promise<string>

  /** Returns the path of the file in POSIX. */
  toString(): string
}

interface ImageModuleEntry extends MediaEntry {
  /** Number of pixels per row in the image. */
  readonly width: number

  /** Number of pixels per column in the image. */
  readonly height: number
}

interface FontModuleEntry extends MediaEntry {
  /** Family name of the font. */
  readonly family: string
}
```

FredKSchott:
**RFC Call:** Thanks everyone for the feedback! Some notes to make sure that history is recorded:

- We reached formal consensus on the idea of using ESM imports as a consistent API for asset references across Astro files, React files, Svelte, etc.
- We did not reach consensus on what the thing returned by that API should look like. More work needed to finalize this part, get involved in the `#assets` channel on Discord if you are interested in participating!

jonathantneal:
We agreed that, in any kind of JS (which includes JS files, JS derivative files like JSX or TS, etc, and any blocks of JS like the front-matter and expressions of Astro files), it is acceptable to import an asset.

We did not attempt consensus regarding what an asset specifically is, but it was well understood to be images and fonts, with PNG, SVG, and WOFF files specifically called out as examples.

We agreed there would be no limits on the path that an asset could be imported from at this time.

It was noted that this is and has been the behavior in Astro. This behavior is documented.

There were issues raised about why the proposal was limiting itself to JS files, as many individuals did not want to see Astro prioritize a JS-first experience.

We did not attempt consensus on what the import would return.

There was informal consensus that imports should probably return the same (not yet defined) thing, whether they were imports of relative assets, imports of external package assets, or imports from within external packages. However, there were also concerns raised about this behavior and how it may collide with packages relying on different import behavior.

The author intends to prototype and continue the discussion here, and to reach further consensus again at a later date.

FredKSchott:
Closing this as something that I don't plan to pick back up personally until after v1.0. As of right now we are relying on Vite defaults to power our ESM import statements.

I would love to see more user-land exploration around an `assets/` folder and how that could work (see the proposal for more suggestions on this)

PR #147, opened 2022-03-10
New Markdown API

FredKSchott:
- Start Date: 03-10-2022
- Status: Draft

## Summary

A v2.0 API for working with local Markdown files in Astro, to solve issues around performance and infinite loops.

```astro
---
// BASIC USAGE:
// Astro.fetchContent() replaced with a generalized `Astro.glob()` helper.
// This now works for non-MD files as well!
const markdownFilesArr = await Astro.glob('../posts/*.md');
const markdownFilesFiltered = markdownFilesArr.filter(post => post.frontmatter.category === 'blog');

// BASIC USAGE:
// Individual markdown imports are now supported as well!
const firstPost = await import('../posts/first-post.md');

// ADVANCED USAGE:
// You can also use the bare-metal `import.meta.glob/globEager` API
// This is useful in non-Astro files, like JSX and Vue.
const markdownFilesObj = import.meta.globEager('../posts/*.md');
const markdownFilesArr = Object.values(markdownFilesObj);
const markdownFilesFiltered = markdownFilesArr.filter(post => post.data.category === 'blog');
---
<!-- Data is parsed from frontmatter without requiring a render -->
<h1>{firstPost.frontmatter.title}</h1>
<p>Author: {firstPost.frontmatter.author}</p>
<p><a href={firstPost.url}>Permalink</a></p>
<!-- Defer rendering as late as possible for performance -->
<firstPost.Content />
```

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/fetch-content-2/proposals/0000-markdown-content-redesign.md)

cc @natemoo-re @tony-sull @rebelchris


proposals/0000-markdown-content-redesign.md
```diff
@@ -0,0 +1,126 @@
...
+A v2.0 API for working with local Markdown files in Astro.
+
+# Example
+
+```astro
+---
+// BASIC USAGE:
+
+// Note: Astro.fetchContent() removed in favor of direct `import.meta.globEager` usage
+// Note: This works in non-Astro framework components!
```
tony-sull, line 16: This is a big plus, and honestly I could see this alone being a great reason to move away from `Astro.fetchContent()` 🎉

proposals/0000-markdown-content-redesign.md
```diff
@@ -0,0 +1,126 @@
...
+const markdownFilesObj = import.meta.globEager('../posts/*.md');
+const markdownFilesArr = Object.values(markdownFilesObj);
+const markdownFilesFiltered = markdownFilesArr.filter(post => post.data.category === 'blog');
+
+// Note: Individual markdown imports are now supported as well.
+const firstPost = await import('../posts/first-post.md');
```
tony-sull, line 22: This is a very cool feature, it's nice being able to just use `await import()` here rather than using an API designed for loading multiple files with a glob pattern

proposals/0000-markdown-content-redesign.md
```diff
@@ -0,0 +1,126 @@
...
+- All Markdown files (ex: `src/posts/foo.md`) are resolved and loaded by this plugin.
+
+1. `src/posts/foo.md?import`
+    1. loaded from file system
+    2. frontmatter is parsed from the file
+    3. a JS module is returned with the following JS:
+    ```js
+    `
+    // Static:
+    export const data = ${JSON.stringify(frontmatter)};
```
tony-sull, line 52: 🚲 since this is specific to markdown files, `frontmatter` is a little more clear for me than `data`

🚲 === we need a bike shedding emoji

proposals/0000-markdown-content-redesign.md
```diff
@@ -0,0 +1,126 @@
...
+# Motivation
+
+- **Performance:** Markdown rendering continues to be an expensive part of Astro's runtime. While we can look into faster renderers (ex: Goldmark) there is still a big inefficincy on our end with how we hold the tool. Astro currently renders markdown on import, which means that multi-file markdown imports (via both `import.meta.glob` and `Astro.fetchContent()`) block the response until all imported files are rendered, even if only a subset of these files are ever used on the page after filtering.
+- **Performance (Memory):** For large projects, this also forces Astro to store all rendered Markdown output in memory at the same time, making it difficult to manage memory efficiently. Our community has reported builds maxing out memory limits in a project of only 800 markdown pages, and builds requiring 50+ GB of memory.
+- **Infinite Loops:** By rendering during import, we also introduce a risk for infinite loops when we call `fetchContent()` in a Markdown layout used by a fetched page, bouncing between render and `fetchContent()` infinitely.
+
+
+# Detailed design
+
+- This is implemented in a v2.0 of the internal `'astro:markdown'` Vite plugin
```
tony-sull, line 42: Would it be possible to actually pull this out of astro core and making it an opt-in integration/plugin?

It may not actually make sense since Astro needs to handle markdown anyway, but requiring a Vite plugin that always claims ownership over every `.md` file import would block any third-party `.md` vite/rollup plugins

aFuzzyBear:
I totally welcome this as its the approach that I have gone for myself, over the fetch content, however I have found that for some reason, post `0.23` onwards this pattern doesnt allow for obtaining files located elsewhere within the scope of the project, ie in another folder somewhere within the monorepo, its partly the reason why I cant upgrade atm to the later releases,

proposals/0000-markdown-content-redesign.md
```diff
@@ -0,0 +1,126 @@
...
+
+# Example
+
+```astro
+---
+// BASIC USAGE:
+
+// Note: Astro.fetchContent() removed in favor of direct `import.meta.globEager` usage
+// Note: This works in non-Astro framework components!
+const markdownFilesObj = import.meta.globEager('../posts/*.md');
```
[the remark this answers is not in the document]
bholmesdev: So this is true, but I always found it odd how `fetchContent` was restricted to markdown alone. `import.meta.globEager` is a pretty convenient way to see all available routes for a directory in a mix of `.astro` or `.md` formats. Serves as an [11ty collections alternative](https://www.11ty.dev/docs/collections/) in a way.

Because of that, I'd love to start pushing users towards this more generic API so they don't feel restricted. Maintaining both `fetchContent` and `import.meta.glob` feels a bit odd to me. As a user, I'd worry that `globEager` is less performant or optimized compared to `fetchContent` since anything starting with `Astro.` feels like the right way :tm:.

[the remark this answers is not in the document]
FredKSchott: Thanks for the feedback yall! Based on this convo I think I have a new approach that keeps the idea of a helper (so that users don't need to mess with `import.meta.glob/globEager` for basic use) but makes that helper for generally useful (supports non-MD files).

Still updating the RFC, but you can see the new API in the example.

proposals/0000-markdown-content-redesign.md
FredKSchott: 👍 In updating the examples, I realized that `.data` is used in our pagination API. Not a huge deal, but based on your feedback + my experience migrating that file and adding a ton of `.data` everywhere, I'm +1 on `frontmatter` now as well.

proposals/0000-markdown-content-redesign.md
FredKSchott: It would be possible in the new integration system, but I think that would require a really easy-to-use `astro add markdown` CLI command. Otherwise that's a bit too much overhead to add for such an important use-case. 

Also curious to hear others thoughts on this question, it's an interesting one!

FredKSchott:
@aFuzzyBear RFC updated based on this feedback! See comment here: https://github.com/withastro/rfcs/pull/147#discussion_r827483369

proposals/0000-markdown-content-redesign.md
```diff
@@ -0,0 +1,152 @@
...
+    export const file = ${JSON.stringify(id)};
+    export const url = ${JSON.stringify(url || undefined)};
+
+    // Deferred:
+    export default async function load(...args) {
+        return (await import(${JSON.stringify(fileId + '?content')}));
+    };
+    export function getContent() {
+        return load().then((m: any) => m.default)
+    }
```
natemoo-re, line 103: I think it would be preferred if the API just exposed the component itself rather than needing to call a getter and render it in an expression.

```suggestion
    export function Content(...args: any) {
        return load().then((m: any) => m.default(...args))
    }
    Content.isAstroComponentFactory = true;
```

## Usage

**Before**
```astro
---
const firstPost = await import('../posts/first-post.md');
---
<!-- Data is parsed from frontmatter without requiring a render -->
<h1>{firstPost.frontmatter.title}</h1>
<p>Author: {firstPost.frontmatter.author}</p>
<p><a href={firstPost.url}>Permalink</a></p>
<!-- Defer rendering as late as possible for performance -->
<article>{firstPost.getContent()}</article>
```

**After**
```astro
---
import { Content, frontmatter, file, url } from "../posts/first-post.md";
---
<!-- Data is parsed from frontmatter without requiring a render -->
<h1>{frontmatter.title}</h1>
<p>Author: {frontmatter.author}</p>
<p><a href={url}>Permalink</a></p>
<!-- Still defers rendering as late as possible for performance, but the user doesn't need to worry about that -->
<article><Content /></article>
```

natemoo-re: Basically, I think we can delay the rendering for performance reasons without passing that cost onto the user. Astro will `await` the component before attempting to render it, which amounts to automatic unwrapping.

A component-based API makes more sense because `firstPost.getContent()` wouldn't be inspectable like users might think.

FredKSchott: Ah yea, I'm game for this 👍 That feels much more natural

FredKSchott: Well, actually with static ESM imports that feels more natural, but in all other cases I'm not sure it does. Look what it does to the example we already have:

```astro
---
const firstPost = await import('../posts/first-post.md');
---
<firstPost.Content />
```

or, for the collection/glob use-case:

```astro
---
const posts = await Astro.glob('../posts/first-post.md');
---
{posts.map(post => <post.Content />)}
```

We do support this syntax, but we don't push users to use it very often. Are we okay with that? 

Also, does `<Content />` give us the ability to pass props to the layout that the markdown page uses? That could be a deciding factor here, if we can think of a use-case there that we want to support.

bholmesdev: I think this is fine, since you can destructure `{ Content }` quite easily in the `.map` example. I was also wondering about props as well. I don't think this would be possible, since this would change frontmatter from a compile-only step to a render step 😬

FredKSchott: Accepted suggestion into RFC!

FredKSchott:
Honestly that may be a bug! Or, intended behavior? I'm not sure, but I think that's separate from this RFC. Feel free to ask as an issue on GH or in discord!

FredKSchott:

mash-graz:
it's really nice, that all this fetchContent()-hell slowly gets a little bit improved.

i already used some  import.meta.glob/globEager workarounds in my own projects to compensate its limitations.

one of most annoying aspects of this kind of access mechanism could be seen in fact, that you can just match "files" but not directories and globe() also doesn't accept simple variables to compose the search pattern but just literal strings. 

if you want to render a nested hierarchy of folder containing .md-content, you always have to scan over all of them or at least huge sub-trees and process the included path information by filtering all returned entries, because you are not able to simply narrow the focus by analyzing the directory hierarchy first and thereby reduce the io- and compute-amount.

PR #150, opened 2022-03-15
Style/Script Defaults

natemoo-re:
- Start Date: 2022-03-15
- Status: ✅ Accepted

## Summary

Astro is inconsist between `<style>` and `<script>` default behavior. Currently, `<style>` has opt-out processing, but `<script>` does not. This RFC aims to settle on a good, consistent default for both `<style>` and `<script>`.

TLDR; `<script hoist>` becomes the default for `<script>`. `is:inline` introduced for `<script>` and `<style>`.

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/main/proposals/0016-style-script-defaults.md)


proposals/0000-style-script-defaults.md
```diff
@@ -0,0 +1,88 @@
...
+
+### `is:scoped`, `is:global`
+
+These new directives are for `<style>` tags. They represent scoped (the default) and global (previously `global`) styles.
+
+### `hoist` by default
+
+Support for `<script hoist>` will be removed. This will become the default `<script>` behavior, where `type="module"` is implied and scripts are preprocessed and bundled by default.
+
+If any script tag has an attribute (ex: `src=`, `type="module"`, `async`) or directive (`define:vars`), it is implied to be `is:inline`. Astro should emit a warning that `is:inline` is needed for clarity.
```
matthewp, line 61: The `src=` caveat here is slightly different from the current hoist behavior in that you can have external hoisted scripts. I don't think there's a technical reason to have this restriction. Is there another reason?

For clarity, you could still do this:

```js
<script>
  import "../scripts/external.js";
</script>
```

I think this is ok and aligns with how `style` currently works.

proposals/0000-style-script-defaults.md
```diff
@@ -0,0 +1,88 @@
+- Start Date: 2022-03-15
+- Reference Issues: [RFC0008](https://github.com/withastro/rfcs/tree/main/proposals/0008-style-script-behavior.md), [#65](https://github.com/withastro/rfcs/discussions/65)
+- Implementation PR: <!-- leave empty -->
+
+# Summary
+
+Astro is inconsist between `<style>` and `<script>` default behavior. Currently, `<style>` has opt-out processing, but `<script>` does not. This RFC aims to settle on a good, consistent default for both `<style>` and `<script>`.
+
+- New `is:inline` directive to avoid bundling `style` or `script`
+- `<style>` => remains scoped, bundled by default. `is:scoped` directive supported for consistency
```
matthewp, line 10: Are we preserving the new behavior added by @FredKSchott that imports inside of style are not scoped? What happens if you add `is:scoped` explicitly, those are still not scoped?

sciencefidelity:
I found a use case for this. In the last few days I've been experimenting with creating html email templates that populate and compile from a CMS (Sanity). For this use case all styles need to be inlined, I'd prefer to use `lang:scss`.

I have some questions:

- Do `is:inline` styles get optimised?
- Where on the complied page do styles get inlined, (ie are they be moved to `head`)?
- What happens when multiple `.astro` components containing `<style is:inline>` compile to a single `.html` file?

matthewp:
@RafidMuhymin `define:vars` already isn't supported for hoisting. We would like to support it but just haven't figured out how to do so. We need to build the JavaScript before rendering, but then some how feed the rendered variables back into it. It's a hard problem.

proposals/0000-style-script-defaults.md
tony-sull, line 10: Since scoping is the default, removing the `is:scoped` directive may avoid some confusion

Not a blocker for me at all, there is value in the option for making it explicit so I could go either way with it

tony-sull (commented):
Added one comment related to potentially removing `is:scoped`, but not a blocker for me

Very happy to see how all this came together!  It makes much more sense having styles and scripts handled in the same way, and defaulting to compiler/bundler magic makes sense since that seems to be the lion's share of real world use cases

VodnikAngus:
> @RafidMuhymin `define:vars` already isn't supported for hoisting. We would like to support it but just haven't figured out how to do so. We need to build the JavaScript before rendering, but then some how feed the rendered variables back into it. It's a hard problem.

I know that currently `define:vars` works by adding `let variable = value`, but that sounds wrong to me. Could Astro move to more of like a preprocessor way of replacing every occurrence of variable with it's value? or at least using `const` insted of `let`?

proposals/0000-style-script-defaults.md
natemoo-re: Updated to remove `src`!

proposals/0000-style-script-defaults.md
natemoo-re: No change here, so yes that behavior is preserved.

FredKSchott:

PR #153, opened 2022-03-21
Astro.request RFC

matthewp:
- Start Date: 2022-03-31
- Status: Draft

## Summary

Make `Astro.request` a [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) to enable SSR use cases (checking for cookies, for example).

```astro
---
const cookie = Astro.request.headers.get('cookie');
const loggedIn = parse(cookie).loggedIn;
---
<html>
<head>
  <title>My Blog</title>
</head>
<body>
  Hello, you are {!loggedIn && 'not'} logged in.
</body>
</html>
```

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/astro-request/proposals/astro-request.md)
- [Previous RFC discussion](https://github.com/withastro/rfcs/discussions/151)


aFuzzyBear:
I am really loving the new SSR release, its really fantastic, well implemented, 
Given the RFC, my only slight concern is the rejiggin of the `Astro` Object to try and accommodate new additions.

I think it might be prudent to see if this discussion can increase the scope a bit by suggesting perhaps a second Global object that is reserved for SSR practices, 

This would have the `Server.request` and `Server.response` objects which would allow for the DX to be augmented with the SSR approach instead of bending to the SSR approach.

The downside with this is increasing the language server support to accommodate the new Global object, I do feel that this extra project cost would be beneficial in the longer term, as it would provide two distinct and clearly defined Globals for us to utilize more freely, and can reduce the cognitive overhead,

matthewp:
Thanks @aFuzzyBear. Putting aside a separate global or not, one reason not to have a separate "server" object is that we want to make SSR and SSG modes as close as possible, to prevent there from being a completely different programming model.

That's what's nice about `Astro.request`, imo, it already exists and is useful even if you are not using SSR. Because some times you need a page's URL and `Astro.request.url` is there for you.

SSR is just regular Astro with a couple of extra features, is how I think of it.

aFuzzyBear:
Ah, that makes more sense from that perspective @matthewp 

Thanks man ☺️

proposals/astro-request.md
```diff
@@ -0,0 +1,110 @@
...
+}
+```
+
+## SSR vs. SSG
+
+When designing the initial version of Astro we intentionally omited things like query parameters and headers from the `Astro.*` APIs to prevent users from depending on things that would not work in production (since they are static pages).
+
+We should continue doing so, which just means:
+
+- In development mode when not using SSR, the `URL` is made to not contain query parameters.
```
FredKSchott, line 87: love these two callouts, glad you're thinking through the static site user

proposals/astro-request.md
```diff
@@ -0,0 +1,110 @@
...
+This change will move `canonicalURL` and `params` up to the `Astro` object and make `request` a Request.
+
+```typescript
+// Partial
+interface Astro {
+    /** get the current canonical URL */
+  canonicalURL: URL;
+
+  /** get page params (dynamic pages only) */
+  params: Params;
```
FredKSchott, line 72: You mentioned below that `Astro.request` will have all query/search params stripped in SSG mode, but I can imagine I'd be confused if `Astro.params` is still populated and the two don't match. 

Instead of stripping these out in SSG, should we just clean them up to match only the params supported by the file-based route? We already implement this internally, so it wouldn't be any more work (I don't think, at least).

Or, if we want to go full purist/breakage, we could remove Astro.params entirely in favor of `Astro.request.searchParams` in the SSG use-case. That feels a bit more less user-friendly though, since `searchParams` isn't the most well-known API.

proposals/astro-request.md
```diff
@@ -0,0 +1,110 @@
...
+- Keeping the existing object but adding a `headers` property that is a [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) (Map-like) object.
+- Making Astro.request be a Request but then also adding the `params` and `canonicalURL` properties.
+  - Feel that doing it this way makes it harder to document and the code becomes slightly more complex.
+
+Either of these options would be *fine*, but if we were designing Astro from the start with SSR in mind we would probably have made it a Request, so doing so now before 1.0 seems like good timing.
+
+# Adoption strategy
+
+- This is a breaking change that would go out before (or during) 1.0.
+- Docs would be updated to reflect that `Astro.request.url` is now a string.
```
FredKSchott, line 110: To help with migration, could we create a simple getter on `Astro.request.canonicalURL` and `Astro.request.params` that give a warning/error?

FredKSchott (approved):
Left a few comments, but overall +1 on this change to update our existing custom API to something a bit more standard.

proposals/astro-request.md
matthewp: Astro.params is the file based params, it's not related to search/query parameters. I can't imagine being confused by this, as this is currently how it works already.

proposals/astro-request.md
matthewp: I added this as part of the adoption strategy.

FredKSchott:

PR #154, opened 2022-03-22
Config Finalization

natemoo-re:
- Start Date: 2022-03-22
- Status: Draft

## Summary

Astro's configuration file is something we'd like to clean up prior to a v1.0 Beta release. Currently, it is not evident which specific options should go in which area, leading to user confusion and occasional duplication.

This RFC proposes a clean-up of the existing Astro config file format for consistency and clarity.

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/config-audit/proposals/0000-config-finalization.md)


proposals/0000-config-finalization.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Markdown
+
+The `markdownOptions` is currently one of the most confusing options in the Astro config.
+This RFC proposes:
+
+- Rename `markdownOptions` => `markdown`
+- Remove `render`
+- Remove references to `@astrojs/markdown-remark` in `render[0]`. Remark is the built-in Markdown integration—it will not be pluggable in v1.0.
+  > Internally this may be implemented in a pluggable manner, but that's an implementation detail.
```
natemoo-re, line 122: Worth calling this change in direction out. I know @FredKSchott and I have discussed leaving open the possibility of pluggable Markdown renderers, but I don't think we need to complicate our config file for a future goal if Remark is still going to be the first-class integration in v1.0.

proposals/0000-config-finalization.md
```diff
@@ -0,0 +1,160 @@
...
+        host: false,
+        port: 8080
+    },
+
+    markdown: {
+        drafts: true,
+        syntaxHighlight: 'shiki',
+    },
+
+    sitemap: {
```
FredKSchott, line 53: apologies if this is contradicting anything I said last week, but looking at this proposal now I think we should use this RFC to tackle the move of `sitemap` out of Astro core and into the `@astrojs/sitemap` integration. It feel silly to make a different breaking change in sitemap usage here if we know that move is coming anyway.

proposals/0000-config-finalization.md
```diff
@@ -0,0 +1,160 @@
...
+
+    vite: {}
+})
+```
+
+# Detailed design
+
+## Top-Level Options
+
+- Add `log`. Previously only supported through CLI flag.
```
FredKSchott, line 70: nit: Thoughts on `logLevel` to match Vite? `log` does sound a bit like a verb or a true/false boolean.

proposals/0000-config-finalization.md
```diff
@@ -0,0 +1,160 @@
...
+    log: 'info',
+
+    rootDir: '.',
+    srcDir: './src',
+    publicDir: './public',
+    outDir: './dist',
+
+    site: 'https://example.com',
+    trailingSlash: 'always',
+    basePath: '/',
```
FredKSchott, line 29: nit: thoughts on `base` to match Vite? https://vitejs.dev/config/#base I don't think `Path` is adding too much here, and `base` does sound pretty clean

proposals/0000-config-finalization.md
```diff
@@ -0,0 +1,160 @@
...
+When this RFC is accepted, a config file for Astro might look something like this:
+
+```js
+import { defineConfig } from 'astro';
+import preact from '@astrojs/preact';
+
+export default defineConfig({
+    log: 'info',
+
+    rootDir: '.',
```
FredKSchott, line 22: nit: I'm a big fan of keeping abbreviations to a minimum, so I'd prefer `rootDirectory` over `rootDir`. I expressed this earlier and think I'm outvoted here, and this is far from a hill I'll die on, but consider this me registering my opinion :)

Also, given my other comments here I do feel a responsibility to call out that Vite uses `Dir` for `publicDir`, so I do see the value in matching there.

proposals/0000-config-finalization.md
FredKSchott: SGTM, we can always add this back later without it being a breaking change

proposals/0000-config-finalization.md
```diff
@@ -0,0 +1,160 @@
...
+- Rename `styleOptions` => `style`
+- Passed through to `vite.css`
+
+## Integrations
+
+There are no proposed changes to the `integrations` array.
+
+## Experimental
+
+This RFC introduces a new `experimental` option. This is where future experimental options will go.
```
FredKSchott, line 144: love making this explicit :)

proposals/0000-config-finalization.md
```diff
@@ -0,0 +1,160 @@
...
+
+# Drawbacks
+
+Churn is the major concern, but now is the time to make breaking "clean-up" changes like this!
+
+# Adoption strategy
+
+In addition to `defineConfig` providing the new config types, this RFC will be aided by documentation in the form of a Mirgration Guide.
+
+If possible, we should introduce a codemod to automate migration.
```
FredKSchott, line 160: In addition (or instead of) a codemod, it would be nice to catch the old config format and map it to the new config format. Based on what I see here, I don't *think* that will be anything more than mapping properties from one to the other, like this:

```js
function updateLegacyConfig(oldConfig) {
  return { 
    rootDir: oldConfig.projectRoot,
    // ...
  };
}
```

This would make this a non-breaking change, and could also do something similar to integrations where it outputs the new config object to the CLI on startup, for easy copy-paste (probably with a few limitations).

FredKSchott:
Overall big +1 on this! I left some comments but they're all subjective & personal preference nits, so feel free to consider them as feedback and advice but nothing blocking consensus from me.

As much as I can say this for myself and maybe to all reviewers: We have a policy of not blocking RFC consensus on final naming decisions, leaving any bike-shedding to be settled in the final PR review process. Obviously this RFC is unique in that it is MOSTLY naming, but I think that same policy can apply here. As reviewers of this RFC, I see our role as giving Nate feedback on these changes that he can take into the final PR, and not necessarily blocking on personal preference. Just a thought!

proposals/0000-config-finalization.md
natemoo-re: I think this delineation makes it pretty clear that `markdown` and `sitemap` options are essentially integrations that come out-of-the-box. I think it's worth keeping both built-in for v1.0.

proposals/0000-config-finalization.md
natemoo-re: I don't disagree but looking at lots of other projects, I concluded that `Dir` was the unspoken standard here.

natemoo-re: It is worth calling out that Vite has `root` and then everything else (`publicDir`) gets the `Dir` suffix. Seemed confusing to have two different ways to name a directory option.

proposals/0000-config-finalization.md
natemoo-re: I agree with this! Will update.

proposals/0000-config-finalization.md
natemoo-re: This would be the only `camelCase` option, but I'm not opposed

proposals/0000-config-finalization.md
FredKSchott: Right, but I guess my question is "does sitemap need to be an integration that comes out-of-the-box?". Especially in a world where SSG vs. SSR is decided by your adapter and not always SSG.

Astro's `sitemap` feature is not currently powered by the integration, so there will be some extra implementation work on your end to stitch it all together if you go the "built-in" route.

proposals/0000-config-finalization.md
FredKSchott: Yea, If we feel comfortable not matching Vite in some cases, I think that means we probably don't feel that strongly about matching Vite in other cases either. IDK, not a hill I'll die on but this does strike me as a bit odd as I try to weigh these objectively.

proposals/0000-config-finalization.md
natemoo-re: Updated to remove `sitemap` option. It should be configured by the `@astrojs/sitemap` integration.

PR #163, opened 2022-03-29
0019: Unify markdown plugins into a single array

JuanM04:
## Summary

Update to 0019

Our markdown integration [doesn't distinguish](https://github.com/withastro/astro/blob/23b8ee2738076225da57ce5e3826373c8964717c/packages/markdown/remark/src/index.ts#L59-L60) Remark from Rehype plugins, so maybe it's better to have them in a single array

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/0019/markdown-plugins/proposals/0019-config-finalization.md)

### DIFF
--- a/proposals/0019-config-finalization.md
+++ b/proposals/0019-config-finalization.md
@@ -114,6 +114,7 @@ This RFC proposes:
   > Internally this may be implemented in a pluggable manner, but that's an implementation detail.
 - Move the `render[1]` options object to the top-level.
 - Move `buildOptions.draft` => `markdown.drafts`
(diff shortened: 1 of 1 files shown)


FredKSchott:
cc @natemoo-re

This change is too small to go through our full RFC process, especially since we don't have another call until 2 weeks from today.

What I might suggest is instead reaching out to Nate (in any dev-related Discord channel) and running this idea by him. The suggestion is small enough that it could be easily incorporated into our PR process, without having to go through the full RFC process. Nate might even be looking for some help implementing this RFC!

PR #178, opened 2022-04-25
API route proposed changes

matthewp:
- Start Date: 2022-04-25
- Status: Draft

## Summary

Change the signature for API routes into a single `APIContext` argument, allowing you to use destructing to get the properties you want:

```js
export async function get({ params, request }) {
  if(!request.headers.has('cookie')) {
    return new Response(null, {
      status: 301,
      headers: {
        Location: '/'
      }
    });
  }

  // ...
}
```

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/b636579ce4e6781c119433147eb93e0f9db8c049/proposals/api-route-signature.md)

### DIFF
--- a/proposals/api-route-signature.md
+++ b/proposals/api-route-signature.md
@@ -78,6 +78,27 @@ export function get({ request }: APIContext) {
 
 In the future you could see APIContext being expanded to add additional properties, such as a read/write cookie interface, without needing to adjust user code.
 
+## Types
+
+There are 2 types provided in the proposal:
+
+- `APIContext` is the context object. You can use it like so:
+  ```ts
+  import type { APIContext } from 'astro';
+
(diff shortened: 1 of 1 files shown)


matthewp:
Feedback from the call:

- Could be helped by showing what other frameworks do here.
- Some people prefer request as the first arg and an APIContext as the second. Want to see how people respond.

FredKSchott:
+1 LGTM

Thanks for taking the extra time to flesh this out, it's nice to see that others are also using the single-argument approach and I agree that that feels like the right path forward.

Princesseuh:
Looks good! I especially really like the APIRoute way of typing the route

matthewp:
This was accepted!

PR #184, opened 2022-04-29
Deprecate the `<Markdown>` Component

FredKSchott:
- Start Date: 2022-04-29
- Status: Draft

## Summary

Hey everyone! We (the core maintainers, @withastro/maintainers-core) have been discussing the `<Markdown />` component a lot over the last few days, and have made the decision to deprecate the `<Markdown />` component ahead of our v1.0 release. **It will continue to be available as a non-core, user-land npm package (`@astrojs/markdown`) post v1.0.**

This is a bit of an odd RFC in that it's driven by a need to fix a buggy experience, and not talking through the best way to add something new. We've been discussing this on the core team on-off for a few weeks now, and have already reached informal consensus within the group, but I would still like to share this with our larger maintainer community for feedback during the next RFC call.

For more details and why we see it as necessary, read on.

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/deprecate-markdown/proposals/0000-deprecate-markdown-component.md)

## History

- Discussed on the 2022-04-26 RFC Call and was received positively. Will attempt official consensus at the next RFC call.

### DIFF
--- a/proposals/0000-deprecate-markdown-component.md
+++ b/proposals/0000-deprecate-markdown-component.md
@@ -1,0 +1,69 @@
+- Start Date: 04-29-2022
+- Reference Issues: https://github.com/withastro/rfcs/discussions/179
+- Implementation PR: <!-- leave empty -->
+
+# Summary
+
+Hey everyone! We (the core maintainers, @withastro/maintainers-core) have been discussing the `<Markdown />` component a lot over the last few days, and have made the decision to deprecate the `<Markdown />` component ahead of our v1.0 release. **It will continue to be available as a non-core, user-land npm package (`@astrojs/markdown`) post v1.0.**
+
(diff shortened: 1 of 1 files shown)


delucis:
Couldn’t make it to the RFC call, but this all makes sense and the RFC reads well 👍

brycewray:
I’ll humbly offer a “peanut-gallery”-level comment because the Full Rendered Proposal explains all this superbly well: I hope that “userland plugin” will be available **before** `<Markdown />` ceases to work, because I can’t find a working or compatible plugin, whether Remark or Rehype, that substitutes for what how I’ve been using `<Markdown />`. My use case is small, but it’s there. :) Thanks in advance for whatever consideration this receives!

PR #204, opened 2022-06-01
feat: add inject-route rfc

thepassle:
- Start Date: 2022-06-01
- Status: Ready

## Summary

Proposes a `injectRoute` API to the integrations `'astro:config:setup'` hook, similar to `injectScript`. This API would allow integrations to add routes to projects, that ship have prebuild, plug-n-play functionality, pages, or route handlers. For consumers of such integrations, this would allow highly plug-and-play like experiences for adding functionality.

## Links

- Full Rendered Proposal: https://github.com/thepassle/rfcs/blob/feat/inject-route/proposals/0023-inject-route.md
- Docs PR: https://github.com/withastro/docs/pull/704
- Implementation PR: https://github.com/withastro/astro/pull/3457

### DIFF
--- a/proposals/0023-inject-route.md
+++ b/proposals/0023-inject-route.md
@@ -1,0 +1,142 @@
+- Start Date: 2022-06-01
+- Reference Issues: https://github.com/withastro/rfcs/discussions/201
+- Implementation PR: https://github.com/withastro/astro/pull/3457
+
+# Summary
+
+Recently, @FredKSchott made a thread on the discord #feedback-and-suggestions channel that proposes a `injectRoute` API to the integrations `'astro:config:setup'` hook, similar to `injectScript`. This API would allow integrations to add routes to projects, that ship have prebuild, plug-n-play functionality, pages, or route handlers. For consumers of such integrations, this would allow highly plug-and-play like experiences for adding functionality.
+
+# Example
+
+```js
+// astro.config.mjs
+export default defineConfig({
+  integrations: [
+    {
+      name: 'my-netlify-integration',
+      hooks: {
+        'astro:config:setup': ({injectRoute}) => {
+          injectRoute({
+            /** The route on which to output the entryPoint */
+            pattern: '/admin',
+            /** Bare module specifier pointing to a pre-made admin page */
+            entryPoint: 'my-netlify-integration/admin.astro'
+          })
+        }
+      }
+    }
+  ]
+});
+```
+
+# Motivation
+
+Some usecases for this could be:
+- Adding an `/admin` page for headless CMSes
+- Implementation of [tailwind-config-viewer](https://github.com/rogden/tailwind-config-viewer)
+- Authentication providers could very easily ship the required redirectCallback routes etc, e.g. `googleProvider()`, `facebookProvider()`
+- Plug and play payment integrations, endpoints, webhooks, all setup via a single integration, e.g. `paypal()`, or `stripe()`
+
+# Detailed design
+
+There is a prototype implementation here: 
+https://github.com/withastro/astro/pull/3457
+
+## Proposed API
+
+```ts
+export interface InjectedRoute {
+  pattern: string,
+  entryPoint: string
+}
+
+function injectRoute(injectRoute: InjectedRoute): void {};
+```
+
+# Drawbacks
+
+No known drawbacks
+
+# Alternatives
+
+No known alternatives
+
+# Adoption strategy
+
+This is a new API and won't require any migrations.
+
+# Unresolved questions
+
+**Resolved:** ✅
+
+**Q:** _Should `_`'s in route names be allowed?_
+
+**A:** Yes. An expected usecase for `injectRoute` is to add "private" routes, like for example `/_admin`, and allowing `_`'s will also help avoid nameclashes. The draft implementation currently already supports this.
+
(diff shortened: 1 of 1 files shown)


matthewp:
Well put together proposal! 

Is there anything to be aware of with regard to SSR? Does building to SSR include these routes? I imagine the logic is similar to how renderers are included?

thepassle:
I did some quick local testing for both ssg and ssr, and the pages seem to get output in the build outputs correctly. However as mentioned in the community call yesterday, ill do some more thorough testing to confirm that everything works as expected for both ssg and ssr.

delucis:
Amazing work @thepassle! This is going to be a really exciting addition. Hope you don’t mind me asking a few questions as I dug quite deep into the potential use cases for [Netlify CMS integration](https://github.com/delucis/astro-netlify-cms/). Sorry I couldn’t make the call on Tuesday, so if I’m rehashing anything discussed, please ignore me/let me know.

### How would an integration support customisation of an injected page?

#### Use cases
- For Netlify CMS, users can add custom CSS and React components to extend the editor and preview pane.
- For e-commerce, checkout routes may need a logo component or custom CSS.
- For auth, API keys may need passing.
- Other configuration of page content (feature selection, passing variables etc.)

My hacky solution for the CMS integration was to dynamically construct the page as a string and use that. To do the same with this API I guess we’d need to use Vite to inject our page string as a virtual module? That could be clunky.

#### Possible solution

Astro pages are components. Components receive data via props. Could the interface be something like this?

```js
export interface InjectedRoute {
  pattern: string;
  entryPoint: string;
  props: Record<string, any>;
}
```


##### Drawbacks

- What happens with spread path patterns? Those would potentially generate `props` from `getStaticPaths` I guess. Is a static `props` passed from `injectRoute` which `getStaticPaths` could override too confusing?

- How would we build dynamic configuration-based imports in a `.astro` file? (see first alternative below)


#### Alternatives

- Allow passing a string instead of a path for the page component. This may be essential so you could build an import like `import CustomLogo from '${logoPathOption}'`. I don’t think that would be possible with the props model above.

   So literally:

   ```js
   const component = `---
   ---
   <h1>${title}</h1>`;
   
   injectRoute({
     pattern: '/admin',
     component,
   });
   ```

- Integrations could require specific fixed locations or a config file the route imports. That means breaking config in two: `astro.config.mjs` and custom integration-specific file locations.


### How are imports resolved inside an injected route?

This really is a question rather than a proposal 😅 If `injectedRoute.astro` includes `import '../thing'` I assume that’s resolved relative to the integration package. How could an injected route import a project-specific file? Can we provide a nice API for that?

thepassle:
Good questions @delucis ! Thanks for the feedback. I definitely think allowing for some kind of customization is required for this feature to be useful indeed, the usecases you mentioned in your post are really good examples.

Im unsure on the implementation of something like you suggest, however. Currently, the implementation really only adds routes to the route manifest which adds them to the build output, and routes and _rendering_ components (or passing props to the components at the route location) are two intrinsically separate things. The way the implementation is currently set up, I don't see a straightforward way to implement this. I'd be interested to hear if the team has any thoughts on how to achieve something like this. cc @natemoo-re @FredKSchott @matthewp 

Wrt alternatives: It's not entirely clear to me what you mean with the `logoPathOption` example, could you elaborate? I'm also not sure if passing components-as-strings in the configuration will work out very well

Wrt import resolving: Imports are just handled via rollup/plugin-node-resolve via vite/rollup internally.

delucis:
Thanks @thepassle!

> Wrt alternatives: It's not entirely clear to me what you mean with the `logoPathOption` example, could you elaborate? I'm also not sure if passing components-as-strings in the configuration will work out very well

My thinking was this:

- An integration wants to add a dashboard so will use `injectRoute`
- It wants to offer some customisation by letting users add some custom Astro components in the integration config
- How do we get the component from the config file to the injected route?
- At the moment the way to use one Astro component inside another is to import it in the frontmatter, so we'd need to build our component dynamically:

   ```js
   const makeRoute = (logoPath) => `---
   import Logo from ${logoPath}
   ---
   <Logo />
   `

   // Generate the string for injectRoute
   // In reality the path would be the option passed from the integration config
   makeRoute('src/components/MyLogo.astro')
   ```

   This dynamic project-provided content also needs resolving correctly in relation to the route's location, so that was the motivation for my question about how things get resolved. A `resolveFromAstroRoot` helper for example would support users providing paths from their root directory.

natemoo-re:
These are great questions @delucis! @thepassle is definitely correct that the current implementation makes it difficult to attach custom props to the page. It's not impossible, but it's definitely a bigger lift than the current proposal.

As for resolving user-specific components from the project root, integrations are welcome to use `/src` relative imports. Combine that with `import.meta.glob` and well documented file structure conventions for each integration and I think that covers most of the use cases you've outlined. I actually built a proof of concept using these primitives and it worked well!

Personally, I'd love to see an initial implementation of `injectRoute` in users hands so we can iteratively improve on the API if we hear that the constraints are too limiting.

delucis:
Sounds good @natemoo-re! Totally understand getting a MVP across the line and just want to make sure we have considered a theoretical path for these enhancements in the future even if they aren’t implemented just yet. Looking forward to today’s call!

PR #210, opened 2022-06-17
RFC for Astro.extend User/Plugin Space

tanepiper:
- Start Date: 2022-06-17
- Status: Draft

## Summary

RFC for `Astro.extend` Userspace, where template plugins can be attached in a style similar to jQuery plugins, here they provide lightweight extensions and templates 

## Links
- [Full Rendered Proposal](https://github.com/tanepiper/astro-rfcs/blob/main/proposals/0024-Astro-user-space.md)

### DIFF
--- a/proposals/0024-Astro-user-space.md
+++ b/proposals/0024-Astro-user-space.md
@@ -86,11 +86,10 @@ export function definePlugin(plugin, config) {
 
 # Motivation
 
+These differ in use from integrations in that they provide globally available template and data functions
+that can be used in any template. They should be lightweight and provide helpful features.
 
-
-Having functions available in the Global namespace that give additional functionality to end users in an
-easy way to create new server-side features that are aimed at being lightweight. This design is not final
-and would serve better from discussion around the needs, vs Integrations
+This design is not finaland would serve better from discussion around the needs, vs Integrations
 
 # Detailed design
 
--- a/proposals/0024-Astro-user-space.md
+++ b/proposals/0024-Astro-user-space.md
@@ -1,0 +1,124 @@
+- Start Date: 2022-06-16
+- Reference Issues: <!-- related issues, otherwise leave empty -->
+- Implementation PR: <!-- leave empty -->
+
+# Summary
+
+Provide a user namespace in the `Astro` global that allows for developers to implememnt plugins that can be used within Astro templates
+without having to import libraries into the templates, instead provide them as part of a user namespace.
+
+The current propsal is `Astro.extend.` and takes inspiration from jQuery plugins, where a function can easily be added to
+the namespace.
+
+# Example
+
+This is a useless example that provides an idea of how you attach to the namespace and could provide a function
+to return some content HTML.
+
+The stucture of a plugin is a default function that is called that passes the `Astro` context and a configuration, as seen below
+
+plugin1.js
+```mjs
+export default function(Astro, config) {
+    return (userConfig) => {
+        const { default1, default2, Content } = {...config, ...userConfig}; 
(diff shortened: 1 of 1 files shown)


crutchcorn:
I worry that this potentially impacts:

- Stability, [as highlighted in the RFC](https://github.com/tanepiper/astro-rfcs/blob/main/proposals/0024-Astro-user-space.md#drawbacks)
- Code migratability. If the API for this changes, it potentially introduces another source of headache both for devs who use this feature and the Astro team
- Debuggability. Globals are always tricky to stack trace, doubly so with build tooling, where it's harder to filter out stacks IMO

I also am not sure I think the idea of using a global for UI elements [in the two defined examples](https://github.com/tanepiper/astro-rfcs/blob/main/proposals/0024-Astro-user-space.md#example) is the cleanest code pattern since it introduces some headache in regards to figuring out the flow of code, debuggability, etc.

I wonder if we're better off being prescriptive about enforcing explicit code imports as opposed to utilizing `extend`

tanepiper:
@crutchcorn 

* I'd argue stability is always a factor in any user plugin, and of course in any 3rd part integration too
* APIs always change, but if your version correctly, announce upcoming deprecations and breaking changes, it should help and popular plugins will tend to be maintained. Of course, the Astro team have to take on board any risk a feature like this introduces, but ideally through design would find a way to reduce ways to harm users (such as sandboxing certain things)
* Yes debugging this kind of stuff is never fun, but see above - maybe some domain ability could provide a way to capture error context correctly

My initial thinking around this was actually more like Angular pipes, but possibly using the new [Pipe operator](https://tc39.es/proposal-pipeline-operator/) syntax - but this would require those functions to be made global - with a namespace, at least there is less chance for abuse.

This could be done via a config option where you have to define them explicitly. The idea is to remove the need to import functions over a lot of different components.

Sure in the menu example, maybe this should be a component - but maybe you want to have functions that render certain types of content in a more declarative way, session-based helpers, etc 

FWIW we've built an OpenAPI renderer that just imports a library, parses the file and we pass the object through a bunch of components - it works fine, but with our setup, we can't pass the file object back up to get it at a higher level, so now we have to export the internal parse function and call it every file we use it in - to simplify it we now import the `getStaticPaths` method from our [spec].astro file which has the loader method, and uses the data now on other pages - but it's not really clean.

I'm sure there are better ways of architecting what we're building, but it feels like there's a space missing for easy-to-create small functions that don't need to be a whole component

okikio:
I'm against this, in it current form. Personally I'd prefer a singular place where I can go to determine, where plugins, extensions are it makes it much easier to debug, terse and understand code. `Astro.extend` would be hard to follow for those new to a code base, or while trying to debug where errors are coming from.

proposals/0024-Astro-user-space.md
```diff
@@ -0,0 +1,123 @@
...
+    return async (userConfig, onlyReturnData) => {
+        Astro.extend.menuCreate = (url) => {
+            const result = await fetch(userConfig?.url || config.url);
+            const rows = await result.json();
+            if (onlyReturnData) {
+                return rows;
+            }
+
+            return (
+                <ul class="menu">
```
matthewp, line 50: Is this JSX?

tanepiper:
Talking to one of the other devs, I think the best example we could find that would be most explainable is to offer a space to have functions like Hugo - https://gohugo.io/functions

proposals/0024-Astro-user-space.md
tanepiper: It's psudocode :)

PR #230, opened 2022-07-18
Add RFC for `Astro.context`

laptou:
- Start Date: 2022-07-17
- Status: Draft

## Summary

Adds a "context" available in Astro files during SSR that can be provided from the SSR entrypoint. In template files, this is `Astro.context`, and in API routes, this is the `context` property of the route's parameter. The purpose of this change is to make
it possible for state that is already derived by the server (such as
authentication state) to easily be passed from Express/Koa to an Astro template
via `Astro.context`, so that this information does not have to be re-derived
from `Astro.request.headers`. 

## A minimal example

In the client, in an Astro template file:

```astro
<html lang="en">
	<head>
		<meta charset="utf-8" />
		<meta name="viewport" content="width=device-width" />
		<title>Astro</title>
	</head>
	<body>
		<p>Welcome, {Astro.context.user.name}</p>
	</body>
</html>
```

In the client, in an API endpoint:

```typescript
export function get({ params, context }) {
  return context.user.name;
}
```

In the server, using NodeJS SSR adapter:

```typescript
import { createServer } from 'http';
const { handler } = await import('./path-to-entrypoint.mjs');
server = createServer((req, res) => {
  handler(req, res, { user: { name: 'Elvis' } });
});
server.listen();
```

## Links

- [Full Rendered Proposal](https://github.com/laptou/astro-rfcs/blob/laptou/ssr-context/proposals/0024-astro-context.md)

### DIFF
--- a/proposals/0024-astro-context.md
+++ b/proposals/0024-astro-context.md
@@ -1,0 +1,139 @@
+- Start Date: 2022-07-14
+- Reference Issues: 
+- Implementation PR: [astro#3934](https://github.com/withastro/astro/pull/3934)
+
+# Summary
+
+Adds a "context" available in Astro files during SSR that can be provided from the SSR entrypoint. In template files, this is `Astro.context`, and in API routes, this is the `context` property of the route's parameter.
+
+# Example
+
+## A minimal example
+
+In the client, in an Astro template file:
+
+```astro
+<html lang="en">
+	<head>
+		<meta charset="utf-8" />
+		<meta name="viewport" content="width=device-width" />
+		<title>Astro</title>
+	</head>
+	<body>
+		<p>Welcome, {Astro.context.user.name}</p>
+	</body>
+</html>
+```
+
+In the client, in an API endpoint:
+
+```typescript
+export function get({ params, context }) {
+  return context.user.name;
+}
+```
+
+In the server, using NodeJS SSR adapter:
+
+```typescript
+import { createServer } from 'http';
+
+const { handler } = await import('./path-to-entrypoint.mjs');
+
+server = createServer((req, res) => {
+  handler(req, res, { user: { name: 'Elvis' } });
+});
+
+server.listen();
+```
+
+## A bit more realistic example
+
+In the server, using NodeJS SSR adapter:
+
+```typescript
+import { createServer } from 'http';
+import express from 'express';
+import expressSession from 'express-session';
+import passport from 'passport';
+import { MyStrategy } from 'my-passport-strategy-library';
+
+const app = express();
+
+app.use(expressSession({ secret: 'parsley' }));
+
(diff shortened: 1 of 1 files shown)


proposals/0024-astro-context.md
```diff
@@ -0,0 +1,139 @@
...
+than the NodeJS integration.
+
+This requires modifying the NodeJS adapter to have a compatible function
+signature, plumbing the `context` value through the SSR stack, and exposing it
+on the `Astro` faux-global. At the time of writing, a possible implementation is
+available in [astro#3934](https://github.com/withastro/astro/pull/3934).
+
+# Drawbacks
+
+This change, as described, would only be usable via the NodeJS adapter. Further
```
matthewp, line 109: Yeah, I think this is a bit of a problem. I think we would be hesitant to add a property to the Astro global that only worked in one adapter. If we had an idea to make it work for others that would make it more palpable.

In the past we have discussed ideas of middleware within Astro itself. Then presumably an Astro middleware could mutate a context object like you've described in this RFC.

RichiCoder1:
As mentioned here: https://github.com/withastro/rfcs/discussions/221#discussioncomment-3171490

A generic mechanism where user code can expose a method that takes in adapter-specific details and outputs a "generic" context would be ideal. Something like:

```typescript
export interface RequestContext {
  user: IUser;
}

export function beforeRequest(event: RequestEvent): Promise<RequestContext> {
  // do auth stuff
  const user = await /* ... */;
  return {
	user {
      name: user.name,
      id: user.id
	}
  };
}
```

Though this specific and common case could also be handled by something like svelete's specifically designed session helper: https://kit.svelte.dev/docs/hooks#getsession

So the above would turn into

```typescript
export interface MySession {
  user: IUser;
}

export function getSession(event: RequestEvent): Promise<MySession> {
  // do auth stuff
  const user = await /* ... */;
  return {
	user {
      name: user.name,
      id: user.id
	}
  };
}
```

matthewp:
@RichiCoder1 How would this work in dev?

I think the underlying problem with this and related ideas is figuring out what role adapters should play in dev. Once we figure that out then features like this will be a bit easier to build.

matthewp:
This idea came up again in relation to Cloudflare's `env` feature.

gijsroge:
I kinda asked the same I think. https://discord.com/channels/830184174198718474/1021517008061345892

wassfila:
> @RichiCoder1 How would this work in dev?
> 
> I think the underlying problem with this and related ideas is figuring out what role adapters should play in dev. Once we figure that out then features like this will be a bit easier to build.

as of Today, middleware is not usable in dev mode, so why make this RFC depend on that ? I also don't see why this would not work with deno.

PR #232, opened 2022-07-18
Create "Astro.url" & Deprecate "Astro.canonicalURL"

FredKSchott:
- Start Date: 2022-07-18
- Status: Draft
- Implementation PR: https://github.com/withastro/astro/pull/3959

## Summary

Create a general-purpose `Astro.url` helper. Deprecate `Astro.canonicalURL` to use `Astro.url` instead.

```js
// Before:
const currentPathname = new URL(Astro.request.url).pathname;
const canonicalURL = Astro.canonicalURL;

// After:
const currentPathname = Astro.url.pathname;
const canonicalURL = new URL(Astro.url.pathname, Astro.site);
```

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/FredKSchott-patch-2/proposals/0000-astro-url-helper.md)


proposals/0000-astro-url-helper.md
```diff
@@ -0,0 +1,118 @@
...
+
+- How does Astro know if multiple pages should have a single canonical URL?
+- In SSR mode, is every request caught by `[...foo].astro` canonically unique?
+
+In addition, there was confusion over the canonical domain used:
+
+- If `Astro.site` is set, what domain does `Astro.canonicalURL` use?
+- If `Astro.site` is not set, what domain does `Astro.canonicalURL` use?
+
+The result is that we have introduced the possibility that in some cases Astro is "lying" to the user about a URL being canonical when it may not be. This is most common when in SSR (where we don't know the built URLs ahead of time) and when Astro.site is not set (where we use the current origin instead of known production origin).
```
matthewp, line 52: I think the SSR point needs a bit more clarification for what you mean. I *think* what you are saying is that you, as a developer can "vouch" for the correctness of a canonical URL in SSG mode because you can see the result before it's merged into production. But I had to think about the sentence a lot to guess at the meaning (if that's even correct)

proposals/0000-astro-url-helper.md
```diff
@@ -0,0 +1,118 @@
+- Start Date: 2022-07-18
+- Reference Issues: N/A
+- Implementation PR: https://github.com/withastro/astro/pull/3959
+
+# Summary
+
+Create a general-purpose `Astro.url` helper. Deprecate `Astro.canonicalURL` to use `Astro.url` instead.
```
matthewp, line 7: This is an overall comment on the RFC, I think the framing as this is a replacement/renaming makes it seem like they are close to the same feature, but they aren't really. 

I think it's more accurate to say that you're adding a new feature, `Astro.url`, is has more general usage in more scenarios and removing Astro.canonicalURL, that has 1 specific use-case.

It's now slightly harder to produce a canonicalURL but it's easier to get the current pathname, etc, things you need more often. I think the examples illustrating this are good.

FredKSchott: Yup, let me see if I can hit this point better. I like the way you've phrased it in this comment.

FredKSchott: added this to the Drawback section and reframed the Motivation section to better frame the RFC as you did.

proposals/0000-astro-url-helper.md
FredKSchott: Good catch, I rewrote this section.

PR #256, opened 2022-08-19
RFC: Add option to @astrojs/rss for including full post content in RSS feed.

slyeargin:
- Start Date: 2022-08-18
- Status: Draft

## Summary

Updating @astrojs/rss to generate feeds that include full `compiledContext()` strings as `content:encoded` in the XML, given a parameter of `contentLength` with a value of `full`.  (This value would default to `summary`, which is the current default behavior.)

## Links

- [Full Rendered Proposal](https://github.com/slyeargin/rfcs/blob/rss-content/proposals/0025-full-length-post-content-in-rss-feeds.md)

### DIFF
--- a/proposals/0025-full-length-post-content-in-rss-feeds.md
+++ b/proposals/0025-full-length-post-content-in-rss-feeds.md
@@ -4,17 +4,17 @@
 
 # Summary
 
-Updating @astrojs/rss to generate feeds that include full `compiledContext()` strings as `content:encoded` in the XML, given a parameter of `contentLength` with a value of `full`.  (This value would default to `summary`, which is the current default behavior.)
+Updating @astrojs/rss to generate feeds that include full `compiledContent()` strings as `content:encoded` in the XML, given a parameter of `contentLength` with a value of `full`.  (This value would default to `summary`, which is the current default behavior.)
 
 # Motivation
 
-I remember when WordPress rolled out the ability to only publish short excerpts in RSS.  Ostensibly, this was to drive traffic to a blog, where analytics about readership could be gathered.  I am not particularly interested in that, so I implemented this change in my own project.  
+ Many authors prefer including full articles in an RSS feed.
 
-# Detailed design
+# Usage
 
 Usage in a project's 	`src/pages/rss.xml.js`:
 
-```
+```js
 const contentLength = 'full';
 
 export const get = () => rss({
@@ -33,16 +33,18 @@ export const get = () => rss({
   });
 ```
 
+# Detailed design
+
 Changes in `astro/packages/astro-rss/src/index.ts`:
 
-```
+```ts
 type RSSOptions = {
 	...
     	contentLength?: "summary" | "full";
 };
 ```
 
-```
+```ts
 type RSSFeedItem = {
 	...
     	/** Item content */
@@ -50,7 +52,7 @@ type RSSFeedItem = {
 };
 ```
 
-```
+```ts
 export async function generateRSS({ rssOptions, items }: GenerateRSSArgs): Promise<string> {
 	const { site, contentLength = 'summary' } = rssOptions;
 	...
@@ -66,7 +68,7 @@ I cannot think of any immediate drawbacks, but I'm open to hearing about them fr
 
 # Adoption strategy
 
-This should be a non-breaking change, as the default behavior will continue unless the developer specifically provides the key/value pair `contentLength: 'full'` *and* a value for `content`.
+This should not be a breaking change, as the default behavior will continue unless the developer specifically provides the key/value pair `contentLength: 'full'` *and* a value for `content`.
 
 # Unresolved questions
 
--- a/proposals/0025-full-length-post-content-in-rss-feeds.md
+++ b/proposals/0025-full-length-post-content-in-rss-feeds.md
@@ -1,0 +1,73 @@
+- Start Date: 2022-08-18
+- Reference Issues: <!-- related issues, otherwise leave empty -->
+- Implementation PR: <!-- leave empty -->
+
+# Summary
+
+Updating @astrojs/rss to generate feeds that include full `compiledContext()` strings as `content:encoded` in the XML, given a parameter of `contentLength` with a value of `full`.  (This value would default to `summary`, which is the current default behavior.)
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


proposals/0025-full-length-post-content-in-rss-feeds.md
```diff
@@ -0,0 +1,73 @@
+- Start Date: 2022-08-18
+- Reference Issues: <!-- related issues, otherwise leave empty -->
+- Implementation PR: <!-- leave empty -->
+
+# Summary
+
+Updating @astrojs/rss to generate feeds that include full `compiledContext()` strings as `content:encoded` in the XML, given a parameter of `contentLength` with a value of `full`.  (This value would default to `summary`, which is the current default behavior.)
```
matthewp, line 7: Should this be `compiledContent` instead?

proposals/0025-full-length-post-content-in-rss-feeds.md
```diff
@@ -0,0 +1,73 @@
...
+
+# Motivation
+
+I remember when WordPress rolled out the ability to only publish short excerpts in RSS.  Ostensibly, this was to drive traffic to a blog, where analytics about readership could be gathered.  I am not particularly interested in that, so I implemented this change in my own project.  
+
+# Detailed design
+
+Usage in a project's 	`src/pages/rss.xml.js`:
+
+```
```
matthewp, line 17: can you add language hints here, adding `js` to the of the backticks so we get syntax hightlighting?

proposals/0025-full-length-post-content-in-rss-feeds.md
```diff
@@ -0,0 +1,73 @@
...
+    	content?: string;
+};
+```
+
+```
+export async function generateRSS({ rssOptions, items }: GenerateRSSArgs): Promise<string> {
+	const { site, contentLength = 'summary' } = rssOptions;
+	...
+	// content:encoded should not be used if description is not present (and first) per RSS 2.0 spec
+       if (result.description && rssOptions.contentLength === 'full' && result.content) {
```
matthewp, line 58: Where is `result` coming from? I don't see it in this example.  Is this code all meant to be pseudo-code?

proposals/0025-full-length-post-content-in-rss-feeds.md
```diff
@@ -0,0 +1,73 @@
...
+        }	
+```
+
+# Drawbacks
+
+I cannot think of any immediate drawbacks, but I'm open to hearing about them from people more familiar with the inner workings of Astro.
+
+# Adoption strategy
+
+This should be a non-breaking change, as the default behavior will continue unless the developer specifically provides the key/value pair `contentLength: 'full'` *and* a value for `content`.
```
matthewp, line 69: Do you mean that this *shouldn't* be a breaking change?

proposals/0025-full-length-post-content-in-rss-feeds.md
```diff
@@ -0,0 +1,73 @@
...
+
+Updating @astrojs/rss to generate feeds that include full `compiledContext()` strings as `content:encoded` in the XML, given a parameter of `contentLength` with a value of `full`.  (This value would default to `summary`, which is the current default behavior.)
+
+# Motivation
+
+I remember when WordPress rolled out the ability to only publish short excerpts in RSS.  Ostensibly, this was to drive traffic to a blog, where analytics about readership could be gathered.  I am not particularly interested in that, so I implemented this change in my own project.  
+
+# Detailed design
+
+Usage in a project's 	`src/pages/rss.xml.js`:
```
matthewp, line 15: I think this example usage should go in a `# Usage` section before the detailed design. The detailed design is intended to explain how the change can be implemented.

proposals/0025-full-length-post-content-in-rss-feeds.md
```diff
@@ -0,0 +1,73 @@
...
+- Reference Issues: <!-- related issues, otherwise leave empty -->
+- Implementation PR: <!-- leave empty -->
+
+# Summary
+
+Updating @astrojs/rss to generate feeds that include full `compiledContext()` strings as `content:encoded` in the XML, given a parameter of `contentLength` with a value of `full`.  (This value would default to `summary`, which is the current default behavior.)
+
+# Motivation
+
+I remember when WordPress rolled out the ability to only publish short excerpts in RSS.  Ostensibly, this was to drive traffic to a blog, where analytics about readership could be gathered.  I am not particularly interested in that, so I implemented this change in my own project.
```
matthewp, line 11: The motivation section is intended to explain the shortcomings of the current design and why we should change it / add a given feature. I would change this paragraph to remove your personal anecdote and instead explain how some authors prefer full articles in the RSS feed.

matthewp:
I think the idea is good and well scoped. There are some issues with the RFC itself that I'd like to see cleaned up, but assuming that can be done I would be in favor of this.

proposals/0025-full-length-post-content-in-rss-feeds.md
slyeargin: Good catch.

proposals/0025-full-length-post-content-in-rss-feeds.md
slyeargin: On it.

proposals/0025-full-length-post-content-in-rss-feeds.md
slyeargin: It's in [the existing code](https://github.com/withastro/astro/blob/main/packages/astro-rss/src/index.ts#L123).  These are my diffs from the file, intended to illustrate the changes.

SeanMcP:
Forgive me if this is a naive question, but is `<content:encoded>` a valid element in RSS 2.0? I don't see it in [the list of item subelements](https://validator.w3.org/feed/docs/rss2.html#hrelementsOfLtitemgt), but it [is mentioned here for 1.0](https://www.w3.org/wiki/RssContent#content:encoded_without_CDATA).

naiyerasif:
I would not want to published the exact compiled content that's rendered by Astro for the following reasons:

- The syntax highlighting itself can significantly increase the size of the file (a rough estimate is by 3x times).
- I may want to customise the content for syndication using additional Remark/Rehype plugins

Ideally, I'd want a separate markdown pipeline to generate the HTML optimised for syndication.

SeanMcP:
This feature was added to `@astrojs/rss` in `1.6.14` with a different API than this RFC suggests. As a result, this can probably be closed.

@naiyerasif: Full content is opt-in, so you don't need to include it if you don't want.

PR #285, opened 2022-09-05
RFC: MDX components in layout component

kylebutts:


wassfila:
in the example, there is a glob for all of the content, '../content/*.mdx'. It is rare that markdown has one .astro file per .md file, so simply shifting the components from where its used to the layout, might not improve visibility.
Also, the layout, is no special component, but any component can have slots. The interesting concept here is how to create a context for child slot component without having to pass them in each step, and that could be solved in a generic way not just for a specific component variable.

kizu:
Want to +1 here, and also mention another flawed alternative: it is possible to bulk export the whole `components` object with all the component associations, making it a single line you'd need to use in your .mdx:

```js
export { components } from '../components/index.js';
```

This works for regular components, _but_, playing with the code a bit (just starting with astro, so might be mistaken) there seems to be no way to create an index file like this and export multiple astro components into it? So this method would work only for non-astro components, and the main issue with this is that it won't then be possible to use client directives for those components.

Having an easy way to export a list of mdx components from a layout component would be a very welcomed feature!

kylebutts:
Thanks @natemoo-re; I think Content Collections works wonderfully!

PR #293, opened 2022-09-10
RFC: change injectScript's "page-ssr" stage behavior

yassinedoghri:
- Start Date: 2022-09-10
- Status: Draft

## Summary

[integration API] The injected script during the `page-ssr` stage should be run
on every page and behave exactly as the frontmatter code in astro pages.

This would allow for more powerful integrations and use cases + make the integration API easier to work with. 

## Links

- [Full Rendered Proposal](https://github.com/yassinedoghri/astro-rfcs/blob/main/proposals/0025-change-inject-script-page-ssr-behavior.md)
- [Github Discussion](https://github.com/withastro/rfcs/discussions/297)
- [(old) Discord discussion](https://discord.com/channels/830184174198718474/987756543946682398)

### DIFF
--- a/proposals/0025-change-inject-script-page-ssr-behavior.md
+++ b/proposals/0025-change-inject-script-page-ssr-behavior.md
@@ -1,0 +1,115 @@
+- Start Date: 2022-09-10
+- Reference Issues: https://github.com/withastro/astro/issues/3601,
+  https://github.com/withastro/docs/issues/806
+- Implementation PR: <!-- leave empty -->
+
+# Summary
+
+[integration API] The injected script during the `page-ssr` stage should be run
+on every page and behave exactly as the frontmatter code in astro pages.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


natemoo-re:
Hey @yassinedoghri, thanks for the proposal! Apologies that we haven't given this any direct feedback yet.

After chatting with the team, we understand that the current `page-ssr` hooks are limited for your particular use case. However, we have fundamental reservations about exposing the ability for integrations to inject invisible code in the same scope as user code.

For those reasons, we're going to close this particular proposal. That being said, we'd love to work with you to expose the correct hooks to unblock your use case! If you'd like to open a new [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/new/choose) outlining your goals rather than a specific API, we'd love to continue this conversation there.

yassinedoghri:
I understand, thanks for taking the time to check the RFC!

PR #300, opened 2022-09-12
RFC: Component interface

Princesseuh:
- Start Date: 2022-09-12
- Status: Draft

## Summary

Introduces a new interface called "Component" that is used to describe components. This unlocks the following benefits:

- Ability to document components
- (Future) Ability to type slots and possibly other things an Astro component could need in the future in one consolidated interface

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/44eec52b148e57b0639fc7746a47f9eaa2819b73/proposals/component-interface.md)

### DIFF
--- a/proposals/0026-component-interface.md
+++ b/proposals/0026-component-interface.md
@@ -15,7 +15,7 @@ This proposal introduces a new interface called "Component" that is used to desc
 ---
 /**
  * @name Button
- * 
+ *
  * My super button
 */
 interface Component {
@@ -60,7 +60,7 @@ Together, this would create the following signature:
 ```tsx
 /**
  * @name Button
- * 
+ *
  * My super button
  */
 export default function Button__AstroComponent_(_props: Component.props) {}
@@ -109,6 +109,8 @@ But, the following drawbacks:
 Since we can fairly easily achieve this in a way that is backward compatible, we can add support for it in our tooling first, document it (replacing our current `interface Props` documentation) and users will passively migrate to it.
 
 Additionally, the editor tooling could warn when `interface Props` is used and provide a code action to migrate to the new interface to make migration as easy as possible.
+
+Ultimately, we can remove support for `interface Props` in a future major version of Astro as we see fit.
 
 # Unresolved questions
 
--- a/proposals/component-interface.md
+++ b/proposals/component-interface.md
@@ -1,0 +1,115 @@
+- Start Date: 2022-09-12
+- Reference Issues: <!-- related issues, otherwise leave empty -->
+- Implementation PR: <!-- leave empty -->
+
+# Summary
+
+This proposal introduces a new interface called "Component" that is used to describe components. This unlocks the following benefits:
+
+- Ability to document components
+- (Future) Ability to type slots and possibly other things an Astro component could need in the future in one consolidated interface
+
+# Example
+
+```astro
+---
+/**
+ * @name Button
+ * 
+ * My super button
+*/
+interface Component {
+ props: {
+  // Equivalent to the current Props interface
+  name: string;
+ }
+ slots: {
+  // In the future, could be used to type slots:
+  // default: string;
+ }
+}
+
+const { name } = Astro.props;
+---
+
+<button title={name}>Click me!</button>
+```
+
+Result in the editor:
+
+<img src="https://user-images.githubusercontent.com/3019731/189713136-12b69257-6876-4c9a-a248-5891fb1755d5.png">
+
+# Motivation
+
+Right now, it's impossible for Astro users to document their component like they would in a JSX or Svelte component. Having a "Component" interface describing the component allows for a intuitive spot for a JSDoc comment.
+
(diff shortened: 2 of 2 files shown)


Jutanium:
Love this idea. The slots typing gets back to some awkwardness I've encountered: how do you accurately type slots and Elements in Astro?

Princesseuh:
> Love this idea. The slots typing gets back to some awkwardness I've encountered: how do you accurately type slots and Elements in Astro?

Admittedly, the slots typing part isn't really part of this RFC as much as this RFC unlocks at least room for a place to type slots 😛

Enteleform:
+1, looks good!  
&nbsp;  
> (Future) Ability to type slots and possibly ***other things***

I think due to the possibility for growth, the `Component` implementation might be preferable to the fragmented implementation in `Alternatives`.

For example, in a scenario where you're extending a common interface in multiple `.astro` files, you would only have to make updates to `SomeCommonInterface.ts` rather than going into each `.astro` file to check whether it has appropriate definitions for all of the specific fragments that now exist/don't-exist.

&nbsp;  
> 

In the final implementation, would this be providing more thorough types for `props`?

e.g. in the case of `Example`:
```ts
function Button(props:{name:string}): any
```

Princesseuh:
> In the final implementation, would this be providing more thorough types for `props`?
> 
> e.g. in the case of `Example`:
> ```ts
> function Button(props:{name:string}): any
> ```

Yes, just like the `Props` interface does at the moment.

natemoo-re (commented):
Love how this is coming together! Would we also be able to ship strict typing for `props` if a user opts-into the new `Component` interface?

Princesseuh:
> Love how this is coming together! Would we also be able to ship strict typing for `props` if a user opts-into the new `Component` interface?

I am not 100% sure if I get what you mean. If you mean typing `Astro.props` strictly using `Component.props`, yes, that's possible.

If you mean making sure the shape of `Component` (or `Component.props`) is conform, I do not think so as it is still the definition of a new interface at the end of the day. This is also a strength in some way since it allows users to add new properties to their Component definition.

delucis:
Love this @Princesseuh! Definitely has a nice migration story too: keep supporting `Props` and when a user is ready they move their `Props` into `Component`. Can’t wait to use this!

okikio:
@Princesseuh Would it be possible to add export to the Component interface? It'd make it clear that the interface is meant for outside leaning tools. 

e.g.
```ts
/**
 * @name Button
 * Description
 *
 * @slot default - Default slot
 * @slot button - Change the button
 * @slot ...
 */
export interface Component {
  // ...
}
```

okikio:
@Princesseuh What do you think about using `@slot` to document slot?

Princesseuh:
> @Princesseuh What do you think about using `@slot` to document slot?

I'd rather not introduce additional rules to normal JSDoc comments to avoid incompatibility with external tools such as documentation generators

I think the format I'd recommend is a JSDoc comment on top of the actual definition of each slots - like you'd do for props right now.

(Should be noted that typing slots and the syntax for them isn't really in the scope of this RFC, it was just meant as an example of how this can be extended)

okikio:
So, if I'm understanding you correctly, when a user has both `Component`, `Props` and `Slots` the language tool will to prioritize `Component`, as users may want to define `Props` and/or `Slots` seperately/to add TSDoc comments to it, and then use it with `Component`

e.g.
```ts
/** 
 * ...
 */
interface Props {
 ...
}

/** 
 * ...
 */
interface Slots {
 ...
}

/**
 * @name Component
 * Description
 */
interface Components {
 slots: Slots,
 props: Props
}
```

Princesseuh:
> So, if I'm understanding you correctly, when a user has both `Component`, `Props` and `Slots` the language tool will prioritize `Component`, as users may want to define `Props` and/or `Slots` seperately/to add TSDoc comments to it, and then use it with `Component`
> 
> e.g.
> ```ts
> /** 
>  * ...
>  */
> interface Props {
>   ...
> }
> 
> /** 
>  * ...
>  */
> interface Slots {
>   ...
> }
> 
> /**
>  * @name Component
>  * Description
>  */
> interface Components {
>   slots: Slots,
>   props: Props
> }
> ```

That is right. For retrocompatibility sake, if the Component interface didn't exist in your example, the Props one would be used

okikio:
Awesome

jasikpark:
> > Love how this is coming together! Would we also be able to ship strict typing for `props` if a user opts-into the new `Component` interface?
> 
> 
> 
> I am not 100% sure if I get what you mean. If you mean typing `Astro.props` strictly using `Component.props`, yes, that's possible, no issue there
> 
> 
> 
> If you mean making sure the shape of `Component` (or `Component.props`) is conform, I do not think so as it is still the definition of a new interface at the end of the day. This is also a strength in some way since it allows users to add new properties to their Component definition.

I'd assume this means transitioning to the default type of Astro.props being Component.props instead of Props & {}

i.e. not merging in the Record<string, any> type and requiring all props and slots to be explicit

PR #323, opened 2022-10-02
`README.md`: snowpack --> withastro

mrienstra:
And link `[core team]` as per https://github.com/vuejs/rfcs

Re: linking `[core team]`: I took a look at https://github.com/vuejs/rfcs/blob/master/README.md and noticed that they linked some but not all mentions of "core team", it seemed a little random (no pattern that made sense to me), so I just linked them all, but I could also see linking only the first instance in each section.


jasikpark (approved):
lgtm, thanks for fixing those old links

natemoo-re (approved):
Thanks for updating these!

PR #326, opened 2022-10-03
App setup RFC

natemoo-re:
- Start Date: 2022-10-03
- Status: Draft

## Summary

This RFC proposes a unified interface that provides a hook into island **setup**. User-provided hooks will run for both server rendering and client-side hydration.

The primary motivation is unblocking `vue.use`, but there are other use cases as well.

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/app-setup/proposals/0000-app-setup.md)

### DIFF
--- a/proposals/0000-app-setup.md
+++ b/proposals/0000-app-setup.md
@@ -1,0 +1,184 @@
+- Start Date: 2022-10-03
+- Reference Issues: <!-- related issues, otherwise leave empty -->
+- Implementation PR: <!-- leave empty -->
+
+# Summary
+
+This RFC proposes a unified interface that provides a hook into island **setup**. User-provided hooks should run for both server rendering and client-side hydration.
+
+# Example
+
+This proposal introduces a new `appEntrypoint` option to the official `@astrojs/*` framework integrations, if relevant.
+
+```js
+// astro.config.mjs
+import { defineConfig } from 'astro/config'
+import vue from '@astrojs/vue'
+
+export default defineConfig({
+  integrations: [
+    vue({
+      appEntrypoint: '_app.js'
+    })
+  ]
+})
+```
+
+```js
+// src/pages/_app.js
+import i18nPlugin from '../plugins/i18n'
+
+export default ({ app }) => {
+  app.use(i18nPlugin, {
+    greetings: {
+      hello: 'Bonjour!'
+    }
+  })
+}
+```
+
+# Motivation
+
+Currently, Astro does not provide any way to hook into the lifecycle of an island. While this causes some annoyances for Provider-based frameworks, it makes common patterns in plugin-based frameworks impossible.
+
+The ability to "setup" the application state is extremely important for:
+
+- Vue users, who are familiar with libraries that use plugins (`vue.use`) or expose components (`vue.component`)
+- React/Preact users, who often need every component wrapped in a custom `Provider` to make working with global data (i18n, themes) easier
+- Anyone that may want to render their application with initial user state
+
+> **Warning**:
+> This RFC explicity does **NOT** concern itself with **sharing state between islands**.
+
+# Detailed design
+
+## `appEntrypoint` support in Renderers
+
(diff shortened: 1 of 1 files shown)


proposals/0000-app-setup.md
```diff
@@ -0,0 +1,184 @@
+- Start Date: 2022-10-03
+- Reference Issues: <!-- related issues, otherwise leave empty -->
+- Implementation PR: <!-- leave empty -->
+
+# Summary
+
+This RFC proposes a unified interface that provides a hook into island **setup**. User-provided hooks should run for both server rendering and client-side hydration.
```
matthewp, line 7: Can you define what setup is?

proposals/0000-app-setup.md
```diff
@@ -0,0 +1,184 @@
...
+}
+```
+
+# Drawbacks
+
+This is a relatively light feature to implement, but it does come with some gotchas:
+
+- This proposal exposes lower-level control to users. Fine for advanced use cases, but makes it easier to introduce performance footguns if not used carefully.
+- This proposal introduces a new pattern that, while modelled after similar patterns in other frameworks, has a unique design due to Astro's constraints
+- This proposal **DOES NOT** solve a common problem for sharing state via `Provider`s. It's extremely likely that people will expect this feature to do that because you can setup a Provider. However, Providers in this proposal **ARE NOT** stateful between islands—updates to one Provider instance will not automatically propogate to other Provider instances unless specifically architected to do so by the user.
```
matthewp, line 156: I understand the desire to avoid feature-bloat, but presumably you are thinking about this. Shared context in React is one very desirable feature. We can skip that for now, but it would be good if you have an idea of what that might look like so that this feature can be a stepping stone and that is not a completely different feature, if that makes sense.

proposals/0000-app-setup.md
natemoo-re: Rephrased this! Hopefully it is more clear now.

proposals/0000-app-setup.md
natemoo-re: Yeah that totally makes sense. This proposal is certainly not blocking an implementation of shared context in the future, since a prototype of that is what I extracted this proposal from. I was hoping to avoid this conversation being focused on shared state because there's still a lot to figure out there, but users asking to hook into `vue.use` comes up almost every day.

natemoo-re: I would even be open to not implementing this `appEntrypoint` proposal for React/Preact until shared state lands. It's possible this might just be a footgun until Providers can be singletons that only run once.

matthewp: I think holding off on React makes sense. I haven't seen as much of a demand for the "pyramid of providers" pattern as there has been for Vue .use() so far.

L1lith: It would be quite neat if there was a way to specify setup code that only gets loaded on server side or client side (like in the game engine melonJS, it would be cool to render the game to a static image on the server side, render that, then on the client side replace it with the actual game)

L1lith:
It would be quite neat if there was a way to specify setup code that only gets loaded on server side or client side (like in the game engine [melonJS](https://github.com/melonjs/melonJS), it would be cool to render the game to a static image on the server side, render that on the initial page load, then on the client side replace it with the actual game after the page has mounted)

See Also: https://github.com/melonjs/melonJS/issues/1134 (to me this issue appears quite difficult to implement atm)

natemoo-re:
> It would be quite neat if there was a way to specify setup code that only gets loaded on server side or client side (like with the game engine [melonJS](https://github.com/melonjs/melonJS), it would be cool to render the game to a static image on the server side, render that on the initial page load, then on the client side replace it with the actual game after the page has mounted)
> 
> See Also: [melonjs/melonJS#1134](https://github.com/melonjs/melonJS/issues/1134) (to me this issue appears quite difficult to implement atm)

This should be possible now, even without a feature like this! Feel free to ping me on Discord and I can walk you through it.

PR #347, opened 2022-10-18
Add testing section to the RFC template

matthewp:
This is not a new RFC but a proposed change to the template.

It adds an explicit section on testing strategies. As Astro grows the cost of new features also grows. Knowing how a feature will be tested gives us greater confidence that the feature will be easy to maintain and fewer regressions will occur. This also forces the RFC creator to think through edge cases they might otherwise gloss over.


FredKSchott (approved):
+1!

PR #362, opened 2022-10-25
Prerender API

natemoo-re:
- Start Date: 2022-10-25
- Status: [Experimental PR](https://github.com/withastro/astro/pull/5297)

## Summary

- Users should be able to deploy an Astro server without sacrificing the speed or cacheability of static HTML.
- The Prerender API (previously "Hybrid Output") allows users to statically prerender specific `pages/` at build time.
- The API surface is relatively small: include `export const prerender = true` in any file in the `pages/` directory that should be prerendered

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/hybrid-output/proposals/0000-prerender-api.md)


proposals/0000-hybrid-output.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Adapter Support
+
+An adapter will be required when building for `hybrid` output, because we will be building a server with some statically generated routes. Adapters will need to be updated to handle `hybrid` output.
+
+# Testing strategy
+
+This implementation will largely be tested with full fixture-based integration tests, because we will need to verify the build output is structured correctly, which will be difficult to do with unit tests.
+
+The Babel plugin that detects `export const output` will be unit tested against all valid `export` formats outlined above.
```
matthewp, line 86: Will it be tested against the _invalid_ formats too?

proposals/0000-hybrid-output.md
```diff
@@ -0,0 +1,133 @@
...
+```
+
+## Internal API
+
+Astro needs to detect/understand the `export const output` syntax. This should be implemented as a specific post-processing Babel plugin. The plugin will detect the existence of valid ESM exports of `output` (`export const output`, `export let output`, `export var output`, `export { output }`, etc) and attach the statically detected `output` value to the module metadata under the `astro` namespace. This will allow us to track the `output` mode of any public route.
+
+> **Note**
+> Astro Routes include `.astro` pages, but also API endpoints (`src/pages/api/user.ts`)! It's important that this RFC handles both of these constraints in a uniform way, hence the use of a post-processing Babel plugin rather than implementing this feature in `@astrojs/compiler`.
+
+The **build** API needs to be updated to handle the new `'hybrid'` output mode. Implementation-wise, this will be similar to the existing `'server'` implementation, but it will also adopt portions of the `'static'` build to statically generate any routes that are not handled at request time.
```
matthewp, line 74: It would be good to have a little more detail on implementation here. It's a little unclear to me how this will be broken apart in the build (although I have no doubt that it can be).

proposals/0000-hybrid-output.md
natemoo-re: Yep, updated!

proposals/0000-hybrid-output.md
natemoo-re: Agreed, MVP is still in progress, but I'm hoping to flesh out this section a lot more in the coming days.

proposals/0000-hybrid-output.md
```diff
@@ -0,0 +1,139 @@
...
+import { defineConfig } from 'astro/config'
+import node from '@astrojs/node'
+
+export default defineConfig({
+  output: 'hybrid',
+  adapter: node()
+})
+```
+
+For individual routes, `'static'` will be the default. To opt-in to `'server'` output, users may `export const output: 'static' | 'server'`.
```
matthewp, line 23: Are we sure that this is the right default? If this turns out to be wrong then it's going to be hard to fix later.

tony-sull:
I'm wondering how this will work for dynamic routes like `[slug].astro` or `[...path].astro`, would `getStaticPaths()` work to build static pages for a subset of matching routes?

proposals/0000-hybrid-output.md
```diff
@@ -0,0 +1,139 @@
...
+# Example
+
+This RFC proposes a new, opt-in `output` mode named `'hybrid'`.
+
+```js
+import { defineConfig } from 'astro/config'
+import node from '@astrojs/node'
+
+export default defineConfig({
+  output: 'hybrid',
```
tony-sull, line 18: Could we actually leverage this as the default value?  So `output: "server"` will build all pages for SSR - individual pages can opt-out with `export const output = "static"`

proposals/0000-hybrid-output.md
```diff
@@ -0,0 +1,139 @@
...
+  adapter: node()
+})
+```
+
+For individual routes, `'static'` will be the default. To opt-in to `'server'` output, users may `export const output: 'static' | 'server'`.
+
+```astro
+---
+export const output = 'server'
```
tony-sull, line 27: will it be possible for our language tools to give type checking and auto-complete as an `export const`?

proposals/0000-hybrid-output.md
```diff
@@ -0,0 +1,139 @@
...
+
+# Adoption strategy
+
+- Adoption will be entirely opt-in by setting `output: 'hybrid'` in your `astro.config.mjs` file.
+- This is not a breaking change, it is new behavior
+- Third-party adapters will need to coordinate changes to support `hybrid` output, but for the most part the logic will be very similar to the existing `server` output.
+
+# Unresolved questions
+
+- Should API Endpoints default to `output = "server"`? This seems more ergonomic than manually opting every API Endpoint into `'server'` output...
```
tony-sull, line 139: RSS may be the example where `output = "static"` still makes sense for the default in an API endpoint

proposals/0000-hybrid-output.md
```diff
@@ -0,0 +1,139 @@
...
+The Vite plugin that detects `export const output` will be unit tested against all valid `export` formats outlined above, as well as invalid dynamic exports.
+
+# Drawbacks
+
+- A new output mode will increase the complexity and maintenance costs of our `build` and `adapter` implementations.
+- The `export const output` syntax is JS-like, but since it requires static analysis, it does not support the full dynamicism of JS that users might expect. Clear errors will help ease this problem.
+- Having a separate mode for `hybrid` increases cognitive overhead and the learning curve of Astro.
+- `hybrid` output should arguably be the default `output` mode (similar to Next.js), but to avoid a breaking change we're going with an opt-in adoption approach
+
+# Alternatives
```
stramel, line 95: For another alternative, has any thought been given to using the filename to indicate the output? (ie. `.server.astro`)

tony-sull, line 95: Might be nice to add cache headers as an alternative here, even if just to show how many hoops have to be jumped through to properly handle caching in a CDN via headers, it's not the easiest balancing act 😅

panwauu:
**Maybe** we should keep in mind to allow the syntax to support other future modes (other than `static` or `server`). 
These are just some thoughts. Of course, we should go for the easiest solution that provides the needed flexibility.

I have seen some interest (myself included) in ISR. This is currently mostly solved with cache headers, which can be painful IMHO.  ISR would require an option to set the rebuild cycle. This would not really be compatible with the current proposal. One way with the current proposal would be to encode the cycle in the string. This would require parsing and make the typing more complex. It does not seem clean at all.
```astro
---
export const output = 'isr-1000' // `isr-{cycleInMs}`
---
```
Or we could add the cycle as an option to add per file. Which would add another option only for ISR.
```astro
---
export const output = 'isr'
Astro.response.isrCycle = 1000
---
```

proposals/0000-hybrid-output.md
panwauu: Could Astro be always `hybrid`? Such that the `output` option in the `astro.config` file sets the default output for each file and the default mode can be overridden by exporting the `output` from the frontmatter. 🤔

- This would be the most flexible approach imho and also would be quite logical and reducing the learning curve. 

- I am totally unclear how the output would look like and how we would handle adapters as we always need an ssr-adapter to do anything on the server. E.g. we would always export a directory with `client` and `server` folders, which is weird if there is no ssr happening.

proposals/0000-hybrid-output.md
panwauu: I completely agree that Astro should just always be hybrid!
In `astro.config` the `output` option could be renamed to `defaultOutput` to set the default for the hybrid mode. The default mode can be overridden by exporting the `output` from the frontmatter. 🤔

This would be the most flexible approach imho and also would be super logical and not change the learning curve.

- I am totally unclear how the output would look like and how we would handle adapters as we always need an ssr-adapter to do anything on the server. E.g. we would always export a directory with client and server folders, which is weird if there is no ssr happening.
- I understand that this would mean a breaking change, but still worth a though.

matthewp:
As from the call, I think we have multiple concepts we are grouping under `output` but are actually 2 or 3 different things.

- `output` is how Astro builds the site.
  - `static` build to HTML
  - `server` build to JS that can be run on a server.

Then there's this other thing called `edge` which supposedly builds for an edge function. But an edge function is a `server` too. The difference between the two is more cultural than real.

imo, when you specify `edge` vs. `server` that's really a signal to the adapter (or to multiple adapters).

So in my mind we have two concepts, one is a signal to Astro and one is a signal to adapters. We could call this other thing `mode` or `target` or `runtime` or `layer` or something else. But I think this other thing is configuration that needs to be set in each route.

proposals/0000-hybrid-output.md
```diff
@@ -0,0 +1,139 @@
...
+- Reference Issues: TODO
+- Implementation PR: <!-- leave empty -->
+
+# Summary
+
+Currently, Astro's [`output`](https://docs.astro.build/en/reference/configuration-reference/#output) enables developers to switch between `'static'` (default) and `'server'` outputs. One of our most requested features is to enable route-level control of `output` so that some routes can be server-rendered while others are statically generated.
+
+# Example
+
+This RFC proposes a new, opt-in `output` mode named `'hybrid'`.
```
FredKSchott, line 11:
```
output: 'hybrid'
```

I'd strongly push for this to remain a feature of output: 'server'. Basically, that we don't introduce a new third mode but instead add this new `export const output` support to anyone with `output: 'server'` enabled, that will let you prerender pages. 

Conceptually, this makes more sense to me. I'd always seen `server` vs. `static` more as a boolean: "does this build output include a server or not?" There are huge differences between the two that we need to build for and that you need to account for as a user deploying your site. But, this feature doesn't really change that story.

As both an implementation and a user experience, I think this makes more sense as a feature of `output: server` vs. a thing that is different from it. As I understand the RFC, the `hybrid` build output and deploy story are both is identical to `server`.

FredKSchott: Implementation-wise, there are a lot of reasons to consider shipping `export const output` support as a feature of `output: 'server'`:

**DX (Defaults):** By making this an opt-in story for `output: 'server'`, you solve for all of the drawbacks and complexities in this RFC related to default behavior concerns and migration strategy. I think this is a deeply tricky part of the current RFC that doesn't have a good answer, since "hybrid" doesn't tell you the user's default intention as much as "static" and "server" do. Because of that, some group of users will always be unhappy with the "hybrid" defaults not being what they expected.

**DX (API Simplicity):** Introducing a third mode requires more documentation and will lead to more user questions. When do I want to use hybrid mode? How does X work in hybrid mode? How does Y work in hybrid mode? I'm having problem Z, does switching to hybrid mode solve that?

**Maintenance:** I think its fair to say that having to support two different top-level "modes" is one of our biggest maintenance overheads today. Ex: Anytime we ship a new feature, we need to 2x our testing to make sure it works consistently across "static" and "server". Anytime a user reports a bug, we need to make sure we know the `output` mode in case the behavior changes across modes, etc. Adding a third top-level option to support this single `export const output` feature would increase this complexity.

proposals/0000-hybrid-output.md
```diff
@@ -0,0 +1,139 @@
...
+> **Note**
+> Astro Routes include `.astro` pages, but also API endpoints (`src/pages/api/user.ts`)! It's important that this RFC handles both of these constraints in a uniform way, hence the use of a post-processing Vite plugin rather than implementing this feature in `@astrojs/compiler`.
+
+The **build** API needs to be updated to handle the new `'hybrid'` output mode. Implementation-wise, this will be similar to the existing `'server'` implementation, but it will also adopt portions of the `'static'` build to statically generate any routes that are not handled at request time.
+
+The **dev** server must be updated to allow `searchParams` and other features which are currently only avaliable with `server` rendering to be allowed on a per-route basis.
+
+## Adapter Support
+
+An adapter will be required when building for `hybrid` output, because we will be building a server with some statically generated routes. Adapters will need to be updated to handle `hybrid` output.
```
tony-sull, line 80: Will adapters be expected to do anything differently for `static` routes, or the adapter only be envolved in the SSR routes?

proposals/0000-hybrid-output.md
```diff
@@ -0,0 +1,139 @@
...
+
+The eventual solution should enable granular control over `output` on a per-route basis. Since Astro defaults to `'static'` output, this should remain the default in the eventual solution as well.
+
+# Detailed design
+
+## Public API
+
+There are two public API surfaces to this RFC:
+
+- Updating the `astro.config.mjs` to accept a new `output` value, `'hybrid'`. The existing values of `'static' | 'server'` would remain supported. A value of `'hybrid'` will enabled detection for the following feature:
```
tony-sull, line 49: I'm still not totally sold on whether we need a build-level solution here or if cache headers will cover most use cases

Cache headers are really tricky to deal with, especially when if I'm living in an SSG world most days.  I could see a nice first step here being a nicer API for setting cache headers that adds better DX with hints on the different cache header options

```js
Astro.cookies.cache({ duration: '1 year', public: true, httpOnly: true })`
```

Not sure exactly what the final API would need to be, but in this idea our type definitions could help explain when the `public` or `httpOnly` flags should be used.  We kept the `"1 year"` shorthand out of the `Astro.cookies` API to start with, this might be a good time to pull that in 🤔 

**Pros**

- no build or adapters changes needed
- `static` and `server` work as-is today
- we could add a proper hybrid mode as a follow-up if caching doesn't cut it for enough users, the `Astro.cookies.cache` helper would still be handy

**Cons**
- still has the first request performance hit, this dodges the true hybrid use case
- improving the DX of cache headers may be a lost cause...

proposals/0000-hybrid-output.md
```diff
@@ -0,0 +1,139 @@
...
+
+```astro
+---
+// Valid – constant string literal
+export const output = 'server'
+// Invalid – cannot be determined at compile-time
+export const output = static ? 'static' : 'server';
+export const output = value;
+export const output = 'ser' + 'ver';
+export let output = undefined; output = 'server';
```
matthewp, line 63: I would assume this is not allowed either: `export let output = 'server'`. Allowing let/var means someone might try to mutate it. Since that's harder to restrict I think it makes sense to require the user of `const`.

proposals/0000-hybrid-output.md
jacobwgillespie: > I'm still not totally sold on whether we need a build-level solution here or if cache headers will cover most use cases

My use-case for a build-level solution is to pre-build certain resources and then omit the dependencies necessary to build those resources from the production server. For instance, I generate some social images by rendering them using Puppeteer, but don't want to ship that dependency or the large browser it requires to prod. Having a way to generate those routes at build-time and then omit unused dependencies would be ideal.

(this applies to large dependencies, but also dependencies that are Node-only if I'm deploying to a non-Node server runtime)

kyr0:
Hi all, 

I just wanted to share with you, that my recent learning was that having SSR and SSG pages split by folders might also be an idea worth thinking about (instead of having them split per route with a new mode). It feels quite nice from the DX point of view.

I spiked this in a different way tho... with the current version 1.6.2, having different projects combined in one deployment; of course: Different projects aren't optimal and I had to come up with some solutions behind the scenes to get it working smoothly, but it does...; 

My learning was that it feels quite clear, intuitive and nice to have SSR and SSG clearly split "at first sight". It also helps to understand separating the concerns and could lead to less boilerplate when the developer doesn't need to specify the type of execution per route. 

If you wanna play with this:

Here is the impl: https://github.com/kyr0/turbo-hybrid-astro-on-vercel#welcome-to-turbo-hybrid-astro-on-vercel
Tech demo / live: https://turbo-hybrid-astro-on-vercel.vercel.app/ (pls scroll down and click on the cards, they lead you to the different pages. It is all deployed on the same Vercel deployment and works seamlessly in dev and build)

I also explained my spike results/learnings here:
https://github.com/kyr0/turbo-hybrid-astro-on-vercel#spike-result

natemoo-re:
Thanks for all the feedback, everyone! I am going to update this RFC to take all of it into account and we'll be chatting about this again in a more focused RFC-specific call this week.

For a quick summary of the expected changes:
- Adding an additional `hybrid` output is unnecessary complexity for many reasons (mentioned above)
- This RFC will shift focus to enabling `static` pre-rendering when using `output: 'server'`. 
  - The default will be `'server'` unless otherwise specified.
  - Dependencies for static routes will _not_ be bundled in with the final server.
- There are still some open questions about how adapters should handle statically generated pages
- Per @kyr0's explorations, there's clearly some interest in configuring entire directories at once. I don't want to increase the scope of this RFC any more, but I think we could address this with a follow-up RFC.

proposals/0000-hybrid-output.md
natemoo-re: Open question, but my thinking is that adapters could safely ignore statically pre-rendered routes? They're essentially just `.html` files treated as assets.

proposals/0000-hybrid-output.md
natemoo-re: Thanks @jacobwgillespie! I think that use case makes a very compelling argument for _not_ doing this solely through cache headers, though we definitely have discussed that as another API that would be generally useful.

proposals/0000-hybrid-output.md
natemoo-re: That seems fair!

proposals/0000-hybrid-output.md
natemoo-re: Yes, this is the direction we're taking in the latest version!

proposals/0000-hybrid-output.md
natemoo-re: Probably possible! Can add a note about this.

proposals/0000-hybrid-output.md
natemoo-re: Added some comments about this in the `alternatives` section.

tony-sull:
@natemoo-re one question I hadn't thought about until this morning, how will this work for integrations or components that either work differently in `static` vs `server` mode, or only support one build mode?  Will an integration/component be able to tell what context it for a specific page being rendered?

Two examples that came to mind for me:

- `@astrojs/image` sets up the build a bit differently for `static` vs `server` and renders `src` differently depending on the output target
- not sure if this is still the case, but at one point `astro-icon` didn't support SSR and was best used for SSG sites

natemoo-re:
@tony-sull Hopefully this is outlined better in the updated RFC, but since this mode is now being treated the same as `server`, adapters will likely just work the same as before. If there is some post-processing that needs to happen on the `.html` files, those will be present in the manifest.

FredKSchott:
@tony-sull image will be an interesting one to think through, but I think `server` mode would still work as-is even for prerendered pages (worth confirming though).

smithbm2316:
I really am a huge fan of this proposal and think that either Alternative option F or @kyr0's proposal would be the options that I favor the most for this feature.

I have no idea how complicated the implementation internally would be for either of these, but I particularly like the idea of alternative F that allows you to name a file with a `.server.astro` or `.static.astro`, similar to how Remix does that. I think it makes it immediately clear what the rendering strategy is for that file and makes it hard to be confused about it, unlike situations that I could see arising with `export const output = 'static'` where you no longer are able to use that variable name in your JS/TS for the file. It might not be immediately obvious to beginners that the variable there does anything special and could be a tripping up point. 

Alternative C I think is also a decent option because then you're only ever using 1 specific named value in the Astro template, and you can extend that object with as many per-page options down the line with no breaking changes required. Feels like a better version of the current proposal with none of the future potential downsides.

Next.js 13 beta recently added quite a few new named variable exports in their files for configuring caching that I just am not a fan of, as I think that for beginners it's gonna be really hard to discover that those values mean something specific. It's nice if you know how it works because it's terse and quick to write, but I think it just is really hard to google/search for answers to if you're a beginner. If there's gonna be a named export that means something, I usually think using an object or function (like `loader`/`action` functions or `meta`/`links` objects in Remix) for it makes it more clear that those might be specific meta-framework features and not just something that a developer has created themselves (hence why i'd prefer alternative C if using a named export variable).

Looking forward to see how this RFC develops!

proposals/0000-hybrid-output.md
Princesseuh: Confirming that it's possible! We'll have to either provide a type for it, or fake completions / diagnostics (like NextJS does)

proposals/0000-hybrid-output.md
natemoo-re: Updated the RFC to address this question. Adapters don't need to do anything differently, the generated assets will be included in the asset manifest and should be picked up automatically.

PR #363, opened 2022-10-25
RFC: throwable response

xstevenyung:
- Start Date: 2022-10-25
- Status: Draft

## Summary

Allows to stop execution of an endpoint or accessing a page with Astro SSR in nested functions by utilizing similar API to Remix with `throw Response`.

## Links

- [Full Rendered Proposal](https://github.com/xstevenyung/rfcs/blob/feat/thrown-responses/proposals/0027-throwable-response.md)


matthewp:
The obstacle here is that with streaming the response is already sent out after the page component's frontmatter runs. If you were to throw in a component inside of the page component it would be too late. I'm not sure how we can address this problem.

PR #381, opened 2022-11-10
Relative links RFC

matthewp:
- Start Date: 2022-11-09
- Status: Draft

## Summary

Add support for relative links written out during the build, such as `<link rel="stylesheet" href="../assets/main.hash.css">` through a new build configuration.

## Links

- [Full Rendered Proposal](https://github.com/withastro/rfcs/blob/98d2df0f3332e8d77393fe4c61780f6e36559fce/proposals/base-relative-assets.md)

### DIFF
--- a/proposals/base-relative-assets.md
+++ b/proposals/base-relative-assets.md
@@ -1,0 +1,132 @@
+- Start Date: 2022-11-09
+- Reference Issues: <!-- related issues, otherwise leave empty -->
+- Implementation PR: <!-- leave empty -->
+
+# Summary
+
+With this proposal it will be possible to build Astro sites that can be deployed to any path on any domain regardless of the base. This will be done through an new config option `build.linkStyle`.
+
+With this option you can tell Astro to build links (for assets like scripts an styles) to be absolute (the default and current behavior) or relative (relative to the base).
+
+# Example
+
+By default Astro will continue to add links with absolute paths from the `base`. For example with this config:
+
+```js
+import { defineConfig } from 'astro/config';
+
+export default defineConfig({
+  base: '/docs',
+  build: {
+    linkStyle: 'absolute'
+  }
+});
+```
+
+Your CSS will build to this:
+
+```html
+<link rel="stylesheet" href="/docs/assets/main.hash.css">
+```
+
+Using the `relative` config option:
+
+```js
+import { defineConfig } from 'astro/config';
+
+export default defineConfig({
+  build: {
+    linkStyle: 'relative'
+  }
+});
+```
+
+You'll instead get this:
+
+```html
+<link rel="stylesheet" href="../assets/main.hash.css">
+```
+
+# Motivation
+
+We often get support questions on how to deploy an Astro site without knowing the base in advance. There has been discussions and RFCs to solve this problem such as [this one](https://github.com/withastro/rfcs/discussions/334) and [this one](https://github.com/withastro/rfcs/discussions/23).
+
+People have tried to use `./` in the `base` option to make this happen to, such as in [this issue](https://github.com/withastro/astro/issues/4229).
+
+The primary use-case is wanting to deploy a site that might live under multiple subdomains. Using relative links makes it portable.
+
+# Detailed design
+
+This new configuration option will affect these types of assets:
+
+- Link tags added by us from bundled CSS.
+- Hoisted script tags added by us from bundled hoisted scripts.
+- Links to scripts added to `<astro-island>`s.
+
+Any other type of asset, such as images, are not controlled by Astro and will continue to not be so. The developer will need to make these links relative themselves. They can also use the `<base>` tag to make this easier.
+
+## Implementation
+
+During the build when we go to generate pages there are a few places where we resolve the final URL for links.
+
+- For link tags it is done [here](https://github.com/withastro/astro/blob/c0cb65b18db888ed46fe9a0b7d66d062de245f5e/packages/astro/src/core/build/generate.ts#L300).
+- For hoisted scripts it is [here](https://github.com/withastro/astro/blob/c0cb65b18db888ed46fe9a0b7d66d062de245f5e/packages/astro/src/core/build/generate.ts#L301)
+- For links added to astro-islands it [is here](https://github.com/withastro/astro/blob/c0cb65b18db888ed46fe9a0b7d66d062de245f5e/packages/astro/src/core/build/generate.ts#L352).
+
+We should generalize this logic so that there's one function that gets called to resolve the URL given the `linkStyle` option, as well as the base and subpath. That function will use the `linkStyle` config to determine which algorithm to use.
+
+For SSR the same things are done [here](https://github.com/withastro/astro/blob/c0cb65b18db888ed46fe9a0b7d66d062de245f5e/packages/astro/src/core/app/index.ts#L63) for example. Again, this should be consolidated so we're not doing the same sort of logic in many different places, creating areas for bugs. Instead SSG and SSR should use shared logic.
+
+## 404.astro
+
+If using `linkStyle: 'relative'` a 404.astro page __will not be allowed__. This is because it is impossible to make these links be relative in SSG mode. That's because a 404.astro might be used by paths such as:
+
+- /foo/
+- /foo/bar/
+- /foo/bar/baz/
+
+If we statically generate relative links then these will be incorrect. So we will throw an error if using a 404.astro page with `'relative'` and suggest using `'absolute'` instead.
+
+## Imported images
+
+Images that are imported will *not* be relative. That means this:
+
+```astro
+---
+import penguinUrl from '../assets/penguin.png';
+---
+
+<img src={penguinUrl} />
+```
+
+It might be *possible* to make this work through a Vite plugin but would be a bit of work and might be error prone. 
+
+# Testing strategy
+
+Given that resolving URLs will be done in a few places, it makes sense to create a module within `src/core` for this algorithm. This module will be unit testable for all of the scenarios. We should probably also move the runtime resolver (which is used internally by the runtime for producing `<astro-island>` URLs) since it is shared logic between SSG and SSR that is related.
+
+Additionally some fixture tests will be included for the matrix of scenarios; SSG and SSR mode, `absolute` and `relative` options, as well as the types of assets; stylesheets, hoisted scripts, and astro islands.
+
+# Drawbacks
+
+- This adds some complexity to the codebase in the usual way that adding a config option does.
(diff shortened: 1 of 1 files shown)


proposals/base-relative-assets.md
```diff
@@ -0,0 +1,132 @@
...
+# Adoption strategy
+
+Please consider:
+
+- Given that this is a common support topic we would want to create documentation around this new feature and probably some sort of guide on how to create Astro pages that use base relative links.
+- `absolute` still makes sense as a sane default. Given the 404 limation of `relative` I don't think we would ever make it the default.
+
+# Unresolved questions
+
+It's unclear to me how/if the Image component in `@astrojs/image` can build relative URLs. It's possible that it *can*, given that it knows the current page URL. That's something to be explored as supporting relative links there would be a good answer for the `<img>` limitation.
```
delucis, line 132: Yes, I was also thinking I’d expect the image integration to be compatible with this config setting.

Could be a good thing to bear in mind so that the tools needed to build relative paths are available to integrations as well as internally.

proposals/base-relative-assets.md
```diff
@@ -0,0 +1,132 @@
...
+
+```html
+<link rel="stylesheet" href="../assets/main.hash.css">
+```
+
+# Motivation
+
+We often get support questions on how to deploy an Astro site without knowing the base in advance. There has been discussions and RFCs to solve this problem such as [this one](https://github.com/withastro/rfcs/discussions/334) and [this one](https://github.com/withastro/rfcs/discussions/23).
+
+People have tried to use `./` in the `base` option to make this happen to, such as in [this issue](https://github.com/withastro/astro/issues/4229).
```
delucis, line 54: We might want a warning for when people do use `./` in `base` telling them to use `build.linkStyle` instead?

delucis (commented):
Thanks for writing this up @matthewp! I can definitely see the benefit here and agree with this being an opt-in thing.

proposals/base-relative-assets.md
bluwy: I'm probably here to chime in that I prefer `base: './'` instead and not have a new option. The functionality overlaps so collapsing it might be easier to understand for the end-user. We can split this internally if it simplifies the internal code.

proposals/base-relative-assets.md
```diff
@@ -0,0 +1,132 @@
...
+
+```astro
+---
+import penguinUrl from '../assets/penguin.png';
+---
+
+<img src={penguinUrl} />
+```
+
+It might be *possible* to make this work through a Vite plugin but would be a bit of work and might be error prone.
```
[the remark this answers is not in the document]
bluwy: If we set `base: './'` in Vite's config, this should work today so I'm not sure if this part is a problem unless I'm missing something.

As long as an asset is imported, Vite can generate a relative link dynamically based on the current chunk.

proposals/base-relative-assets.md
```diff
@@ -0,0 +1,132 @@
...
+- For hoisted scripts it is [here](https://github.com/withastro/astro/blob/c0cb65b18db888ed46fe9a0b7d66d062de245f5e/packages/astro/src/core/build/generate.ts#L301)
+- For links added to astro-islands it [is here](https://github.com/withastro/astro/blob/c0cb65b18db888ed46fe9a0b7d66d062de245f5e/packages/astro/src/core/build/generate.ts#L352).
+
+We should generalize this logic so that there's one function that gets called to resolve the URL given the `linkStyle` option, as well as the base and subpath. That function will use the `linkStyle` config to determine which algorithm to use.
+
+For SSR the same things are done [here](https://github.com/withastro/astro/blob/c0cb65b18db888ed46fe9a0b7d66d062de245f5e/packages/astro/src/core/app/index.ts#L63) for example. Again, this should be consolidated so we're not doing the same sort of logic in many different places, creating areas for bugs. Instead SSG and SSR should use shared logic.
+
+## 404.astro
+
+If using `linkStyle: 'relative'` a 404.astro page __will not be allowed__. This is because it is impossible to make these links be relative in SSG mode. That's because a 404.astro might be used by paths such as:
```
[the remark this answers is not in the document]
bluwy: I've think long and hard about this but I can't seem to find a nice solution too. Unless we get hybrid support soon maybe then it could work as a middle ground.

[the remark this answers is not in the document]
FredKSchott: And just as a sanity check: `<base>` can't help us here right? Or is it that the user has to manually add the base tag themselves, which is a footgun.

```
  <base href="/">
```

proposals/base-relative-assets.md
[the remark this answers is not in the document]
matthewp: @bluwy Relative to the file that's importing it though, right? Not relative to the page. Vite is not aware of the page that an import is being used on (it can be used on multiple pages!). So either way you as a user need to convert it to a relative link.

Even without this, we could have a directive like `relative:src` that uses the current URL and converts the provided absolute URL into a relative one.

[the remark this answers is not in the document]
bluwy: I did some test locally and seems like you might be right. Looks like Vite doesn't handle `base` in SSR at all too. Maybe Vite's SSR manifest or [advanced base option](https://vitejs.dev/guide/build.html#advanced-base-options) could help here, but this seems like a tricky one indeed.

proposals/base-relative-assets.md
matthewp: Do you prefer `./` here because that's what Vite does? Does this work in a multi-page app for them? I'd love to see what this looks like.

bluwy: Yeah I partly prefer `./` because of Vite, and the idea of "the least config option" 😬 But I could be persuaded to `linkStyle` if it supports something that `./` can't.

`base: './'` does work in an MPA today but only for the client-side ([stackblitz](https://stackblitz.com/edit/vitejs-vite-nym34j?terminal=dev)).

matthewp:
@kelvin-zhao Are you ok with not having support for `404.astro` in SSG mode with this proposal? I still haven't thought of a way of avoiding that restriction.

The other thing missing is a good story around adding `<a href>`, ``, etc. I'm currently thinking about doing directive like `<a relative:href="/other/page">` which would be transformed to `<a href="../other/page">`.

proposals/base-relative-assets.md
```diff
@@ -0,0 +1,132 @@
...
+
+Given that resolving URLs will be done in a few places, it makes sense to create a module within `src/core` for this algorithm. This module will be unit testable for all of the scenarios. We should probably also move the runtime resolver (which is used internally by the runtime for producing `<astro-island>` URLs) since it is shared logic between SSG and SSR that is related.
+
+Additionally some fixture tests will be included for the matrix of scenarios; SSG and SSR mode, `absolute` and `relative` options, as well as the types of assets; stylesheets, hoisted scripts, and astro islands.
+
+# Drawbacks
+
+- This adds some complexity to the codebase in the usual way that adding a config option does.
+  - However this is also an opportunity to clean up the code and have just 1 algorithm for determing how a URL gets resolved that we create.
+- This will not fix the problem of `<img>`s you create. this is especially important in Markdown files, where manually adding relative links is required. So people might find this solution inadequate since it doesn't fix all of your problems.
```
matthewp, line 114: @delucis since you know a lot about custom md syntax, is it possible to have a syntax that would be a little like a directive is in Astro where the user can say "this should be relative" for an image?

kelvin-zhao:
@matthewp Tbh, I'm okay with any possible option... Currently I'm building the files, then manually go into different folders to update the paths on all links... It's so painful I'm halfway onto switching entire projects to other frameworks.

falco467:
> @kelvin-zhao Are you ok with not having support for `404.astro` in SSG mode with this proposal? I still haven't thought of a way of avoiding that restriction.
> 
> The other thing missing is a good story around adding `<a href>`, ``, etc. I'm currently thinking about doing directive like `<a relative:href="/other/page">` which would be transformed to `<a href="../other/page">`.

The 404 Problem only exists when serving the 404 Page without redirect, right? You could also implement 404 by sending an 302 Temporary redirect with the location of 404.html and then everything would work with a relative base.

wassfila:
A note with regard to motivation, or providing a story to support the importance of this RFC.
web content management is quite opinionated towards absolute path usage for efficiency (asset exists once), while markdown content management is inclined towards independence of content and mostly its abstraction from the integration plaform.
As example, both VSCode and Github, support relative notations
```markdown
![Tree](./tree.svg)
<img src="./tree.svg" />
```
the reason I mention markdown does not imply the solution having to be markdown specific, markdown is only translated to html, so if Astro supports relative notations out of the box, by that I really mean `` and not having to add an import which requires content refactoring, I don't think content creators would apreciate. I met plenty of support posts, some desperate even fell back on dynamic relative import within the src (given the dynamic import not being handled by Vite)

The good things about relative vs absolute is that relative always starts with '.' every developer knows that, so why would we ask them to prefix with any other special custom keyword when it's already clear that it's relative.
This way hopefully, both abolute and relative can be supported without any inconvenience for the user. This might be difficult from Vite perspective, yet Introducing intermediate workaroud that gets later on rejected by the community might be harder to fix.
Please correct me if this is not the target of this RFC, I would bring it in discussions or host it in a new one.

wassfila:
before posting about relative assets management, does this RFC covers relative src `` or is its scope only for href links ?

bstro:
Are there any known hacks to get around this until official support lands? I need a relative path to css assets and I'm afraid I can't even manually edit the paths as I'm building with vercel. Perhaps a post-build npm script might work for me.

matthewp:
I'm not happy with the way this RFC makes Astro work 2 different ways. I'd prefer to find a path that works the same way in all cases. For that reason I'm going to close this PR as it stands and look into a different solution to the problem.

wassfila:
> I'm not happy with the way this RFC makes Astro work 2 different ways. I'd prefer to find a path that works the same way in all cases. For that reason I'm going to close this PR as it stands and look into a different solution to the problem.

oof, wise decision.

falco467:
@matthewp can we put a link here to an RFC with a different solution?

ixkaito:
I have published an integration and hope it will be a solution.

https://www.npmjs.com/package/astro-relative-links

nusendra:
> I have published an integration and hope it will be a solution.
> 
> https://www.npmjs.com/package/astro-relative-links

my build problem solved with this, Thanks mate

Nicochess:
> I have published an integration and hope it will be a solution.
> 
> https://www.npmjs.com/package/astro-relative-links

This lib solves the problem.

PR #392, opened 2022-11-15
[META] Add Goals/Non-Goals to RFC template

FredKSchott:
This is something that has become a pattern in official RFCs that I would like to standardize across all RFCs by breaking up the "Motivation" section.

Calling out an RFC's explicit Goals and Non-Goals help focus the conversation. Without this, often the "Motivation" section becomes more about story-telling and less about clear, explicit context setting. By calling these out and using a bulleted list format, it forces the RFC author to think through and verbalize what they are trying to accomplish with their RFC. 

While converting 1 section to 3 sections may look like adding more work, its intent is to cut down on the overall length of the RFC by encouraging the use of bulleted lists for some of the content that previously lived in "Motivation".

### DIFF
--- a/template.md
+++ b/template.md
@@ -11,18 +11,17 @@ A brief, one or two sentence explanation of the proposal.
 If the proposal involves a new or changed API, then include a basic code example.
 Otherwise, omit this section if it's not applicable.
 
-# Background
+# Background & Motivation
 
 Include any useful background detail that that explains why this RFC is important.
-What situation created the problems and RFC goals outlined below? Why now?
-Be brief, if possible!
+What are the problems that this RFC sets out to solve? Why now? Be brief!
 
 It can be useful to illustrate your RFC as a user problem in this section.
 (ex: "Users have reported that it is difficult to do X in Astro today.")
 
 # Goals
 
-A **bulleted-list** outlining the intended goals of this RFC. 
+A **concise, bulleted-list** outlining the intended goals of this RFC. 
 
 - What are the exact problems that you are trying to solve with this RFC?
 - Separate these goals from the solution that you will outline below.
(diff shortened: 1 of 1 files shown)


FredKSchott:
@matthewp I actually didn't see your post here: https://github.com/withastro/rfcs/pull/373#issuecomment-1315784781

curious to get your thoughts since this RFC came out of Ben using a different template that had the explicit Goals/Non-Goals that are brought in here. This PR tries to standardize this as a Good Pattern :tm: but to also be much more concise (bullets required to encourage clear thinking).

template.md
```diff
@@ -8,22 +8,36 @@ A brief, one or two sentence explanation of the proposal.
...
 
-> Why are we doing this? Why does it matter?
-> What use cases does it support? What is the expected outcome?
+# Goals
 
-Please focus on explaining the motivation, so that if this RFC is not accepted,
-the motivation could be used to develop alternative solutions. In other words,
-enumerate the constraints you are trying to solve without coupling them too
-closely to the solution you have in mind.
+A **bulleted-list** outlining the intended goals of this RFC.
```
matthewp, line 25: Bulleted lists constrain you to one paragraph. What happens when a longer explanation is needed? I guess you can have a 1 sentence goal in the list and then expand on it in the same section.

template.md
```diff
@@ -8,22 +8,36 @@ A brief, one or two sentence explanation of the proposal.
...
+A **bulleted-list** outlining the intended goals of this RFC. 
+
+- What are the exact problems that you are trying to solve with this RFC?
+- Separate these goals from the solution that you will outline below.
+- If this RFC isn't approved, these goals can be used to develop alternative solutions.
+
+# Non-Goals 
+
+A **bulleted-list** outlining what is intentionally left out of this RFC. 
+This gives the reader the correct context on what is and isn't meant to be considered:
```
matthewp, line 34: You can list a Non-Goal, but a reviewer can disagree with it. Otherwise Non-Goals become a way to predict and subvert possible objections. Not sure if the text needs to be updated here, just a thought.

matthewp:
I think I'm ok with the split overall. In my head a Motivation is _a little_ different. The motivation is often "lots of users ask for this" and that doesn't neatly fit into the Goal / Non-goal, but that probably doesn't need to be stated.

FredKSchott:
> 
I'm happy with the split overall. In my head a Motivation is a little different. The motivation is often "lots of users ask for this" and that doesn't neatly fit into the Goal / Non-goal, but that probably doesn't need to be stated.

True, I'll think about this a bit more. I agree that the subtext is that "Background" really exists to give evidence why the goals are important & valuable, so maybe "Motivation" is still an accurate fit. "Background & Motivation" is a bit of a mouthful but could also work.

template.md
FredKSchott: A bulleted list can still contain multi-sentence bullets if needed, and yes the "Background" section or even this section itself could contain more info to go into the details. 

But overall, it's an intentional move to push the author to summarize their goals as concisely as possible in a single place in the document.

template.md
FredKSchott: Yea, agreed that a non-goal can still be discussed and debated just like anything else in the RFC.

matthewp:
lgtm

PR #407, opened 2022-12-06
Update renderEntry with implementation details

matthewp:
Update renderEntry with new implementation.


bholmesdev (commented):
Left some edits to make sure I understand head propagation!

bholmesdev (approved):
I might call this the best Astro feature ever, but I won't get a`head` of myself

PR #435, opened 2023-01-12
[Meta] Update RFC Process

natemoo-re:
Revamps our proposal process to turn this repository into the **Astro Project Roadmap**.

- [Glossary](https://github.com/withastro/rfcs/blob/roadmap/README.md#glossary)
- [Stage 1: Proposal](https://github.com/withastro/rfcs/blob/roadmap/README.md#stage-1-proposal)
- [Stage 2: Accepted Proposal](https://github.com/withastro/rfcs/blob/roadmap/README.md#stage-2-accepted-proposal)
- [Stage 3: RFC \& Development](https://github.com/withastro/rfcs/blob/roadmap/README.md#stage-3-rfc--development)
- [Stage 4: Ship it!](https://github.com/withastro/rfcs/blob/roadmap/README.md#stage-4-ship-it)


README.md
```diff
@@ -1,172 +1,67 @@
-# Astro RFC
+# Astro Project Roadmap
 
-## What is an RFC?
+Status: Draft, In progress
```
FredKSchott, line 3: safe to remove once merged, so I think safe to remove here!

.github/DISCUSSION_TEMPLATE/proposal.yml
```diff
@@ -0,0 +1,49 @@
...
+      value: |
+        # Proposal
+
+        Welcome! 👋
+
+        If you are interested in proposing a new feature or improvement to Astro, 
+        then you are in the right place! 
+
+        This template covers most of the questions you will be asked about your proposal.
+        This template is completely optional, so feel free to delete it if you'd like and
```
FredKSchott, line 16: > "so feel free to delete it"

Is this possible using this "structured form" style of template? The idea/spirit behind the new roadmap is that structure is added as you go. And I'll admit, I'm a bigger fan of the "Markdown-with-comments" template vs. the "structured form" when the thing being written is more unstructured/freeform (like an early feature idea). 

It gives the author the ability to modify the template based on the context of what they're proposing. Or, just delete it with a quick `CMD-A + Delete`.

natemoo-re: Ah right, just copy + pasted. Yeah I'm not sure I really like the structured form versus the markdown template. It's a bit more user friendly but at the expense of the format not being consistent across stages.

natemoo-re: Switched this up so we're using the templates, but they only have this content prefilled

natemoo-re:
Docs involvement?

stage-3--rfc-template.md
```diff
@@ -0,0 +1,95 @@
...
+A **concise, bulleted-list** outlining anything intentionally left out of this RFC:
+
+- Non-goal: A goal that is intentionally not addressed in this RFC.
+- Out-of-scope: A goal that is related, but intentionally avoided here.
+- Future: A goal that is related, but left to be addressed in the future.
+
+This gives the reader the correct context on what is intentionally left out of scope.
+It is okay to leave this section empty.
+
+# Detailed Design
```
natemoo-re, line 53: Split out into **Detailed Design** and **Detailed Usage / API reference** (for different audiences)?

stage-3--rfc-template.md
```diff
@@ -0,0 +1,95 @@
...
+What other designs have been considered? What is the impact of not doing this?
+
+# Adoption strategy
+
+Please consider:
+
+- If we implement this proposal, how will existing Astro developers adopt it?
+- Is this a breaking change? Can we write a codemod?
+- Can we provide a runtime adapter library for the original API it replaces?
+- How will this affect other projects in the Astro ecosystem?
```
natemoo-re, line 90: - Where in our documentation should this go?
- Does this impact many documentation pages?

FredKSchott:
Sad I missed the RFC call where this was discussed! I'd just call out that all of these templates are mostly copy-pasted from the current RFC template contents. Since this PR is more focused on the overall process, I'd personally consider any serious/deeper changes to the template contents/body as not directly related to this PR and something that could be tackled separately.

If we can keep this PR more focused on the process, then a follow-up PR could focus entirely on improving the actual RFC template itself, and probably benefit from the better focus.

natemoo-re:
Decided to tackle the specific feedback around the exact templates after merging.

PR #459, opened 2023-01-30
`0022-frontmatter-plugins`: add PR etc.

mrienstra:
As per https://github.com/withastro/astro/pull/3411#issuecomment-1407671566

I threw in a docs link, which is non-standard (not part of the template), so maybe I should back that change out? Provided two links, the docs change a lot and aren't versioned, so seemed prudent to link to a specific commit as well.


natemoo-re (approved):
Thanks for updating!

PR #500, opened 2023-02-24
RFC: A plan for a core image story

Princesseuh:
# Summary

This RFC aims to outline a plan to simplify the usage, optimization and resizing of images in Astro without the need of an integration.

# Links

### DIFF
--- a/proposals/0030-core-image-story.md
+++ b/proposals/0030-core-image-story.md
@@ -1,0 +1,402 @@
+- Start Date: 2023-21-02
+- Reference Issues: #5202 (and all the issues about `@astrojs/image`, really)
+- Implementation PR: https://github.com/withastro/astro/pull/6344
+
+# Summary
+
+This RFC aims to simplify the usage, optimization and resizing of images in Astro.
+
+# Nomenclature
+
+- **CLS**: Content Layout Shift. A score determining how much content shifted on your page during loading, you want to be as close to 0 (or "no CLS") as possible.
+
+# Example
+
+> Note: All the `src` paths in the exemples below represent the final paths at build. During development, those paths may be different due to implementations details.
+
+> Note 2: All the results below, unless specified otherwise, are done using the default image services provided with Astro.
+
+## Image Component
+
+### Image located in `src`, in an Astro file
+
+```astro
+---
+import { Image } from 'astro:assets';
+import myImage from "../assets/my_image.png"; // Image is 1600x900
+---
+
+<Image src={myImage} alt="..." />
+```
+
+#### Result
+
+```html
+<!-- Image is optimized, proper attributes are enforced -->
+<img
+  src="/_astro/my_image.hash.webp"
+  width="1600"
+  height="900"
+  decoding="async"
+  loading="lazy"
+  alt="..."
+/>
+```
+
+### Remote images
+
+```astro
+---
+import { Image } from "astro:image";
+// Image is 1920x1080
+---
+
+<Image src="https://example.com/image.png" alt="..." />
+<!-- ERROR! `width` and `height` are required -->
+
+<Image src="https://example.com/image.png" width={1280} alt="..." />
+<!-- ERROR! `height` is required -->
+
+<Image src="https://example.com/image.png" width={1280} height={720} alt="..." />
+```
+
+#### Result
+
+```html
+<!-- Remote images are not optimized or resized. Nonetheless, you still benefit from the enforced attributes and the guarantee of no CLS, albeit manually. -->
+<img
+  src="https://example.com/image.png"
+  decoding="async"
+  loading="lazy"
+  width="1280"
+  height="720"
+  alt="..."
+/>
+```
+
+## Image in Markdown
+
+```markdown
+! [My article cover](./cover.png)
+^ Image is 1280x720
+```
+
+### Result
+
+```html
+<!-- Image is optimized, proper attributes are enforced -->
+<img
+  src="/_astro/cover.hash.webp"
+  width="1280"
+  height="720"
+  decoding="async"
+  loading="lazy"
+  alt="My article cover"
+/>
+```
+
+Remote images are handled as in vanilla Markdown and have no special behaviour.
+
+## JavaScript API for using optimized images
+
+```astro
+---
+import { getImage } from "astro:assets";
+import image from "../assets/image.png"; // Image is 1920x1080
+
+const myImage = await getImage({ src: image, width: 1280, height: 720 });
+---
+
+<img src={myImage.src} {...myImage.attributes} />
+```
+
+### Result
+
+```html
+<img
+  src="/_astro/image.hash.webp"
+  width="1280"
+  height="720"
+  loading="lazy"
+  decoding="async"
+/>
+```
+
+## Defining a new image service
+
+### Astro Config
+
+```ts
+import { defineConfig } from "astro/config";
+
+// https://astro.build/config
+export default defineConfig({
+  image: {
+    service: "your-entrypoint", // 'astro/image/services/squoosh' | astro/image/services/sharp | string
+  },
+});
+```
+
+### Export for local services
+
+```ts
+import type { LocalImageService } from "astro";
+
+const service: LocalImageService = {
+  getURL(options: ImageTransform) {
+    return `/my_super_endpoint_that_transforms_images?w=${options.width}`;
+  },
+  parseURL(url: URL) {
+    return {
+      width: url.searchParams.get("w"),
+    };
+  },
+  transform(options: ImageTransform) {
+    const { buffer } = mySuperLibraryThatEncodesImages(options);
+
+    return {
+      data: buffer,
+      format: options.format,
+    };
+  },
+};
+
+export default service;
+```
+
+### Export for external services
+
+```ts
+import type { ExternalImageService } from "astro";
+
+const service: ExternalImageService = {
+  getURL(options: ImageTransform) {
+    return `https://mysupercdn.com/${options.src}?q=${options.quality}`;
+  },
+};
+
+export default service;
+```
+
+The kind of service exposed is completely invisible to users, the user API is always the same: the `Image` component and `getImage`. Refer to previous examples for usage.
+
+### Additional method
+
+```ts
+getHTMLAttributes(options: ImageTransform) {
+    return {
+        width: options.width,
+        height: options.height
+        // ...
+    }
+}
+```
+
+This method is available on both local and external services.
+
+# Background & Motivation
+
+The currently available `@astrojs/image` is great, however, users ultimately find it very confusing for multiple reasons:
+
+- Error messages are confusing and unhelpful
+- Types historically haven't been super great
+- Missing documentation
+- Due to not being a core integration, it requires manual setup
+- Partly due to the previous points, but also for other reasons, it wasn't always clear for users how the integration behaved.
+- Hard to use in Markdown / MDX.
+
+In this RFC, we'd like to outline a plan / API for a core story for images. The main motivation is to create a story that feels fitting to the Astro core, in that it's easy to understand, intuitive to use, and extendable while also being kind and respectful of the user.
+
+# Goals
+
+- Make an image component that is easy to use and intuitive to understand.
+- Make it easy to use optimized images in Markdown, MDX and future formats.
+- Good, core-worthy, DX with awesome error messages, good types, good documentation
+- No more integration to install!
+
+# Non-Goals of this RFC
+
+- Advanced usage in Markdown (ability to set width, height, quality etc)
+- Using optimized images inside client-side framework components
+- Automatic generation of `srcset` in the integrated services
+- Automatic `loading="eager"` for above the fold images
+- Placeholders generation
+- Background generation
+- Picture component
+- Optimizing & resizing remote images
+- `.svg` support
+
+To be clear, we hope to tackle many of those points in the future, in separate, more precise RFCs. Images are hard.
+
+# Detailed Design
+
+## `astro:assets`
+
+A new virtual module will be exported from a Vite plugin (similar to `astro:content`) exposing the different tools the user can access.
+
+- `Image` (see [[#Image component]])
+- `getImage` (see [[#JavaScript API]])
+- `getImageService` (see [[#Image Services]])
+
+We choose `astro:assets` over `astro:image` on purpose, as to make it intuitive that more things might get exposed from there over time. (maybe a crossover with a way to create generic assets for users?)
+
+## `src/assets`
+
+A new reserved folders for assets. Assets do not have to live in this folder, however putting them in this folder will unlock several benefits. Notably, we consider this folder to be a safe place for git-based CMSes to put any uploaded assets in.
+
+Through an included alias `~/assets` pointing to `src/assets`, it'll be easy for users to refer to any uploaded assets there.
+
+### Content Collections integration
+
+It is fairly common for one of the property of a Markdown piece of content to need to be a reference to an asset (think, cover image for an article, picture for an author etc).
+
+In tandem with the `src/assets` folder, we'd like to introduce a way for users to specify that a specific property needs to refer to a valid asset from the `src/assets` folder:
+
+```ts
+import { asset, defineCollection, z } from "astro:content";
+
+const blogCollection = defineCollection({
+  schema: z.object({
+    title: z.string(),
+    image: asset({ width: 1120, height: 1120 }),
+  }),
+});
+```
+
+Additionally, assets referred this way will be transformed automatically to the same shape as if the image was imported (see section below.), so they're ready to be used optimally. However, optimization / resizing is left to the user to do using the JavaScript API if they desire to.
+
+## New ESM shape for images imports
+
+Currently, importing images in Astro returns a simple `string` with the path of the image. The `@astrojs/image` integration and this RFC enhance this by instead returning the following shape: `{src: string, width: number, height: number, format: string}`.
+
+This shape has multiple benefits:
+
+- It allows users to easily construct `img` tags with no CLS, without needing to opt-in the entire `Image` pipeline if they don't require it:
+
+```astro
+---
+import image from "../my_image.png";
+---
+
+<img src={image.src} width={image.width} height={image.height} />
+```
+
+- This shapes gives all the information needed to build your own image integration. Depending on the transformer used, you'll often need the original metadata (ex: width and height) to resize the image while keeping the aspect ratio. Or, you might need a different encoder depending on the file format. With this output, all of this is given without needing further steps.
+
+## Image component
+
+A component that outputs a simple `<img> `tag with all the required attributes needed to show the image.
+
+### Properties
+
+#### width and height
+
+Those properties are used for resizing the image and ensuring there's no CLS. In most cases, the `width` and `height` would be automatically inferred from the source file and not passing them would just result in the same dimensions.
+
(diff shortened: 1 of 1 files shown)


proposals/0030-core-image-story.md
```diff
@@ -0,0 +1,402 @@
...
+
+// https://astro.build/config
+export default defineConfig({
+  image: {
+    service: "your-entrypoint", // 'astro/image/services/squoosh' | astro/image/services/sharp | string
+  },
+});
+```
+
+### Export for local services
```
FredKSchott, line 140: nit: is "export" the right word here? probably "service API" or something similar

proposals/0030-core-image-story.md
```diff
@@ -0,0 +1,402 @@
...
+  loading="lazy"
+  alt="..."
+/>
+```
+
+### Remote images
+
+```astro
+---
+import { Image } from "astro:image";
```
FredKSchott, line 50: check for outdated import specifiers!

proposals/0030-core-image-story.md
```diff
@@ -0,0 +1,402 @@
...
+Through an included alias `~/assets` pointing to `src/assets`, it'll be easy for users to refer to any uploaded assets there.
+
+### Content Collections integration
+
+It is fairly common for one of the property of a Markdown piece of content to need to be a reference to an asset (think, cover image for an article, picture for an author etc).
+
+In tandem with the `src/assets` folder, we'd like to introduce a way for users to specify that a specific property needs to refer to a valid asset from the `src/assets` folder:
+
+```ts
+import { asset, defineCollection, z } from "astro:content";
```
FredKSchott, line 256: (bikeshed comment, not blocking my approval on this RFC)

Any thoughts on matching `asset` naming? My brain says that `import {asset}` is too generic, but I see that you're trying to match the format of `z.asset()` which makes sense. `import {assetSchema}` is less generic, but now feels so different from `z.asset()`.

What about `import {a} from 'astro:content';`? Then this snippet would look like:

```ts
import { defineCollection, a, z } from "astro:content";
const blogCollection = defineCollection({
  schema: z.object({
    title: z.string(),
    image: a.asset({ width: 1120, height: 1120 }),
  }),
});
```

Or, another idea that's probably less controversial:


```ts
import { defineCollection, schemas, z } from "astro:content";
const blogCollection = defineCollection({
  schema: z.object({
    title: z.string(),
    image: schemas.asset({ width: 1120, height: 1120 }),
  }),
});
```

IDK, maybe too early to be thinking about this since it's the first custom schema we're shipping. Like I said, no need to block on this, and also happy to revisit when we add our second or third custom schema.

proposals/0030-core-image-story.md
```diff
@@ -0,0 +1,402 @@
...
+  - Transform and return the image. It is necessary to return a `format` to ensure that the proper MIME type is served to users.
+
+Ultimately, it is up to the local endpoint (that `getURL` points to) to call both `parseURL` and `transform`. Those two methods are exposed as convention and used by the two base services, but your endpoint is free to achieve this in a different way if needed.
+
+**Optional**
+
+- `getHTMLAttributes(options: ImageTransform): Record<string, any>`
+  - Return all additional attributes needed to render the image in HTML. For instance, you might want to return a specific `class` or `style`, or `width` and `height`.
+
+Overall, it's important to remember that 99% of users won't create services, especially local ones. Adapters (think: Vercel's Image Optimization) and integrations (ex: A Cloudinary integration) will supply and configure services for the users.
```
FredKSchott, line 375: I love this callout. Can you share more about how users will customize this in practice? For example: 

- I'd assume Vercel's image service would be added by the Vercel adapter automatically, but only in production? Is that a correct assumption? And if so, how do I customize local (squoosh vs. sharp) during development.
- What if I want to use something like Cloudinary, is the new way to do that via an integration (`astro add cloudinary`)? If so, how does that conflict with Vercel if I also use the Vercel adapter? How would I pick either one?
- Did you give any thought to having built-in services, chosen via string? Something like `optimize: 'cloudinary'`? Or, integration adds the service, but then you still need to opt-into one specific service via an explicit string in config or via `Image` prop? 
- Speaking of, assuming switching between services on a per-image basis is out of scope? If so, then that would be worth adding to non-goals.

FredKSchott (commented):
Looking good! Left some comments

proposals/0030-core-image-story.md
Princesseuh: > I love this callout. Can you share more about how users will customize this in practice? For example: 
> 
> - I'd assume Vercel's image service would be added by the Vercel adapter automatically, but only in production? Is that a correct assumption? And if so, how do I customize local (squoosh vs. sharp) during development.

Presumably you always want to use the same service in both dev and prod, otherwise you'll have different results? If you do want a different one however, you can do so the same way you can do different settings for dev and prod in the Astro config.

For Vercel's, I'd assume that it would work in dev when you run their CLI perhaps? I'm not sure..

> - What if I want to use something like Cloudinary, is the new way to do that via an integration (`astro add cloudinary`)? If so, how does that conflict with Vercel if I also use the Vercel adapter? How would I pick either one?

It would work the same way as other settings in Astro, which I must admit, I don't know. It's not necessarily intended that integrations set the service for you, but since they can change the Astro config, they can

User is always free to choose through the `image.service` property however

> - Did you give any thought to having built-in services, chosen via string? Something like `optimize: 'cloudinary'`? Or, integration adds the service, but then you still need to opt-into one specific service via an explicit string in config or via `Image` prop? 

I'm not sure I understand the question. This is kinda how it works.

`image.service` can point to any entrypoints, so you can do: 

- Install the `Cloudinary integration or package that has an entrypoint`
- Change `image.service` to `cloudinary-package-entrypoint`

Built in services, the two local ones, are also acceded this way, though they don't need to be installed.

> - Speaking of, assuming switching between services on a per-image basis is out of scope? If so, then that would be worth adding to non-goals.

Yep, out of scope because I'm not sure how to do it in SSR at the moment. Will add, thank you!

proposals/0030-core-image-story.md
```diff
@@ -0,0 +1,427 @@
...
+// Image is 1920x1080
+---
+
+<Image src="https://example.com/image.png" alt="..." />
+<!-- ERROR! `width` and `height` are required -->
+
+<Image src="https://example.com/image.png" width={1280} alt="..." />
+<!-- ERROR! `height` is required -->
+
+<Image src="https://example.com/image.png" width={1280} height={720} alt="..." />
```
ematipico, line 60: I wonder if, among these examples, we should show the `alt` prop as mandatory. While it is not required per se, not having `alt` is a considerable downside for accessibility.

proposals/0030-core-image-story.md
```diff
@@ -0,0 +1,427 @@
...
+  image: {
+    service: "your-entrypoint", // 'astro/image/services/squoosh' | 'astro/image/services/sharp' | string
+  },
+});
+```
+
+### Local services API
+
+```ts
+import type { LocalImageService } from "astro";
```
ematipico, line 143: This is a bikeshed comment:

Have you considered using an abstract class for `LocalImageService`?

With an abstract class, we can declare which methods must be implemented and throw errors if these methods are not implemented.

proposals/0030-core-image-story.md
```diff
@@ -0,0 +1,427 @@
...
+    };
+  },
+  transform(options: ImageTransform) {
+    const { buffer } = mySuperLibraryThatEncodesImages(options);
+
+    return {
+      data: buffer,
+      format: options.format,
+    };
+  },
```
ematipico, line 161: Will the compiler throw an error if some functions are not implemented?

proposals/0030-core-image-story.md
```diff
@@ -0,0 +1,427 @@
...
+#### quality
+
+`quality` inform the quality to use when transforming the image. It can either be a number from `0` to `100`, or a preset from `low`, `medium`, `high`, `max`. Specific implementation here is left to the image service used, as they do not necessarily all abide by the same rules.
+
+Default is left to the services to choose. But for the services exposed by Astro, this property is always optional.
+
+#### alt
+
+`alt` is a required property in all cases when using the component. It can be explicitly set to `""`, for cases where an alt-text is not required.
+
```
ematipico, line 315: Are there cases where `alt` is not needed?? 😱

proposals/0030-core-image-story.md
```diff
@@ -0,0 +1,427 @@
...
+  // Contain additional HTML attributes needed to render the image (ex: `width`, `height`, `loading`)
+  attributes: Record<string, any>;
+}
+```
+
+> Specific types not necessarily accurate to implementation. A more or less open-ended type for `options` and `attributes` is nonetheless required, as different services are allowed to return different things (including non-standard HTML attributes.)
+
+## Image Services
+
+By default, Astro will ship with two services can users can choose to transform their images: Squoosh, and Sharp. Both services more or less offer the same results and mostly differ in where they can run. Squoosh will be the default, because albeit slower than Sharp, it supports more platforms.
```
ematipico, line 343:
```suggestion
By default, Astro will ship with two services that users can choose from to transform their images: Squoosh, and Sharp. Both services more or less offer the same results and mostly differ in where they can run. Squoosh will be the default, because albeit slower than Sharp, it supports more platforms.
```

Are there links we can provide for these two services?

proposals/0030-core-image-story.md
Princesseuh: `@astrojs/image` actually used classes for this! But it was super confusing for users, since classes are not super common in JavaScript, it just looked like a lot of boilerplate. So, despite my own thoughts on the subject, I went with objects.

I think there might also be some considerations regarding the fact that using classes for this would mean that packages needs a hard dependency on Astro, whereas with the type they can leave it as a dev dep and their package can be leaner if it includes other stuff unrelated to Astro

proposals/0030-core-image-story.md
Princesseuh: From MDN:

> Setting this attribute to an empty string (alt="") indicates that this image is not a key part of the content (it's decoration or a tracking pixel), and that non-visual browsers may omit it from [rendering](https://developer.mozilla.org/en-US/docs/Glossary/Rendering_engine).

proposals/0030-core-image-story.md
ematipico: Thank you for providing such background!

proposals/0030-core-image-story.md
Princesseuh: Sorry, I'm not sure if I understand correctly. If you mean the Astro compiler, no. If you mean `tsc`, yes, as the type require those methods to exists on the object.

ematipico: Yeah sorry for the poor choice of words 😆 

Yes, I meant the TS compiler! Thank you!

proposals/0030-core-image-story.md
Princesseuh: I can't say I have any particular thoughts on this. I like the `schemas.asset` one though, but maybe `schema` is a bit too foreign for users not used to Zod?

proposals/0030-core-image-story.md
FredKSchott: > Built in services, the two local ones, are also acceded this way, though they don't need to be installed.

Ah good to know, I missed that above.

> Presumably you always want to use the same service in both dev and prod, otherwise you'll have different results? If you do want a different one however, you can do so the same way you can do different settings for dev and prod in the Astro config.

Well, sort of... the Vercel image optimization is only available in prod. It doesn't work locally since it only works by proxying images out of their own backend/storage. This is how Next.js does it: they use the local image optimizations (squoosh or sharp) in development, and then the Vercel magic `_vercel` endpoint in production.

Princesseuh: Updated with some information on how it's possible to set a different service using something like `process.env.MODE`

GoulvenC:
I had spoken with @Princesseuh  about how cool it would be to store blog images in the `content` folder: one folder per article with its MDX file and its images, rather than cluttering the `public` folder.

Super happy to see this feature could be landed on Astro, especially with all the optimization and resizing tools... I'm testing this immediately!

Thanks for the excellent work 😊

proposals/0030-core-image-story.md
```diff
@@ -0,0 +1,430 @@
...
+
+### External services API
+
+```ts
+import type { ExternalImageService } from "astro";
+
+const service: ExternalImageService = {
+  getURL(options: ImageTransform) {
+    return `https://mysupercdn.com/${options.src}?q=${options.quality}`;
+  },
```
jamesarosen, line 176: This doesn't seem to include the path to the image. Is the (fingerprinted) path part of `options`?

proposals/0030-core-image-story.md
```diff
@@ -0,0 +1,430 @@
...
+
+A component that outputs a simple `<img> `tag with all the required attributes needed to show the image.
+
+### Properties
+
+#### width and height
+
+Those properties are used for resizing the image and ensuring there's no CLS. In most cases, the `width` and `height` would be automatically inferred from the source file and not passing them would just result in the same dimensions.
+
+The only time those properties are required is for remote images.
```
jamesarosen, line 299: Some other viable options, depending on the service: `width` and `aspectRatio` or `height` and `aspectRatio`.

proposals/0030-core-image-story.md
Princesseuh: Sorry, I'm not sure I get what you mean. This is a remote service, there's no hashed files here

proposals/0030-core-image-story.md
Princesseuh: We purposely didn't go the `aspectRatio` route for this compared to `@astrojs/image` due to user confusion. You're definitely free to support it in your service or create a wrapper though

proposals/0030-core-image-story.md
jamesarosen: > there's no hashed files here

I think that's a failure. Specifically, I envision the following course of events:

1. Developer configures an `ExternalImageService` per your example
2. Developer adds a new `src/images/foo.png` and builds site. Published site refers to ``
3. User loads site. Browser makes request to `https://mysupercdn.com/_astro/images/foo.png?q=80` to fetch the optimized version.
4. MySuperCDN makes request to `https://myorigin.com/_astro/images/foo.png` to fetch the original version. Origin responds. MySuperCDN caches the original, performs transforms, caches the transforms
5. Developer updates `src/images/foo.png` and builds site. Published site still refers to ``

There are three possibilities, depending on configuration

1. If the origin uses a long cache time for `foo.png` (e.g. `cache-control: public, max-age=31536000`), then this will result in MySuperCDN having outdated assets. Performance will be good, but the asset will never get cache-busted.
2. If the origin uses a short cache time for `foo.png` (e.g. `cache-control: public, max-age=60`), then this will result in MySuperCDN having to refetch `foo.png` and re-perform transforms frequently, hurting site performance. It also means the asset will be outdated for 60 seconds after a deploy.
3. Some image-transforms CDNs allow content uploads. In this case, the site would have an Astro integration that pushes images to the image service. The service would know that the image has changed and would expire its caches. This won't work with services like Imgix, Fastly, or Cloudflare because their image optimizers all work as caching reverse proxies. It would work with Cloudinary, which has a Digital Asset Management product attached to its image optimizer. But in the Cloudinary case, it would introduce a new failure mode: the image asset would be updated before the deploy finishes. If the deploy fails, you now have old HTML with the new image.

For any of these cases, the solution is simple: pass the fingerprinted path to the image service.

Princesseuh: Hm, I see what you mean. Do you know how other tools help with this? None of the other tools with this kind of API seems to offer a fingerprint, as far as I could tell. 

I'm not sure either how we would offer this, as kinda no fingerprint exists for the original version either.

Princesseuh: Ah, I had misunderstood what you meant initially. If you pass an ESM imported image, you'll get a fingerprinted path, yes.

proposals/0030-core-image-story.md
```diff
@@ -0,0 +1,443 @@
...
+<!-- Remote images are not optimized or resized. Nonetheless, you still benefit from the enforced attributes and the guarantee of no CLS, albeit manually. -->
+<img
+  src="https://example.com/image.png"
+  decoding="async"
+  loading="lazy"
+  width="1280"
+  height="720"
+  alt="..."
+/>
+```
```
jamesarosen, line 76: Do you think pixel-density is within scope of this API?

The output would look something like this:
```html
<img
  srcset="
    /_astro/my_image.dp1hash.webp 1600w,
    /_astro/my_image.dp2hash.webp 3200w
  "
  sizes="(min-width: 3200px) 3200px, 100vw"
  src="/_astro/my_image.dp1hash.webp"
/>
```

I'm not quite sure what the inputs should be. Currently the `@astrojs/image` `Picture` component accepts a `widths: number[]`. But `sizes` and `srcset` are available on `` without a `<picture>`, so I would expect `Image` to support them as well.

At a minimum, I think there should be _some way_ to tell `Image` to generate image files for a list of widths.

Could it be calculated from `srcset`? From `sizes`? Is `widths` required? I'm not sure about these details.

See also [Responsive images with srcset and sizes](https://medium.com/@woutervanderzee/responsive-images-with-srcset-and-sizes-fc434845e948) and [Optimal Images in HTML](https://www.builder.io/blog/fast-images).

Princesseuh: This is considered out of scope for this current proposal (see the non-goals section, it's part of it)

You can achieve this with the APIs shipped in 2.1 technically, but in the future we aim to have official support, just not in this version.

jamesarosen: > This is considered out of scope for this current proposal

As long as the APIs proposed here don't _preclude_ it being added in the future, I'm hugely supportive of shipping one thing at a time 😄👍

wassfila:
@Princesseuh I tried the relative asset in Markdown as described in RFC here https://github.com/withastro/roadmap/blob/feat/core-image-story/proposals/0030-core-image-story.md#image-in-markdown
with Astro 2.1.2 and `experimental: {    assets: true   }` as it was not described, this scope of this question is not to troubleshoot but only to ask if it is still in the race as part of this RFC or not ? As I could not find a reference to it in the doc
https://docs.astro.build/en/guides/images/

Princesseuh:
> @Princesseuh I tried the relative asset in Markdown as described in RFC here https://github.com/withastro/roadmap/blob/feat/core-image-story/proposals/0030-core-image-story.md#image-in-markdown
> with Astro 2.1.2 and `experimental: {    assets: true   }`.
> The scope of this question is not to troubleshoot but only to ask if it is still in the race as part of this RFC or not ? As I could not find a reference to it in the doc
> https://docs.astro.build/en/guides/images/
> 
> I was also wondering if SVG could profit of colocation without necessarily enduring any optimization or is it rather just excluded for the moment ?

Relative images in Markdown are still part of the plan, as you can see on this page: https://docs.astro.build/en/guides/assets/

They're currently supported in the experimental feature, although people have been reporting issues getting it working (that I'm hoping to fix next week!)

SVGs are out of scope for this specific feature at the moment, but could certainly be included in the feature in the future (specifically meaning relative assets, not optimization)

matthewp (approved):

matthewp:
We are now entering the final comment period of this RFC. If there are no objections it will be merged later this week. Thanks!

FredKSchott:
An emphatic +1 on this RFC! Thank you for all of your hard work on this @Princesseuh!

PR #508, opened 2023-02-28
Markdoc support

### LINKED ISSUE #496: Markdoc support in content collections
### Body

- Accepted Date: 2023-02-22
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/478
- Implementation PR: https://github.com/withastro/astro/pull/6209
- Champion: @bholmesdev 

# Summary

This is a proposal to add Markdoc support to content collections.

# Background & Motivation
We've received multiple user requests for Markdoc support since its public launch. In fact, we've [seen early community projects](https://twitter.com/flexdinesh/status/1605685191703687168) bringing Markdoc to robust project themes.

Markdoc is also designed to solve existing limitations of MDX in Astro.
1. **Performance suffers at scale.** Unlike plain Markdown that outputs a string of HTML, MDX outputs a JavaScript module of JSX components. This requires Astro and Babel preprocessing for even the simplest MDX documents. Notably, this required a bump to our maximum memory usage when building / deploying docs.astro.build after migrating to MDX.
2. **Your content is tied to your UI.** MDX can import styles and components directly, which _is_ convenient from a developer standpoint. However, this causes issues when content needs to be reused in multiple contexts. A common example is RSS, where you may want a component-rich version from your blog and a simplified HTML output for your RSS feed.

Markdoc is built to solve (2) by separating content from the components, styles, and assets you choose to render. You can use an Astro component renderer when using on your site, [Markdoc's own `html` renderer](https://markdoc.dev/docs/render#html) for RSS, and even write your own renderer to traverse Markdoc pages yourself. (1) Is something we're excited to test, requiring a thorough performance benchmark.

The content collections API was built generically to support this future, choosing format-agnostic naming like `data` instead of `frontmatter` and `body` instead of `rawContent`. Because of this, introducing new authoring formats is possible without breaking changes.

# Goals
- **Create an `@astrojs/markdoc` integration** that adds `.mdoc` support to content collections.
- **Support Astro components and server-rendered UI components** (React, Vue, Svelte, etc) within Markdoc files. Note this _excludes_ client-rendered UI components (see non-goals).
- **Benchmark Markdoc performance against MDX** at 1000+ documents. This addresses problem (1) from the previous section. Metrics to compare: SSG build time, SSG and SSR build memory usage, SSR response speed.

# Non-goals
- **ESM import and `src/pages/` support** for Markdoc files. [See discussion](https://github.com/withastro/roadmap/discussions/478#discussioncomment-4975072) for context.
- **Allowing the `.md` extension.** This would mean overriding Astro's `.md` renderer, which is tightly coupled to remark and your `markdown` configuration options today. We agree using `.md` for Markdoc is a common use case, and deserves a separate proposal to make Astro's Markdown rendering flexible.
- **A solution for client-rendered UI components.** Unlike MDX, Markdoc doesn't have a concept of directives, and our compiler doesn't have a clear way to dynamically render client-side components ([see challenges](https://gist.github.com/bholmesdev/491fa440efa8bd73d410d66c0c2143c2)). We will recommend users wrap their components in an Astro component to apply the `client:` directive.
- **Full alignment with Markdown and MDX** rendered result. Namely, the computed `headings` property (which can be tackled in future releases) and frontmatter manipulation via remark (since remark is incompatible with Markdoc).

# Example implementation
Markdoc will be introduced as an integration. To standardize our process for adding new collection teams, we may experiment with a (private) integration helper internally. This example shows an `addContentEntryType` hook to setup the `.mdoc` extension, and attach logic for parsing the `data` and `body` properties:

```ts
// @astrojs/markdoc/index.ts
export const markdoc: AstroIntegration = () => ({
		'astro:config:setup'({ addContentEntryType }) {
			addContentEntryType({
				extensions: ['.mdoc'],
				parser: '@astrojs/markdoc/contentEntryParser',
			});
		}
	}
});

// @astrojs/markdoc/contentEntryParser.ts
import parseFrontmatter from 'gray-matter';
export default {
	getEntryInfo({ contents }) {
		const parsed = parseFrontmatter(contents);
		return {
			// The unparsed data object that can be passed to a Zod schema.
			data: parsed.data,
			// The body of the data file. This should be the raw file contents with metadata (i.e. frontmatter block) stripped
			body: parsed.content,
			// (Optional) The untouched frontmatter block with newlines and formatting preserved. Used for computing error line hints.
			rawData: parsed.matter,
		}
	}
}

// astro.config.mjs
import markdoc from '@astrojs/markdoc';

export default {
	integrations: [markdoc()],
}
```

# Example Usage
Say you've authored a collection of blog posts using Markdoc. You can store these entries as a `blog` collection, identically to Markdown or MDX:
```bash
src/content/
	blog/
		# Could also use `.md`
		post-1.mdoc
		post-2.mdoc
		post-3.mdoc
...
```

Then, you can query entry frontmatter with the same `getCollection()` and `getEntryBySlug()` APIs:
```astro
---
import { getCollection, getEntryBySlug } from 'astro:content';

const blog = await getCollection('blog');
const firstEntry = await getEntryBySlug('blog', 'post-1');
---
```

Users should also be free to render Markdoc contents using a `Content` component. This will be exposed from the `render()` result, and feature two props:
- `components?: Record<string, ComponentRenderer>`: A mapping from Markdoc tags or elements to Astro components.
- `config?: import('@markdoc/markdoc').Config`: An (optional) Markdoc config to be used [during the transformation step](https://markdoc.dev/docs/render#transform).

```astro
---
import Title from '../components/Title.astro';
import Marquee from '../components/Marquee.astro';
import { getEntryBySlug } from 'astro:content';

const mdocEntry = await getEntryBySlug('blog', 'test');
const { Content } = await mdocEntry.render();
---

<html lang="en">
	<body>
		<Content
			config={{
				variables: { underlineTitle: true },
			}}
			components={{
				h1: Title,
				marquee: Marquee,
			}}
		/>
	</body>
</html>
```

## Sharing config

This solution is flexible, but we expect users to reuse `config` and `components` across their project. For this, we will recommend creating a utility component to encapsulate that config. Here is one example that can render any `blog` collection entry with an `{% aside /%}` shortcode:

```astro
---
// src/components/BlogContent.astro
import Aside from './Aside.astro';
import type { CollectionEntry } from 'astro:content';

type Props = {
	entry: CollectionEntry<'blog'>;
};

const { entry } = Astro.props;
const { Content } = await entry.render();
---

<Content
	config={{
		tags: {
			aside: {
				render: 'Aside',
				attributes: {
					type: { type: String },
					title: { type: String },
				},
			},
		},
	}}
	components={{ Aside }}
/>
```

Now, you can pass any blog collection entry to render the result with this config:

```astro
---
import { getEntryBySlug } from 'astro:content';
import BlogContent from '../components/BlogContent.astro';

const mdocEntry = await getEntryBySlug('blog', 'test');
---

<h1>{intro.data.title}</h1>
<BlogContent entry={mdocEntry} />
```

[See this example video](https://github.com/withastro/astro/pull/6209) for more.

## Advanced use case: component prop mapping
Component renderers can also include a `props()` function to map Markdoc attributes and AST entries to component props. This is useful when:
- computing props based on the Markdoc AST
- mapping Markdoc's generated attributes to prop names

This example maps Markdoc's generated `data-language` attribute for code blocks to the `lang` prop used by Astro's `Code` component, and stringifies the contents to HTML for use with Shiki:

```astro
---
import { Code } from 'astro/components';
import { Title } from '../components/Title.astro';
import Markdoc from '@markdoc/markdoc';
...
---

...
<Content
	components={{
		h1: Title,
		pre: {
			component: Code,
			props({ attributes, getTreeNode }) {
				return {
					lang: attributes['data-language'],
					code: Markdoc.renderers.html(getTreeNode().children),
				};
			},
		},
	}}
/>
```

bholmesdev:
This RFC introduces Markdoc support to Astro Content Collections. Resolves #496 

## Links

- [Fully rendered proposal](https://github.com/withastro/roadmap/blob/feat/markdoc/proposals/0030-markdoc.md)
- [Stage 2 proposal](https://github.com/withastro/roadmap/issues/496)
- [Stage 1 proposal](https://github.com/withastro/roadmap/discussions/478)

### DIFF
--- a/proposals/0030-markdoc.md
+++ b/proposals/0030-markdoc.md
@@ -105,16 +105,50 @@ It can be useful to illustrate your RFC as a user problem in this section.
 
 Markdoc support will be specific to content collections. This means Markdoc files can be included alongside other content collection entries via the `.mdoc` extension. Entries can also be queried using the same `getCollection()` and `getEntryBySlug()` APIs for Markdown and MDX.
 
+### Markdoc configuration
+
+The Markdoc integration accepts [all Markdoc configuration options](https://markdoc.dev/docs/config).
+
+These options will be applied during [the Markdoc "transform" phase](https://markdoc.dev/docs/render#transform). Markdoc parsing and transforms are run **at build time** (rather than server request time) both for static and SSR Astro projects. This ensures all Markdoc validation is run _before_ your site is deployed, so you avoid shipping invalid Markdoc files and runtime errors to your users.
+
+You can pass these options from the `markdoc()` integration in your `astro.config`. This example declares a `countries` variable and an `includes` function for use across all Markdoc Content Collection entries:
+
+```js
+// astro.config.mjs
+import { defineConfig } from 'astro/config';
+import markdoc from '@astrojs/markdoc';
+
+// https://astro.build/config
+export default defineConfig({
+  integrations: [
+    markdoc({
+      variables: {
+        // Declare a global list of countries
+        countries: ['EN', 'ES', 'JP'],
+      },
+      functions: {
+        // Check if array includes value
+        includes: {
+          transform(parameters) {
+            const [array, value] = Object.values(parameters);
+            return array.includes(value);
+          },
+        },
+      },
+    }),
+  ],
+});
+```
+
 ### Rendering Markdoc content
 
-Users can render contents and configure Markdoc via the `<Content />` component. This will be exposed from the `render()` result, and feature two props:
+Users can render Markdoc entry contents via the `<Content />` component. This feautres a `components` prop mapping configured Markdoc nodes and tags to Astro components or server-rendered UI components.
 
-- `config?: import('@markdoc/markdoc').Config`: An (optional) Markdoc config to be used [during the transformation step](https://markdoc.dev/docs/render#transform). This includes configuration for Markdoc tags and variables.
-- `components?: Record<string, ComponentRenderer>`: An (optional) mapping from Markdoc tags or elements to Astro components.
+Type: `components?: Record<string, ComponentInstance>`
 
-We expect users to share a common Markdoc `config` and `components` setup throughout their application. Because of this, we will recommend creating a utility component to encapsulate that config.
+We expect users to share a common Markdoc `components` setup throughout their application. Because of this, we will recommend creating a utility component to encapsulate that config.
 
-For example, say a user wants to share config specific to their `blog` collection. This `BlogContent` component will add support for an `{% aside %}` tag across blog entries:
+For example, say a user wants to to render their configured `aside` tag with an `Aside` component across their blog collection. This `BlogContent` component will configure a shared `components` prop:
 
 ```astro
 ---
@@ -131,17 +165,6 @@ const { Content } = await entry.render();
 ---
 
 <Content
-	config={{
-		tags: {
-			aside: {
-				render: 'Aside',
-				attributes: {
-					type: { type: String },
-					title: { type: String },
-				},
-			},
-		},
-	}}
 	components={{ Aside }}
 />
 ```
--- a/proposals/0030-markdoc.md
+++ b/proposals/0030-markdoc.md
@@ -74,9 +74,7 @@ It can be useful to illustrate your RFC as a user problem in this section.
 
 - **Create an `@astrojs/markdoc` integration** that adds `.mdoc` support to content collections.
 - **Support Astro components and server-rendered UI components** (React, Vue, Svelte, etc) within Markdoc files. Note this leaves client-rendered UI components out-of-scope (see non-goals).
-- **Benchmark Markdoc performance against Markdown and MDX** at 1k and 10k documents. This addresses problem (1) from the previous section. Metrics to compare:
-  - SSG build time and memory usage when using plain text, Astro components, and UI framework components like React.
-  - SSR render time for the same use cases.
+- **Benchmark Markdoc performance against Markdown and MDX** at 1k and 10k documents. This addresses problem (1) from the previous section.
 
 # Non-goals
 
@@ -250,9 +248,25 @@ declare module 'astro:content' {
 
 # Testing Strategy
 
-How will this feature's implementation be tested? Explain if this can be tested with
-unit tests or integration tests or something else. If relevant, explain the test
-cases that will be added to cover all of the ways this feature might be used.
+Markdoc will have both functionality and performance tests.
+
+## Performance benchmark tests
+
+Write a content collection performance benchmark comparing Markdoc to our existing Markdown and MDX pipelines. The benchmark should compare the following metrics at 1k and 10k documents:
+
+- SSG build time and memory usage when using plain text, Astro components, and UI framework components like React.
+- SSR render time for the same use cases.
+
+Tests should also introduce a flag (ex. `ASTRO_PERFORMANCE_BENCHMARK`) to disable nonessential remark plugins for Markdown and MDX (GFM, Smartypants, Shiki, heading IDs). This ensures a level playing field comparing against Markdoc.
+
+These tests do not need to be tied to our CI pipeline. However, **a detailed summary of testing methods, results, and conclusions** should be drafted to present alongside the `@astrojs/markdoc` integration.
+
+## Integration tests
+
+Add integration tests for using Markdoc content collections in SSG and SSR across development and production. These tests should ensure:
+
+- `getCollection()` and `getEntryBySlug()` return expected properties.
+- rendered `<Content />` is correct for simple text, tags, and components.
 
 # Drawbacks
 
--- a/proposals/0030-markdoc.md
+++ b/proposals/0030-markdoc.md
@@ -76,7 +76,7 @@ const { Content } = await entry.render();
 />
 ```
 
-See [the detailed API section](#api) for information on supported Markdoc configuration, shared config, and advanced component mapping.
+See [the detailed API section](#api) for information on supported Markdoc configuration and shared components config.
 
 # Background & Motivation
 
@@ -158,40 +158,6 @@ const mdocEntry = await getEntryBySlug('blog', 'test');
 
 <h1>{intro.data.title}</h1>
 <BlogContent entry={mdocEntry} />
-```
-
-### Advanced: mapping Markdoc nodes to component props
-
-Component renderers can also include a `props()` function to map Markdoc attributes and AST entries to component props. This is useful when:
-
-- computing props based on the Markdoc AST
-- mapping Markdoc's generated attributes to prop names
-
-This example maps Markdoc's generated `data-language` attribute for code blocks to the `lang` prop used by Astro's `Code` component, and stringifies the contents to HTML for use with Shiki:
-
-```astro
----
-import { Code } from 'astro/components';
-import { Title } from '../components/Title.astro';
-import Markdoc from '@markdoc/markdoc';
-...
----
-
-...
-<Content
-	components={{
-		h1: Title,
-		pre: {
-			component: Code,
-			props({ attributes, getTreeNode }) {
-				return {
-					lang: attributes['data-language'],
-					code: Markdoc.renderers.html(getTreeNode().children),
-				};
-			},
-		},
-	}}
-/>
 ```
 
 ## Internal implementation
--- a/proposals/0030-markdoc.md
+++ b/proposals/0030-markdoc.md
@@ -80,11 +80,16 @@ See [the detailed API section](#api) for information on supported Markdoc config
 
 # Background & Motivation
 
-Include any useful background detail that that explains why this RFC is important.
-What are the problems that this RFC sets out to solve? Why now? Be brief!
+We've received multiple user requests for Markdoc support since its public launch. In fact, we've seen an early community project, [Blogster](https://github.com/flexdinesh/blogster), that brings Markdoc to robust project themes. This includes an [Astro Markdoc renderer](https://github.com/flexdinesh/blogster/tree/main/packages/astro-markdoc-renderer) that may inspire a first party integration.
 
-It can be useful to illustrate your RFC as a user problem in this section.
-(ex: "Users have reported that it is difficult to do X in Astro today.")
+Markdoc is also designed to solve existing limitations of MDX in Astro:
+
+1. **Performance suffers at scale.** Unlike plain Markdown that outputs a string of HTML, MDX outputs a JavaScript module of JSX components. This requires Astro and Babel preprocessing for even the simplest MDX documents. Notably, this required a bump to our maximum memory usage when building / deploying docs.astro.build after migrating to MDX.
+2. **Your content is tied to your UI.** MDX can import styles and components directly, which _is_ convenient from a developer standpoint. However, this causes issues when content needs to be reused in multiple contexts. A common example is RSS, where you may want a component-rich version from your blog and a simplified HTML output for your RSS feed.
+
+Markdoc is built to solve (2) by separating content from the components, styles, and assets you choose to render. You can use an Astro component renderer when using on your site, [Markdoc's own `html` renderer](https://markdoc.dev/docs/render#html) for RSS, and even write your own renderer to traverse Markdoc pages yourself. (1) Is something we're excited to test, requiring a thorough performance benchmark.
+
+The content collections API was built generically to support this future, choosing format-agnostic naming like `data` instead of `frontmatter` and `body` instead of `rawContent`. Because of this, introducing new authoring formats is possible without breaking changes.
 
 # Goals
 
--- a/proposals/0030-markdoc.md
+++ b/proposals/0030-markdoc.md
@@ -270,19 +270,13 @@ Add integration tests for using Markdoc content collections in SSG and SSR acros
 
 # Drawbacks
 
-Why should we _not_ do this? Please consider:
-
-- Implementation cost, both in term of code size and complexity.
-- Whether the proposed feature can be implemented in user space.
-- Impact on teaching people Astro.
-- Integration of this feature with other existing and planned features
-- Cost of migrating existing Astro applications (_is it a breaking change?_)
-
-There are tradeoffs to choosing any path. Attempt to identify them here.
+- Introducing a new content format adds overhead to Astro core and (potentially) learning curve for new users. This should be addressed with clear messaging that Markdoc is experimental, and MDX is still our recommendation for component-rich content.
+- Adding performance benchmarks for Markdoc may add complexity to our CI pipeline. We should ensure automated test times remain reasonable across PRs when adding new test suites.
 
 # Alternatives
 
-What other designs have been considered? What is the impact of not doing this?
+- **Double down on MDX support**, investing more engineering resources into the format's performance, language tools, and featureset. We have already invested time into improving Rollup build times to accommodate higher memory usage. Still, there are certain aspects of MDX (coupling content to UI) that are unavoidable without proposing fundamental changes.
+- **Introduce a new Markdown format** specific to Astro. This has echoes in our former Astro-flavored Markdown format, and has been mentioned in [recent roadmap discussion posts](https://github.com/withastro/roadmap/discussions/503#discussioncomment-5125773). This has a higher maintenance cost attached that was deemed unsustainable during the Astro 1.0 Beta period. Even so, it would be the most optimal path for Astro to own the language tooling and performance story similar to the `.astro` format today.
 
 # Adoption strategy
 
--- a/proposals/0030-markdoc.md
+++ b/proposals/0030-markdoc.md
@@ -76,7 +76,7 @@ const { Content } = await entry.render();
 />
 ```
 
-See [the detailed API section](#api) for information on supported Markdoc configuration and shared components config.
+See [the detailed API section](#api) for information on supported Markdoc configuration, shared components config, and manual rendering.
 
 # Background & Motivation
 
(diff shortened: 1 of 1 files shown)


proposals/0030-markdoc.md
```diff
@@ -0,0 +1,295 @@
...
+
+# Non-goals
+
+- Non-goal: **ESM import and `src/pages/` support** for Markdoc files. [See discussion](https://github.com/withastro/roadmap/discussions/478#discussioncomment-4975072) for context.
+- Out-of-scope: **Allowing the `.md` extension.** This would mean overriding Astro's `.md` renderer, which is tightly coupled to remark and your `markdown` configuration options today. We agree using `.md` for Markdoc is a common use case, and deserves a separate proposal to make Astro's Markdown rendering flexible.
+- Future: **A solution for client-rendered UI components.** Unlike MDX, Markdoc doesn't have a concept of directives, and our compiler doesn't have a clear way to dynamically render client-side components ([see challenges](https://gist.github.com/bholmesdev/491fa440efa8bd73d410d66c0c2143c2)). We will recommend users wrap their components in an Astro component to apply the `client:` directive.
+- Future: **Full alignment with Markdown and MDX** rendered result. Namely, the computed `headings` property (which can be tackled in future releases) and frontmatter manipulation via remark (since remark is incompatible with Markdoc).
```
FredKSchott, line 86: nit: I think we're undecided if remark support for Markdoc is something we actually want (I'm at least somewhere between "unsure" and "against"). Wouldn't want to imply that this is decided as wanted for the future.

proposals/0030-markdoc.md
```diff
@@ -0,0 +1,295 @@
...
+Include any useful background detail that that explains why this RFC is important.
+What are the problems that this RFC sets out to solve? Why now? Be brief!
+
+It can be useful to illustrate your RFC as a user problem in this section.
+(ex: "Users have reported that it is difficult to do X in Astro today.")
+
+# Goals
+
+- **Create an `@astrojs/markdoc` integration** that adds `.mdoc` support to content collections.
+- **Support Astro components and server-rendered UI components** (React, Vue, Svelte, etc) within Markdoc files. Note this leaves client-rendered UI components out-of-scope (see non-goals).
```
FredKSchott, line 78: I really like that client-side UI component logic is out of scope for now. It feels very Astro to default to HTML-only but then still gives you all of the power of Astro islands inside of Astro's syntax.

Was there any thought to only supporting Astro components and not server-rendered UI components? If it's more complexity to disable non-Astro components than to allow them then I guess I don't feel strongly, but there is a nice, limited-scope MVP quality to Astro-only to start.

proposals/0030-markdoc.md
```diff
@@ -0,0 +1,295 @@
...
+const { entry } = Astro.props;
+const { Content } = await entry.render();
+---
+
+<Content
+	config={{
+		tags: {
+			aside: {
+				render: 'Aside',
+				attributes: {
```
FredKSchott, line 124: just confirming: if attributes don't match in the `mdoc` file, does this error?  That's what I would expect, but wanted to confirm.

proposals/0030-markdoc.md
```diff
@@ -0,0 +1,295 @@
...
+
+const { entry } = Astro.props;
+const { Content } = await entry.render();
+---
+
+<Content
+	config={{
+		tags: {
+			aside: {
+				render: 'Aside',
```
FredKSchott, line 123: this use of a string is the only issue I have with in the current API. Keeping this string value matching the JS variable name is always an annoying and not-super-obvious API. 

Is there any way to pass `Aside` as a reference here, and then skip the `components={}` step entirely? Something like:

```
<Content
	config={{
		tags: {
			aside: {
				component: Aside,
				attributes: {
				    type: { type: String },
					title: { type: String },
				},
			},
		},
	}}
/>
```

I can't overstate how poor a DX it is to maintain a string value referencing a variable name in the same file like this. TS type checking may help here but removing the problem entirely would be better.

proposals/0030-markdoc.md
```diff
@@ -0,0 +1,295 @@
...
+
+## API
+
+Markdoc support will be specific to content collections. This means Markdoc files can be included alongside other content collection entries via the `.mdoc` extension. Entries can also be queried using the same `getCollection()` and `getEntryBySlug()` APIs for Markdown and MDX.
+
+### Rendering Markdoc content
+
+Users can render contents and configure Markdoc via the `<Content />` component. This will be exposed from the `render()` result, and feature two props:
+
+- `config?: import('@markdoc/markdoc').Config`: An (optional) Markdoc config to be used [during the transformation step](https://markdoc.dev/docs/render#transform). This includes configuration for Markdoc tags and variables.
```
FredKSchott, line 98: meta-point: this isn't a useful type annotation since we don't actually have that type anywhere accessible inside the RFC. To fix this, I'd add a code example similar to the one below that I ended up commenting on, but just to illustrate the basic config first.

proposals/0030-markdoc.md
```diff
@@ -0,0 +1,295 @@
...
+```
+
+### Advanced: mapping Markdoc nodes to component props
+
+Component renderers can also include a `props()` function to map Markdoc attributes and AST entries to component props. This is useful when:
+
+- computing props based on the Markdoc AST
+- mapping Markdoc's generated attributes to prop names
+
+This example maps Markdoc's generated `data-language` attribute for code blocks to the `lang` prop used by Astro's `Code` component, and stringifies the contents to HTML for use with Shiki:
```
FredKSchott, line 156: Do we have any other use-cases? If this is the only one, then I'd say this is out of scope and the `Code` component could just support this, to be revisited when we had a real use-case.

If we already have other/multiple use-cases, then okay to ignore this!

FredKSchott (commented):
Looking good! Left some comments

proposals/0030-markdoc.md
delucis: Probably related to the conversation Ben & I had here: https://github.com/withastro/roadmap/issues/496#issuecomment-1446402335

proposals/0030-markdoc.md
bholmesdev: Hm, this type is based on the `@markdoc/markdoc` package and not the Astro internals, so it seemed sufficient to me. Would a hyperlink to the type def in the Markdoc repository be helpful instead of an annotation?

proposals/0030-markdoc.md
bholmesdev: Ah, I noted this since Remark is fundamentally not compatible with Markdoc right now. This would require changes to Markdoc core to implement support. I could change the language to imply we _could_ explore this with the Markdoc team.

proposals/0030-markdoc.md
bholmesdev: Yes, it was more engineering work to disable server rendered non-Astro components. We can check for the `AstroComponentFactory` flag to disable this though! I understand this could mean a clearer user story at launch 👍

proposals/0030-markdoc.md
bholmesdev: Yes, if you're missing a `tags` definition for a tag used in a Markdoc file, it will error.

proposals/0030-markdoc.md
bholmesdev: I understand the DX concern here! However, I'm opposed to changing the shape of `config` in any way before launch. This follows Markdoc's transform config 1-1, and it feels wrong to modify certain aspects of that config for our purposes. That would feel like modifying the shape of the Vite config when using Astro. 

I also understand your concern over using plain strings, though this was an intentional choice by the Markdoc team to allow rendering a plain HTML element instead of a component (i.e. `'aside'`). I agree we could explore modifications to Markdoc's config in the future, though I'd much prefer to follow the precedent set by Markdoc's React renderer:
- `config` used for the transformation step, with no reference to rendered components
- `components` used for the render step, now with reference to rendered components

proposals/0030-markdoc.md
bholmesdev: This was the initial use case that prompted the need. I'll come back to you if there are more clear use cases. Otherwise, we can remove!

proposals/0030-markdoc.md
bholmesdev: Update: we may need to move the `config` prop off of the `Content` component and towards some global export as well to address @delucis' "run transforms at build time" suggestion. This could clarify the separation that's fairly opaque in this example.

proposals/0030-markdoc.md
bholmesdev: Update: realized Astro's `slots.render()` can fill this use case well. Removing from RFC and the final API!

proposals/0030-markdoc.md
bholmesdev: Discussed offline! This API has been redesigned to colocate config in your `astro.config.mjs`. This should make the string -> component map tradeoff more usable for an MVP. We plan to improve from here by experimenting with the [NextJS integration's approach](https://markdoc.dev/docs/nextjs) or a reserved folder to autowire this config (i.e. `embeds/`)

bholmesdev:
Alright, excited to share our findings from the performance benchmark!

# Performance Benchmark

As part of the `@astrojs/markdoc` integration, we built a performance benchmark to compare Markdoc build performance at scale against our existing content formats, Markdown and MDX.

## Method
This suite generates content collection entries across 3 separate test fixtures (one for each file type). This will be repeated across 3 different test templates at both 1000 and 3000 generated posts:
- [**Simple case (md, mdx, mdoc):**](https://github.com/withastro/astro/pull/6209/files#diff-fd3b60a61bc850912e5b6a3d57b98f06c5702d72def22950c5eaff205f7c3786) includes 1 frontmatter entry, a heading, and plain Lorem Ipsum text content. This should compare text rendering speed.
- [**Astro components case (mdx, mdoc):**](https://github.com/withastro/astro/pull/6209/files#diff-0829862bf35c543a3149abd04d81283f4d149a237543c8ecc691478245be81f9) this includes the same frontmatter as the simple case, plus 2 examples for rendering Astro components. The first example maps HTML elements (the `h1` tag) to an Astro component. This use the `components` Content prop available to both Markdoc and MDX. The second example imports and renders an Astro component in the file, using ESM imports for MDX and tags + necessary transform config for Markdoc.
- [**React components case (mdx, mdoc):**](https://github.com/withastro/astro/pull/6209/files#diff-3f528672934e0cf5cbab7958b6cfdf066ebb237d6dfc1828ede0622bbf953b5b) this also includes the same frontmatter as the simple case, plus both a server-rendered and hydrated React component ([a simple `LikeButton`](https://github.com/withastro/astro/pull/6209/files#diff-de4227facedef9b49b4709d985b534d6a62f27d5b34eeffbcc1ad1667c8465b2)). This uses ESM imports for MDX and tags + necessary transform config for Markdoc.

Each fixture is built using the `astro build` command. We have also included an environment flag to disable nonessential remark plugins for Markdown and MDX, including GFM, Smartypants, heading ID generation, and Shiki syntax highlighting. Since our Markdoc integration does not (yet) include these features, this should ensure an even playing field for base render performance.

## Results
Below are the overall build times for each file type across each test template, with times averaged across 3 separate test runs:

### 1000 generated posts

|      | Simple | With Astro components | With React components |
|------|--------|-----------------------|-----------------------|
| md   | 17.52s | N/A                   | N/A                   |
| mdx  | 18.17s | 18.27s                | 18.73s                |
| mdoc | 11.03s | 7.89s                 | 6.99s                 |

### 3000 generated posts

|      | Simple | With Astro components | With React components |
|------|--------|-----------------------|-----------------------|
| md   | 55.31s | N/A                   | N/A                   |
| mdx  | 48.47s | 133.24s ⚠️ CPU ceiling | 133.96s ⚠️ CPU ceiling |
| mdoc | 30.66s | 23.27s                | 24.56s                |

_**⚠️ CPU ceiling** - CPU usage hit 100%, throttling builds_

## Discussion

There are several promising findings from this test. First, Markdoc was noticeably faster than Markdown and MDX across all test cases. This could partially be due to the speed of Markdoc's bespoke transform pipeline compared to remark and rehype. There is also a difference in rendering strategies. While MDX compiles to JSX directly, requiring a pass through Astro's JSX runtime to render as JavaScript, Markdoc compiles to a serializable JSON tree we can render directly at template build or SSR runtime. This process proves to be less CPU intensive, with MDX hitting a CPU ceiling at 3000 posts on our test machine.

_Aside: This also means the `@markdoc/markdoc` package will be shipped to your SSR runtime, which can prohibitively large (40kb compressed) for edge compute._

Second, we noticed that Markdoc is slowest during the **simple** test case, while Markdown and MDX see a performance hit when rendering components. This means,  unlike Markdown and MDX, Markdoc is slowed less by complexity of elements and more by the _volume_ of HTML elements rendered (Lorem Ipsum paragraphs rendered in the simple case). It's worth analyzing whether our own render stage or Markdoc's transform stage is to blame here.

## Future exploration
There are several areas we'd like to explore to further compare Markdoc against our existing MDX + remark pipeline:
- **How do Markdoc transforms compare to remark plugins?** This test operates without any Markdoc transformation overrides. What will be the % change in build times adding, say, a Shiki configuration to Markdoc? How does the % change compare to our MDX rehype plugin today?
- **How does Markdoc compare on memory usage?** This test only compares build times. We're curious if Markdoc can address the spike in memory usage MDX experiences in our prebuild step.
- **How does a large library of tags compare to a large library of components?** This test uses just one or two components across test files. Once component libraries scale into the dozens or even hundreds, how will Markdoc compare?

proposals/0030-markdoc.md
```diff
@@ -0,0 +1,327 @@
...
+# Goals
+
+- **Create an `@astrojs/markdoc` integration** that adds `.mdoc` support to content collections.
+- **Support Astro components and server-rendered UI components** (React, Vue, Svelte, etc) within Markdoc files. Note this leaves client-rendered UI components out-of-scope (see non-goals).
+- **Benchmark Markdoc performance against Markdown and MDX** at 1k and 10k documents. This addresses problem (1) from the previous section.
+
+# Non-goals
+
+- Non-goal: **ESM import and `src/pages/` support** for Markdoc files. [See discussion](https://github.com/withastro/roadmap/discussions/478#discussioncomment-4975072) for context.
+- Out-of-scope: **Allowing the `.md` extension.** This would mean overriding Astro's `.md` renderer, which is tightly coupled to remark and your `markdown` configuration options today. We agree using `.md` for Markdoc is a common use case, and deserves a separate proposal to make Astro's Markdown rendering flexible.
```
flexdinesh, line 103: A little history into how I tried to parse markdoc in `*.md` extensions in Blogster and why it introduced a whole lot of unintended complexity and why I ended up not doing in.

This was one of the dilemmas I had when building Blogster. The original motivation was that I didn't want users to choose b/w markdoc and markdown. I wanted markdoc to be an extension of markdown in a way that when users don't write the special markdoc component syntax (`{% custom-component attr="val" %}`) in their `*.md` files, the content would be parsed with Astro's markdown parser during build and if markdoc specific syntax is detected, the markdoc parser would be used.

Few reasons why I thought this would be good for users:
- Aligns well with Astro's philosophy of default to simple and extend progressively
- Users don't have to decide b/w markdown vs markdoc. All files are `*.md` files and markdoc is just an extension of markdown
- `*.md` is a standard extension and UI tools like GitHub web UI for PRs/etc has a preview mode for markdown files. So markdown compatible syntax in markdoc files could still benefit from the syntax highlighting and preview options in such tools

However when I tried writing a custom build plugin to solve this I realised how much complex it was to smartly switch parsers when both markdoc and markdown content have the same entension `*.md`. I dropped the idea and ended up building the parser config as a utility here: [Blogster utility to read and parse *.md files with markdoc parser](https://github.com/flexdinesh/blogster/blob/386a10a70d3e2f7dd92bf887fda7a62896cfd96f/templates/bubblegum/src/lib/markdoc/read.ts#L99) instead of building it as an Astro integration.

bholmesdev: Well said! I agree that `*.md` support would be a nicety that aligns with Markdoc's own documentation. Still, [I'll echo Erika's perspective on this](https://github.com/withastro/roadmap/discussions/503#discussioncomment-5123556): Pointing multiple formats to a single extension makes language tooling difficult. We've noticed that Markdoc is _not_ a strict superset of Markdown (it namely does not allow plain HTML in Markdown), so this can lead to a muddled experience as different extensions compete. I also think that flipping a file extension from `.md` to `.mdoc` is still a simple story for incremental adoption.

I do like Blogster's approach of making `.md` support a separate plug-and-play experience though. Would you be open to making a discussion post on this if it's a feature we want?

louiss0:
Can we please get loops eventually? The mardoc people don't seem to want to give people loops at all. I'm not sure whether or not people will just use markdoc. But it's supposed to be the recommend way of writing content. People will use mdx just for the power to do loops. The markdoc people will not support loops [Proof Here](https://github.com/markdoc/markdoc/discussions/308) but you guys can.

louiss0:
I must say that a component that iterates through an array of things is possible. The problem is that the variables there is no way to access variables in the loop at all.

proposals/0030-markdoc.md
```diff
@@ -0,0 +1,327 @@
...
+          attributes: {
+            // Component props as attribute definitions
+            // See Markdoc's documentation on defining attributes
+            // https://markdoc.dev/docs/attributes#defining-attributes
+            type: { type: String },
+          }
+				},
+			},
+		}),
+	],
```
fredericoo, line 57: could these be defined on the content level too? e.g.: `projects` have a different set of tags than `blog`, etc. I see we’re getting the `<Content/>` component from `entry.render()` which makes this viable. Perhaps later rather than sooner, as I believe this would come with its own type-safety challenges and require specialised tooling

bholmesdev: The idea's definitely come up! This would require some checks to see which config to apply, but it wouldn't be too difficult. Though we're not sure dividing configs per-collection is the cleanest way to divide these up. Did you hit a particular use case in your own projects you can share?

fredericoo: Well that's the example really, projects haven't got `aside`, but have `carousel` etc, whereas blog doesn't. Not that it can't be made to work like that, but it'd be like comparing typescript unions with optionals: we _can_ make a single type full of optionals but if it's either or it often times makes sense to use unions.

bholmesdev: Fair enough! The thinking is you'll apply all of these tags globally, and only reach for the tags you want by-collection. Astro will only ship resources for a tag for content that uses it.  I see the optional vs. union parallel, but there's never a case where a tag is _required_ in a document, so I don't think it's quite the same?

Only trouble I could see is whether a given tag like `aside` _should act differently_ depending on the collection. So far, I haven't seen that become an issue. Again, we could make this happen, just want to find a case where globals are a blocker!

matthewp:
Closing as Markdoc support is now stable in Astro. We are going to create a new RFC when/if Markdoc comes into core, but it can continue to evolve as an integration for now.

Issue #540, opened 2023-03-24
CSS Ordering


bluwy:
### Body

- Accepted Date: 2023-03-24
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/510 and https://github.com/withastro/roadmap/discussions/371
- Implementation PR: 

# Summary

Allow styles in Astro files to override global styles.

# Background & Motivation

A common pattern in an application is to have some global styles in a global style file like so:

__global.css__

```css
input {
  border-width: 1px;
}
```

And then to use them in an Astro component.

```astro
---
import '../global.css';
---
<style>
  input {
    border-width: 2px;
  }
</style>

<input type="text">
```

The intent is to override the styles on this page (or component). But due to chunking of shared dependencies, we cannot *guarantee* that CSS ordering will be correct.

And due to the use of `:where()` for scoping, scoped styles can't _win_ in specificity to apply the overriden styles. (as `:where` preserves specifity, unlike Vue or Svelte that adds attributes/classes to increase specificity) (See [original RFC](https://github.com/withastro/roadmap/blob/main/proposals/0012-scoped-css-with-preserved-specificity.md) for details)

We need a solution for this.

# Goals

- A way to have scoped styles take priority over global styles.
- A consistent and predictable implementation of CSS ordering.
- Better browser support (or an easy way to support old browsers).

# Non-Goals 

- Auto-polyfilling CSS features.

# Example

Below are possible solutions:

1. Go back to using scoped-attributes/classes (like Vue and Svelte), this would increase specificity by 1.
2. Use `@layer`: https://developer.mozilla.org/en-US/docs/Web/CSS/@layer. However, this doesn't have better browser support than `:where`.
3. Manually mark certain CSS files to be global (via `astro.config`?), and order them first to have lower priority.
4. [FUTURE] Use `@scope`: https://www.w3.org/TR/css-scoping-1/

matthewp:
@bluwy can up with the idea for this configuration value:

```ts
scopedStyleStrategy: 'where' | 'class' | 'layer'
```

Note, we don't have `'layer'` now and don't intend to add it, just a demonstration of a possible future approach.

natemoo-re:
[...]

PR #543, opened 2023-03-28
Scoping RFC

matthewp:
This RFC introduces a new `scopedStyleStrategy` option to re-enable class based scoping, preventing bugs associated with CSS ordering that occur with the current `where` based strategy.

## Links

- [Fully rendered proposal](https://github.com/withastro/roadmap/blob/51d7fd2fa42f7e23b11f904e03b63b5efc7b2770/proposals/style-scoping.md)
- [Stage 2 proposal](https://github.com/withastro/roadmap/issues/540)
- [Stage 1 proposal](https://github.com/withastro/roadmap/discussions/510)

### DIFF
--- a/proposals/style-scoping.md
+++ b/proposals/style-scoping.md
@@ -1,0 +1,73 @@
+- Start Date: 2023-03-28
+- Reference Issues: https://github.com/withastro/roadmap/issues/540
+- Implementation PR: <!-- leave empty -->
+
+# Summary
+
+Make the scoping strategy be configurable via a new config option, and add a strategy that uses class names in the selector, giving all Astro styles a +1 in specificity.
+
+# Example
+
+A new configuration option available:
+
+```js
+import { defineConfig } from 'astro/config';
+
+export default defineConfig({
+  scopedStyleStrategy: 'class'
+});
+```
+
+The valid options for this config are:
+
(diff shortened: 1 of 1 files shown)


tony-sull:
@matthewp could have sworn I sent a reply to this from mobile, looks like it never published!  Sorry for the delay!

I could see us handling that by defining what layers Astro supports out of the box.  Global styles could be wrapped in something like `@layer astro.global { }` or similar to make sure it's always layered correctly compared to other layers like `astro.imports` or `astro.components`

I'm always torn on that kind of solution since it's really dependent on documenting a naming convention for the layers, I could go either way on whether that's worth the tradeoff.

matthewp:
This RFC is now under consensus vote. If anyone sees a reason as to why this RFC should *not* be merged, please chime in now. Thank you!

ThatOneCalculator:
I've reviewed the RFC, and it seems good. However, the "Switch the default in 3.0" part may be a bit polarizing -- this may work better as an opt-in feature?

tony-sull:
Big 👍 from me!

My comments above fall more in the "what could we do here in a future RFC" bucket - I really like this proposal for a user configurable toggle to switch between `:where()` and class scoping, clean solution to the problem at hand!

ThatOneCalculator:
Fair enough.

Issue #534, opened 2023-03-23
CDN support


bluwy:
### Body

- Accepted Date: 2023-03-23
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/456
- Implementation PR: 

# Summary

Provide an option to specify a CDN URL for static assets to serve from in production.

# Background & Motivation

Large traffic sites often have CDN servers that are optimized to serve assets only. For example, a site served from `https://astro.build` would reference all it's assets from `https://cdn.astro.build`.

A CDN URL would also allow assets from multiple sites to be deployed to the same CDN server to share the performance, where they can be separated by URL subpaths.

There are also prior art from other frameworks that users want in Astro too:

- Nextjs: https://nextjs.org/docs/api-reference/next.config.js/cdn-support-with-asset-prefix
- Nuxt: https://nuxt.com/docs/api/configuration/nuxt-config#cdnurl
- SvelteKit: https://kit.svelte.dev/docs/configuration#paths

# Goals

- Provide an option to prepend generated asset links with the CDN URL.
- Provide an env variable to access the CDN URL (`import.meta.env.*`) to prepend links manually.
- Works with static and server output.
- Works with existing Astro features, like content collections, `astro:assets` images, and `@astrojs/image`.

# Non-Goals 

- Auto-prefixing CDN URLs to user-created `<link href>` or `<img src>` etc.
- Allow changing the CDN URL in runtime.
- Handling CDN URLs for files in the `public` directory.
  - Users have to preprend the URL themselves if needed.
- The Astro CDN service.

# Example

Astro config:

```js
// astro.config.js
export default defineConfig({
  build: {
    cdnUrl: 'https://cdn.example.com'
  }
})
```

Generated HTML example:
```html
<!DOCTYPE html>
<html>
  <head>
    <link rel="stylesheet" href="https://cdn.example.com/index.123456.css" />
  </head>
  <body>
    <!-- if astro images are used -->
    <img src="https://cdn.example.com/penguin.png" />
  </body>
 </html>
```

Details and behaviours:

- With the default `build.assets` option of `_astro` directory, the user will copy all the files from the `dist/_astro` directory to the `https://cdn.example.com` server.
- The CDN URL is not affected by `base` option, because `base` is the serving base of the user-facing site URL, e.g. `https://example.com/foo/`. (similar to prior art)
- The CDN URL can accept any pathname, for example `https://cdn.example.com/project-a`, and Astro would include it in the built URL.

# More prior art research

[...]

PR #545, opened 2023-03-30
CDN support

bluwy:
This RFC adds CDN support through the `build.assetsPrefix` option. Allowing sites to fetch all assets from an external CDN.

## Links

- [Fully rendered proposal](https://github.com/withastro/roadmap/blob/7dd302e0134813f1de67637792883680de40d60d/proposals/cdn-support.md)
- [Stage 2 proposal](https://github.com/withastro/roadmap/issues/534)
- [Stage 1 proposal](https://github.com/withastro/roadmap/discussions/456)

### DIFF
--- a/proposals/cdn-support.md
+++ b/proposals/cdn-support.md
@@ -1,0 +1,114 @@
+- Start Date: 2023-03-30
+- Reference Issues: https://github.com/withastro/roadmap/issues/534
+- Implementation PR: https://github.com/withastro/astro/pull/6714
+
+# Summary
+
+Provide a `build.assetsPrefix` option to specify a CDN URL for static assets to serve from in production.
+
+# Example
+
+```js
+// astro.config.mjs
+import { defineConfig } from 'astro'
+
+export default defineConfig({
+  build: {
+    assetsPrefix: 'http://cdn.example.com'
+  }
+})
+```
+
+# Background & Motivation
+
+Large traffic sites often have CDN servers that are optimized to serve assets only. For example, a site served from `https://astro.build` would reference all it's assets from `https://cdn.astro.build`.
+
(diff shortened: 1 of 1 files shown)


proposals/cdn-support.md
```diff
@@ -0,0 +1,114 @@
...
+
+# Adoption strategy
+
+This should be transparent to all users. Only when `build.assetsPrefix` is used where the feature kicks in.
+
+For third-party Astro libraries that contribute assets externally, they would need to consider `build.assetsPrefix` too. But I'm not aware of any libraries that does this (that's outside of Astro's asset pipeline).
+
+# Unresolved Questions
+
+1. Is the `build.assetsPrefix` name fine? I picked it to complement the existing `build.assets` option. `build.cdnUrl` feels "sticking out".
```
FredKSchott, line 109: That feels fine to me. A bit odd that it's `assetsPrefix` instead of Next's `assetPrefix` but I get that matching `build.assets` is probably more important.

FredKSchott (approved):
LGTM. Read the RFC and everything you outlined makes sense to me, awesome stuff!

GrantBirki:
This looks awesome! I think it would be a massive win to serve assets from a CDN for larger projects using Astro.

The only thing I would add is perhaps some sort of section that outlines a checklist of all major CDNs that will be tested to ensure they work with this newly purposed feature.

- Fastly
- Cloudflare
- Akamai
- Cloudfront
- Using an S3 bucket perhaps?
- Etc

bluwy:
Tested with Cloudflare and it seems to work: https://github.com/bluwy/astro-cdn-test and https://bluwy.github.io/astro-cdn-test/. The core workings is the same so I'd expect it to work in other platforms too as long as the links are correct.

Issue #531, opened 2023-03-22
Astro middleware


matthewp:
### Body

- Accepted Date: 2023-03-22
- Reference Issues/Discussions:
  - https://github.com/withastro/roadmap/discussions/469
  - https://github.com/withastro/roadmap/discussions/174
  - Legacy `Astro.context` RFC https://github.com/withastro/roadmap/pull/230
- Implementation PR: 

# Summary

Introduce a middleware to Astro, where you can define code that runs on every request. This API should work regardless of the rendering mode (SSG or SSR) or adapter used. Also introduces a simple way to share request-specific data between proposed middleware, API routes, and `.astro` routes.

# Background & Motivation

Middleware has been one of the most heavily requested feature in Astro. It's useful for handling common tasks like auth guards and setting cache headers. For me, it would make handling authentication much easier.

This proposal is heavily inspired by [SvelteKit's `handle` hooks](https://kit.svelte.dev/docs/hooks#server-hooks-handle);

# Goals

- Provide a way intercept requests and responses, allowing users to set cookies and headers
- Works both in SSR and SSG mode
- Allow users to use community-created middlewares (libraries)
  - Make available via integrations API.
- Provide an API for request-specific data
- Non-Node runtimes specific APIs. ie. Cloudflare Durable Objects.
  - Add middleware from adapter.

# Non-Goals 

- Route specific middleware, middlewares that are run **only on specific** routes

# Example

[...]

PR #555, opened 2023-04-14
Astro middleware

ematipico:
This RFC introduce a new middleware API in Astro

## Links

- [Fully rendered proposal](https://github.com/withastro/roadmap/blob/rfc/stage-3-middleware/proposals/0032-middleware-api.md)
- [Stage 2 proposal](https://github.com/withastro/roadmap/issues/531)
- Stage 1 proposal
  - https://github.com/withastro/roadmap/discussions/469
  - https://github.com/withastro/roadmap/discussions/174

### DIFF
--- a/proposals/0032-middleware-api.md
+++ b/proposals/0032-middleware-api.md
@@ -68,14 +68,22 @@ For me, it would make handling authentication much easier.
 - Provide a way intercept requests and responses, allowing users to set cookies and headers
 - Works both in SSR and SSG mode
 - Allow users to use community-created middlewares (libraries)
-  - Make available via integrations API.
-- Provide an API for request-specific data
+
+
+# Out of Scope
+
+This is a list of requirements that won't be implemented as part of this RFC but 
+will be implemented in the second iteration of the middleware project:
+
 - Non-Node runtimes specific APIs. ie. Cloudflare Durable Objects.
-  - Add middleware from adapter.
+- Add middleware from adapter.
+- Type-safe payload. Being able infer types from the previous middleware.
 
 # Non-Goals
 
 - Route specific middleware, middlewares that are run **only on specific** routes
+
+
 
 # Detailed Design
 
@@ -93,7 +101,8 @@ For me, it would make handling authentication much easier.
 
 ## Implementation instructions
 
-To define a middleware, a user must create a physical file under the `src/` folder, called `middleware.js`.
+To define a middleware, a user must create a physical file under the [`config.srcDir`](https://docs.astro.build/en/reference/configuration-reference/#srcdir) 
+folder, called `middleware.js`.
 
 The resolution of the file follows the ECMA standards, which means that the following
 alternatives are all valid in Astro:
@@ -139,11 +148,22 @@ const onRequest: MiddlewareRequestHandler = async (context: APIContext, next: Mi
 export { onRequest }
 ```
 
+Alternatively, a user can use an utility API to type the middleware:
+
+```ts
+import {defineMiddleware} from "astro/middleware";
+
+
+const onRequest = defineMiddlware(async (context, next) => {
+    
+});
+```
+
 The `locals` object is a new API introduced by this RFC. The `locals` object is a new
 Astro global object that can be manipulated inside the middleware, and then it can be
 accessed inside any `.astro` file:
 
-```md
+```astro
 ---
 const user = Astro.locals.user;
 ---
@@ -464,6 +484,9 @@ export default defineConfig({
   }
 })
 ```
+
+Plus, a user can enable this experimental feature via CLI using a 
+new argument called `--experimental-middleware`.
 
 The team will seek feedback from the community and fix bugs if they arise.  
 After an arbitrary about of time, the experimental flag will be removed.
--- a/proposals/0032-middleware-api.md
+++ b/proposals/0032-middleware-api.md
@@ -453,13 +453,9 @@ requests, they will need to store that information somewhere else.
 
 ## Restrictions and expectations
 
-In order to set user expectations, the Astro middleware have the following restrictions:
-- a middleware needs to return a `Response`;
-- a middleware needs to call `next`;
+In order to set user expectations, the Astro middleware **must** return a `Response`.
 
-If the user doesn't do any of these two, Astro will throw an error. Plus,
-the user is required to return exactly a `Response`. Failing to do so will result in
-Astro throwing an error.
+If the user doesn't do that, Astro will throw an error.
 
 # Testing Strategy
 
--- a/proposals/0032-middleware-api.md
+++ b/proposals/0032-middleware-api.md
@@ -91,8 +91,6 @@ will be implemented in the second iteration of the middleware project:
 
 - `resolve` has been renamed to `next`;
 - `next` doesn't accept an `APIContext` to work;
-- `locals` values need to be serializable to avoid the introduction of
-  non-user code from third-party libraries that can run scripts;
 - `middleware` export function has been renamed `onRequest`. This name change
   has two benefits:
   1. It shows intent and explains when this function is called;
@@ -192,37 +190,7 @@ By doing so, the user can leverage the type-checking and auto-completion of Type
 called `middleware.ts` or `middleware.js` using JSDoc.
 
 
-The `locals` object has the following restrictions:
-1. it can store only serializable information;
-2. it can't be overridden by other values that are different from objects;
-
-## `locals` needs to be serializable
-
-The information must be serializable because storing
-information that evaluates at runtime is unsafe. If, for example, we were able to store
-With a JavaScript function, an attacker could exploit the victim's website
-and execute some unsafe code.
-
-Astro will do a sanity check **in development mode**.
-Some code like this:
-
-```js
-export const onRequest = (contex, next) => {
-    context.locals.someInfo = {
-        f() {
-            alert("Hello!!")
-        }
-    }
-} 
-```
-
-Storing unsafe information will result in an Astro error:
-
-> The information stored in Astro.locals are not serializable when visiting "/index" path.
-Make sure you store only serializable data.
-
-> **Note**: The content of the error is not final. The docs team will review it.
-
+The `locals` object can't be overridden by other values that are different from objects;
 
 ## `locals` can't be overridden
 
--- a/proposals/0032-middleware-api.md
+++ b/proposals/0032-middleware-api.md
@@ -243,6 +243,25 @@ Astro will emit an error like this:
 
 > **Note**: The content of the error is not final. The docs team will review it.
 
+`locals` can be typed using the `env.d.ts` file. In order to do so, this RFC will 
+expose a new `namespace` called `App`. This will be the first `namespace` exposed
+by Astro:
+
+```ts
+/// <reference types="astro/client" />
+declare global {
+    namespace App {
+        interface Locals {
+            user: {
+                name: string
+            }
+        }
+    }
+}
+
+export {}
+```
+
 ### `context` and `next`
 
 When defining a middleware, the function accepts two arguments: `context` and `next`
--- a/proposals/0032-middleware-api.md
+++ b/proposals/0032-middleware-api.md
@@ -42,8 +42,8 @@ export const onRequest = (context, next) => {
 Or, set some logic to make redirects:
 ```js
 const redirects = new Map([
-    ["/old_1", "/new_1"],
-    ["/old_1", "/new_1"]
+    ["/old-1", "/new-1"],
+    ["/old-1", "/new-1"]
 ])
 
 export const onRequest = (context, next) => {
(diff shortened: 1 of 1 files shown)


proposals/0032-middleware-api.md
```diff
@@ -0,0 +1,469 @@
...
+
+For me, it would make handling authentication much easier.
+
+
+# Goals
+
+- Provide a way intercept requests and responses, allowing users to set cookies and headers
+- Works both in SSR and SSG mode
+- Allow users to use community-created middlewares (libraries)
+  - Make available via integrations API.
```
matthewp, line 71: I think this part is not a goal for this phase of the API. To use 3rd party middleware you'd import them into your own src/middleware file.

proposals/0032-middleware-api.md
```diff
@@ -0,0 +1,469 @@
...
+- Provide an API for request-specific data
```
matthewp, line 72: Not sure I understand this one, isn't this covered by the first goal?

proposals/0032-middleware-api.md
```diff
@@ -0,0 +1,469 @@
...
+- Non-Node runtimes specific APIs. ie. Cloudflare Durable Objects.
+  - Add middleware from adapter.
+
+# Non-Goals
```
matthewp, line 76: You might add another section that is `# Out of Scope`. That would be things that are out of scope for this iteration of the API but are in scope for in the future such as:

- Running middleware on the edge, outside of the app itself, in production.
- Type-safe middleware, for ex. an API that would allow the types of `locals` to be mutated each time you add a new middleware to the chain.

proposals/0032-middleware-api.md
```diff
@@ -0,0 +1,469 @@
...
+  non-user code from third-party libraries that can run scripts;
+- `middleware` export function has been renamed `onRequest`. This name change
+  has two benefits:
+  1. It shows intent and explains when this function is called;
+  2. It allows adding more functions with the `on*` prefix in the future, which could show the intent
+     of when the function is called in the Astro route life cycle;
+
+## Implementation instructions
+
+To define a middleware, a user must create a physical file under the `src/` folder, called `middleware.js`.
```
matthewp, line 96: To be precise, this should be `config.srcDir`, see https://docs.astro.build/en/reference/configuration-reference/#srcdir

proposals/0032-middleware-api.md
```diff
@@ -0,0 +1,469 @@
...
+```
+
+Every time a page or endpoint is about to be rendered, the middleware is called.
+
+A middleware will look like this in TypeScript:
+
+```ts
+import { MiddlewareRequestHandler, APIContext, MiddlewareNextResponse } from "astro"
+
+const onRequest: MiddlewareRequestHandler = async (context: APIContext, next: MiddlewareNextResponse) => {
```
matthewp, line 128: Discussed on discord, but it could be nice to have a `defineMiddleware` export from `astro/middleware`. That would be like `defineConfig` in the Astro config. That function doesn't actually do anything other than make it so the what you pass to it is properly typed.

proposals/0032-middleware-api.md
```diff
@@ -0,0 +1,469 @@
...
+}
+
+export { onRequest }
+```
+
+The `locals` object is a new API introduced by this RFC. The `locals` object is a new
+Astro global object that can be manipulated inside the middleware, and then it can be
+accessed inside any `.astro` file:
+
+```md
```
matthewp, line 146: You can use `astro` in code blocks on GH.

proposals/0032-middleware-api.md
```diff
@@ -0,0 +1,469 @@
...
+While Astro could have provided some API to reduce the boilerplate, this felt tedious. 
+Plus, this would have forced us to **crate another configuration field** where the user
+could specify the order of the middleware.
+
+
+# Adoption strategy
+
+Considering how big the feature is, the API will be released under an experimental flag.
+
+Users can opt in via new flag:
```
matthewp, line 458: We usually also have a CLI flag, like `--experimental-middleware`. So you could either use the CLI flag or add it to the Astro config file.

matthewp:
One other comment, if you call `next()` but don't care about the result (because you don't need to do anything to the response), are you expected to return the return value? Ie do you have to do `return next();` or is just `next();` ok? I don't have an opinion on this and am kind of assuming that the first way is how it's currently implemented. If so I could see this as being a good thing, that you must return some response.

ematipico:
> One other comment, if you call `next()` but don't care about the result (because you don't need to do anything to the response), are you expected to return the return value? Ie do you have to do `return next();` or is just `next();` ok? I don't have an opinion on this and am kind of assuming that the first way is how it's currently implemented. If so I could see this as being a good thing, that you must return some response.

I suppose it really depends on how the user will it. `next();` and `return next()` will yield two different results. Without the `return`, the next middleware won't access to the `Response` object. If the user **knows** that that middleware won't need it, I suppose `next();` is a correct pattern.

lilnasy:
Could a middleware tell if it's running in dev mode?

The use case I'm thinking of is a middleare that caches responses. It should not act during dev.

lilnasy:
Could a middleware communicate with an integration?

The use case I'm thinking of is an integration that hashes inline resources, and a middleare that uses the hashes to create a strict Content-Security-Policy header.

I understand letting middleware be added by integrations was made out of scope for now, but receiving data from an integration alone would be valuable.

ematipico:
> Could a middleware tell if it's running in dev mode?
> 
> The use case I'm thinking of is where a middleare caches responses in memory, similar how to how Next's ISR works. The middleware should not act during dev.
> 
> cc: #228

Not at this stage. It wasn't part of the original RFC. Currently, the middleware is code that **goes in production**, and the `mode` is a piece of information that is lost after the build.

ematipico:
> Could a middleware communicate with an integration?
> 
> The use case I'm thinking of where an integration hashes inline resources, and a middleare uses the hashes to create a strict Content-Security-Policy header.
> 
> I understand letting middleware be added by integrations was made out of scope for now, but receiving data from an integration alone would be valuable.
> 
> cc: #377

We can explore this valid use case after the middleware is released under an experimental flag. During the exploration phase, I evaluated creating a middleware as an integration, but that was too far from the initial RFC.

lilnasy:
> > Could a middleware tell if it's running in dev mode?
> > The use case I'm thinking of is where a middleare caches responses in memory, similar how to how Next's ISR works. The middleware should not act during dev.
> > cc: #228
> 
> Not at this stage. It wasn't part of the original RFC. Currently, the middleware is code that **goes in production**, and the `mode` is a piece of information that is lost after the build.

@ematipico I looked around a little more, and mathhewp gave an example for a dev-only route in the [issue](https://github.com/withastro/roadmap/issues/531#issuecomment-1508933254). If that example works, prod-only ISR could too.

lilnasy:
> > Could a middleware tell if it's running in dev mode?
> > The use case I'm thinking of is where a middleare caches responses in memory, similar how to how Next's ISR works. The middleware should not act during dev.
> > cc: #228
> 
> Not at this stage. It wasn't part of the original RFC. Currently, the middleware is code that **goes in production**, and the `mode` is a piece of information that is lost after the build.

@ematipico I looked around a little more, and matthewp gave an example for a dev-only route in the [issue](https://github.com/withastro/roadmap/issues/531#issuecomment-1508933254). If that example works, prod-only ISR could too.

matthewp:
Yeah, I would expect `import.meta.env.MODE` to work just like it does in any other file in the project.

ematipico:
I wasn't aware of that metadata, and I personally haven't tried. But if that metadata is available in endpoints, it will definitely be available in middleware too!

snimavat:
I come from Java background, and `middleware` word has always been confusing to me.
If it does the job of intercepting requests/response - "Interceptors" seems more meaningful

ematipico:
> I come from Java background, and `middleware` word has always been confusing to me. If it does the job of intercepting requests/response - "Interceptors" seems more meaningful

Yeah, there are different names to address the same pattern. SvelteKit calls it "hooks", like fastify.

raulfdm:
> > I come from Java background, and `middleware` word has always been confusing to me. If it does the job of intercepting requests/response - "Interceptors" seems more meaningful
> 
> Yeah, there are different names to address the same pattern. SvelteKit calls it "hooks", like fastify.

In Next.js world they also call middleware, but their middleware is less capable than the pattern we're seeing here.

> https://nextjs.org/docs/pages/building-your-application/routing/middleware

gkatsanos:
how safe are we to use a middleware in **production** that intercepts a request and sets a global variable or a value in a nano store based on the request domain name?

ematipico:
> how safe are we to use a middleware in **production** that intercepts a request and sets a global variable or a value in a nano store based on the request domain name?

The feature is under an experimental flag for different reasons:
- it might contain bugs
- the APIs might still change, which are not semver complaint 

Although, the code written inside the middleware is stored in the backend (like the rest of the Astro code). It's advised to ensure you don't write any sensible data inside `Astro.locals`

pilcrowOnPaper:
Quick question, why should `locals` be serializable? 

> The information must be serializable because storing information that evaluates at runtime is unsafe. If, for example, we were able to store a JavaScript function, an attacker could exploit the victim’s website and execute some unsafe code

How would an attacker inject or execute unsafe code?

pilcrowOnPaper:
It also looks like API routes run before the middleware?

ematipico:
> It also looks like API routes run before the middleware?

That was a bug :) I believe it's been fixed in https://github.com/withastro/astro/pull/7106

pilcrowOnPaper:
@ematipico

I'm not sure serializing data prevents this?

```ts
locals.email = returnsUnsafeScript(); // <script>eval(alert('hacked'))</script>
```

>  if we were able to store a JavaScript function, an attacker could exploit the victim’s website and execute some unsafe code

But if you're using third party dependencies, they can already run unsafe code:

```ts
// node_modules/some-library/index.js

export someFunction = () => {};

console.log("unsafe code");
```

Not having the ability to store any data type seems to be a massive limitation IMO.

ematipico:
> I'm not sure serializing data prevents this?

It indeed doesn't; probably, we should also escape the strings, just to be sure.

> But if you're using third-party dependencies, they can already run unsafe code:

It's more about **data** that gets **rendered** on the page, not about the code per se. 


> Not having the ability to store any data type seems to be a massive limitation IMO.

It is, I know! What would you suggest instead? How can we support more data types without overlooking the security?

tropperstyle:
I have a request to expose the exports of the matched route module to the middleware.

Since auth guards are a primary example for this proposal, this could help avoid path logic growing in the middleware and allow explicitly flagging pages as needed:

```
---
export const requireAdmin = true;
---
```

```
// Make all non-Astro defined exports available to context
export const onRequest = ({ exports, request, redirect, locals }, next) => {
  if (exports.requireAdmin) {
    locals.admin = await fetchAdmin(request);
    if (!locals.admin) return redirect('/login');
  }
  return next();
}
```

pilcrowOnPaper:
@ematipico 

Doesn't Astro escape strings inside html templates anyway? If you’re using `set:html`, it should be your responsibility to make sure your dependencies are safe. I really think this is a non issue that doesn’t need to be addressed. SvelteKit has locals for like 2 years and I’m not aware of any html injection, and Express has had it for even longer.

jtropper-corefact:
I also do not understand the security threat that requires `locals` to be serializable.

To add another use-case, I would hope to use this API to defer a Promise. Meaning, I'd like to _immediately invoke_ an async function on page request, but not block streaming until the data is needed.

page.astro
```
---
import Items from './_items.astro';
Astro.locals.itemsPromise = fetch('/api/items'); // do not await
---
<html>
  ...
  <Items />
```

_items.astro
```
---
const items = await Astro.locals.itemsPromise;
---
{items.map(item => (
  ...
```

If the items are needed in multiple components, choices are:
1. Block the page from initially streaming and prop-drill all components.
2. Repeat the fetch within each component, but face duplicate calls and delayed execution of those calls

matthewp:
I think the primary driver for the serializable restriction is that we plan to make it possible for adapters (like Vercel) to split the middleware into a separate bundle to deploy to an Edge. This way you can have things like auth-checks happen at edge locations and then your app code in serverless.

For this to work we need to be able to serialize locals to send from the edge worker to the origin. We weren't sure that people would even notice this!

But it sounds like a lot of people are wanting to put complex objects onto locals. This would just mean you couldn't use the edge feature. But for a lot of people that's a fine tradeoff. So I think this user feedback is a good reason to lift the restriction.

tropperstyle:
Sorry for the deleted comment, and thanks for the clarification about the edge workers.

So if I want to:
1. Immediately invoke an async function on page request, but not block streaming until the data is needed
2. Only call this function once, even though the data might be needed for multiple components

Without the serializable restriction, with `locals` it looks like:

page.astro
```
---
import Items from './_items.astro';
Astro.locals.itemsPromise = fetch('/api/items'); // do not await
---
<html>
  ...
  <Items />
```

_items.astro
```
---
const items = await Astro.locals.itemsPromise;
---
{items.map(item => (
  ...
```

Should there be a more idiomatic way to go about this to reserve `locals` for pure data?

matthewp:
cc @ematipico Curious what you think about the above 👆

matthewp:
@ematipico Ah yeah, if that's a restriction of that use-case it does make senes to move the check there. I'm not sure exactly how that will be implemented, but I imagine that the adapter will add it's own middleware at the end of the chain. It can do the check there.

matthewp:
@gkatsanos I'm not very familiar with nanostores, but looking at the code it appears that `brand` is a global which doesn't work well in a server environment, you have multiple requests all being processed at the same time.

I would recommend putting `branch` onto the `locals` object.

pilcrowOnPaper:
For the Vercel adapter, if we're going to add it, please make deploying to Vercel's Edge middleware optional. The biggest limitation with Next.js' Middleware is that you can't deploy it to just a regular serverless server(?).

matthewp:
@pilcrowOnPaper Yeah, that will definitely be an opt-in feature.

matthewp (approved):

Issue #556, opened 2023-04-14
CSS Inlining


matthewp:
- Accepted Date: 2023-04-14
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/343
- Author: @lilnasy
- Champion(s): @lilnasy
- Implementation PR: https://github.com/withastro/astro/pull/6659

# Summary

Provide a configuration option to specify inlining behavior of styles authored or imported in astro modules.

# Background & Motivation

There has been a constant interest in inlining styles while still taking advantage of scoping and other processing steps since before 1.0 (see: withastro/astro#918), with many users incorrectly assuming that `is:inline` directive is the solution (see: withastro/astro#6388).

Simple one-page websites do not benefit from an external stylesheet, since there is no other page that could use the cached stylesheet. On the other hand, large websites are overoptimized for cacheability, since our chunking splits the styles too granularly. Case in point, Astro docs homepage has 20 linked stylesheets, 14 of them are less than 1kb (see: withastro/astro#6528).

So far we have not provided a way to allow inlining stylesheets, prompting workarounds. However, coming from other frameworks, users might expect to be able to configure this behavior.
- SvelteKit allows configuring a threshold under which CSS files will be inlined:  https://kit.svelte.dev/docs/configuration#inlinestylethreshold
- Fresh lets plugins inject inline styles: https://fresh.deno.dev/docs/concepts/plugins#hook-render
- Nuxt has a global configuration: https://nuxt.com/docs/api/configuration/nuxt-config#inlinessrstyles
at Astro and why this problem is worth solving. Why solve it now? Be brief!

# Goals

- Provide a way to reduce the number of HTTP requests for stylesheets.
- Maintain current behavior when not explicitly configured.
- Works with both `server` and `static` outputs.

# Non-Goals

- Identify "critical" CSS rules.
- Enable a single external stylesheet.
- Preloading.
- Inlining of scripts

matthewp:
RFC: https://github.com/withastro/roadmap/pull/557

jessaleks:
Hey there, great idea! 

Playing devil's advocate: how will it be better/different than critters, for which there is an integration for Astro already?

lilnasy:
> Hey there, great idea!
> 
> Playing devil's advocate: how will it be better/different than critters, for which there is an integration for Astro already?

@jessaleks 
- fast
- ssr

matthewp:
This RFC was merged as approved.

PR #557, opened 2023-04-14
Inline stylesheets configuration

lilnasy:
# Summary
Provide a configuration to control inlining behavior of styles authored or imported in astro modules.

# Links
* [Full rendered proposal](https://github.com/lilnasy/roadmap/blob/inline-stylesheets/proposals/0032-inline-stylesheets.md)
* [Stage 2 proposal](https://github.com/withastro/roadmap/issues/556)  #556
* [Stage 1 proposal](https://github.com/withastro/roadmap/discussions/343#discussioncomment-5539608)  #343

### DIFF
--- a/proposals/0032-inline-stylesheets.md
+++ b/proposals/0032-inline-stylesheets.md
@@ -20,7 +20,7 @@ export default defineConfig({
 ```ts
 export default defineConfig({
     build: {
-        // stylesheets smaller than `ViteConfig.build.assetsInlineLimit` (default: 4kb) are inlined
+        // stylesheets smaller than `ViteConfig.build.assetsInlineLimit` (or 4kb if not defined by the user) are inlined
         inlineStylesheets: "auto"
     }
 })
--- a/proposals/0032-inline-stylesheets.md
+++ b/proposals/0032-inline-stylesheets.md
@@ -1,0 +1,89 @@
+
+- Start Date: 2023-04-15
+- Reference Issues: [#556](https://github.com/withastro/roadmap/issues/556)
+- Implementation PR: [withastro/astro#6659](https://github.com/withastro/astro/pull/6659)
+
+# Summary
+
+Provide a configuration to control inlining behavior of styles authored or imported in astro modules.
+
+# Example
+
+```ts
+export default defineConfig({
+    build: {
+        // all styles necessary are sent in external stylsheets, default; maintains current behavior
+        inlineStylesheets: "never"
+    }
+})
+```
+```ts
+export default defineConfig({
+    build: {
+        // stylesheets smaller than `ViteConfig.build.assetsInlineLimit` (default: 4kb) are inlined
+        inlineStylesheets: "auto"
+    }
+})
+```
+```ts
+export default defineConfig({
+    build: {
+        // all styles necessary are inlined into <style> tags
+        inlineStylesheets: "always"
+    }
+})
+```
+
+# Background & Motivation
+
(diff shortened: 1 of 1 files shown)


proposals/0032-inline-stylesheets.md
```diff
@@ -0,0 +1,89 @@
...
+    build: {
+        // all styles necessary are sent in external stylsheets, default; maintains current behavior
+        inlineStylesheets: "never"
+    }
+})
+```
+```ts
+export default defineConfig({
+    build: {
+        // stylesheets smaller than `ViteConfig.build.assetsInlineLimit` (default: 4kb) are inlined
```
ematipico, line 23: Relying on the external tool has another drawback that we should highlight: if `vite` decides to change its defaults, we are forced to create a **major** release because it changes **astro's defaults**, unless we explicitly say that this API is subject to Vite's behaviour and Astro's doesn't intend to respect it.

proposals/0032-inline-stylesheets.md
```diff
@@ -0,0 +1,89 @@
...
+
+Astro has run into CSS ordering issues and currently ensures that the `<link>` tags are in a specific order. Currently, this order can't be guaranteed when some stylesheets are to be inlined as `<style>` tags: the rendering pipeline receives the two separtely and inserts all the `<link>` tags first. Therefore, an implementation must pass both inline and external stylesheets in the same array or set that preserves the order, and the rendering pipeline should make sure each `SSRElement` gets serialized with the appropriate tag.
+
+# Testing Strategy
+
+An implementation can be tested adequately with fixture-based tests, ensuring the DOM from SSR responses and generated `.html` files has the expected count of style and link tags and that they are placed in the expected order.
+
+# Drawbacks
+
+- Might lead to unexpected results with our current CSS chunking strategy - it tends to create a lot of tiny stylesheets, and each of them might be under the limit, leading to more CSS being inlined than the user might expect.
```
ematipico, line 75: I would say that another Drawback of this new option is that it could increase the disparity between the `development` mode and the `production` mode.

proposals/0032-inline-stylesheets.md
lilnasy: We are relying not on vite, but on vite subconfig in Astro config, the [defaults](https://github.com/withastro/astro/blob/main/packages/astro/src/core/config/schema.ts#L36) for which [we decide](https://github.com/withastro/astro/blob/2d8dad1a/packages/astro/src/core/build/plugins/plugin-css.ts#L222).

proposals/0032-inline-stylesheets.md
lilnasy: Dev mode inlines, so this would let you opt into reducing the disparity.

ematipico: If you chose `"auto"`, would that generate a different result from `inlineStylesheets: "never"` and `development`?

proposals/0032-inline-stylesheets.md
ematipico: Sorry, that wasn't clear! (I don't know the internals of that part)

lilnasy: You are right, that isn't clear. At one point, I realized it might sound like 4kb is vite's default being used, but I forgot about it soon.

Editing to clarify.

proposals/0032-inline-stylesheets.md
lilnasy: yes, it would

matthewp:
This feature has been available since 2.4 and this RFC up for a month without comment. Given that, I think this feature is safe to do a call for consensus.

This will be the final comment period (~3 days), if there are issues with this RFC please voice them now, thank you.

matthewp:
The final comment period is over and there are no objections, so this change will be merged in.

Issue #539, opened 2023-03-24
Enhance hybrid rendering to prerender all pages by default


Princesseuh:
### Body

- Accepted Date: 24/03/23
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/455
- Author: @zadeviggers
- Core Champions: @natemoo-re, @FredKSchott 
- Implementation PR: 

# Summary

Make an enhancement to Astro's [Hybrid Rendering](https://docs.astro.build/en/guides/server-side-rendering/#hybrid-rendering) support to make pre-rendering the default for all pages, and opt-out instead of opt-in.

# Background & Motivation

Currently, pre-rendering pages in SSR is possible thanks to [Hybrid Rendering](https://docs.astro.build/en/guides/server-side-rendering/#hybrid-rendering). However, this uses `output: 'server'` which defaults all pages to server-rendered. This can be annoying for mostly-static sites that then need to add `export const prerender = true;` to most pages.

One common use-case (mentioned in the original pre-rendering RFC) is a fully static site with a few server-side api routes. The current implementation requires a migrating user to add `output: 'server'` to their configuration and then manually add `export const prerender = true` to each `.astro` page in their `src/pages` directory before continuing.

This was originally discussed in the pre-rendering RFC. However it was ultimately removed because it was assumed that this would require a new third internal implementation (`static` and `server` (SSR) already exist today) which was considered too much of a maintenance burden to take on.

# Goals

- Support some version of Hybrid Rendering where pages default to `export const prerender = true`
- Avoid a third internal build implementation. A new `output` mode in the user-facing configuration API is okay, as long as it leverages the existing `static` or `server` internal build implementation.

# Non-Goals 

- N/A

zadeviggers:
I'm excited to see progress on this!

Since creating this original proposal, I've realised that support from the editor tooling will be needed to make this feature more accessible to users. An indicator will be needed to communicate to the user what output mode is set for a particular page - perhaps an icon in the corner or a message in the status bar at the bottom. This will be even more important if the `output: 'hybrid'` (see below) route is chosen, but both options would benefit from this indicator.


There was also some discussion around adding an `output: 'hybrid'` mode that enables this behaviour to avoid a breaking(?) change to build outputs. 
[...]

PR #565, opened 2023-04-27
Hybrid rendering stage 3 proposal

matthewp:
# Summary

Provide a `output: 'hybrid'` mode to allow static sites to opt-in certain pages to be dynamic (not prerendered).

# Links
* [Full rendered proposal](https://github.com/withastro/roadmap/blob/213cf99cd9ded5ac03b300366e79da21bb5f4332/proposals/hybrid-rendering.md)
* [Stage 2 proposal](https://github.com/withastro/roadmap/issues/539)
* [Stage 1 proposal](https://github.com/withastro/roadmap/discussions/455)

### DIFF
--- a/proposals/hybrid-rendering.md
+++ b/proposals/hybrid-rendering.md
@@ -62,7 +62,7 @@ The Astro config definition and types will need to be updated to allow the `'hyb
 
 In `packages/astro/src/core/routing/manifest/create.ts` each route is set up to be `prerender: false` by default. This should be changed to be based on the `output` config option. If `'server'` then it should remain false, if `'hybrid'` it should be interpreted as true.
 
-In `packages/astro/src/vite-plugin-scanner/scanner.ts` it currently throws for falsey values. Since in hybrid rendering users will set `export const prerender = false`, this code will need to be updated to allow the false value when the `output: 'hybrid'`.
+In `packages/astro/src/vite-plugin-scanner/scan.ts` it currently throws for falsey values. Since in hybrid rendering users will set `export const prerender = false`, this code will need to be updated to allow the false value when the `output: 'hybrid'`.
 
 Additionally there are a few places in the codebase that assume if `output !== 'static'` that it is server mode. Those might need to be changed, depending on what they are doing with that information. For example, in static mode you cannot access the `Astro.url.searchParams`. However, because this happens before we load a page component we cannot know at the time if the route is prerendered or not, so we cannot continue to enforce this restriction in hybrid rendering. For this reason, and for simplicity, in hybrid rendering these restrictions are lifted for all routes.
 
--- a/proposals/hybrid-rendering.md
+++ b/proposals/hybrid-rendering.md
@@ -1,0 +1,106 @@
+- Start Date: 2023-04-27
+- Reference Issues: https://github.com/withastro/roadmap/issues/539
+- Implementation PR: <!-- leave empty -->
+
+# Summary
+
+Provide a new `output` option `'hybrid'` that treats all pages as prerendered by default, allowing an opt-out through `export const prerender = false`.
+
+# Example
+
+An existing static site can be changed to hybrid rendering, allowing some specific routes to not be prerendered.
+
+__astro.config.mjs__
+
+```js
+import { defineConfig } from 'astro/config';
+
+export default defineConfig({
+  output: 'hybrid'
+});
+```
+
+__pages/api/form.ts__
+
+```ts
+export const prerender = false;
+
+export function get({ request }) {
+  // ...
+}
+```
+
+# Background & Motivation
+
+In 2.0 Astro introduced prerendering as an option when using `output: 'server'`. Prerendering allows certain pages to be prerendered to HTML during the build. This means that a dynamic app can have certain pages, like a landing page be served faster through a CDN, while preserving dynamic pages to SSR.
+
+An immediate point of feedback from the community was that it felt odd that this worked in server output, but not in static output given Astro's roots as a static site generator. This choice was made to prevent maintenance burden of having a 3rd way that the build works. A build where some routes are not prerendered is more like the `'server'` output than like the `'static'`.
+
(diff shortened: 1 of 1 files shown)


proposals/hybrid-rendering.md
```diff
@@ -0,0 +1,106 @@
...
+
+However, there are use-cases and reasons for wanting to have some dynamic pages in a static site.
+
+## Use-cases
+
+A few of the use-cases collected when talking to users who have requested this feature:
+
+- A marketing site for an agency that contains a contact form. Most of the site can be static and served via CDN, but the endpoint to serve the contact form needs to be dynamic so that it can store the contact information in a database and alert the admin.
+- A SaaS product where the API is the main product. Dynamic parts of pages are built with client components, API endpoints are the only server routes needed.
+- A content site such as a recipe site that allows users to mark their favorite recipes. API routes would be dynamic, as would any pages that display the dynamic list sof favorites.
```
MoustaphaDev, line 47:
```suggestion
- A content site such as a recipe site that allows users to mark their favorite recipes. API routes would be dynamic, as would any pages that display the dynamic list of favorites.
```

proposals/hybrid-rendering.md
```diff
@@ -0,0 +1,106 @@
...
+
+- Any extra features outside of marking certain pages to not be prerendered.
+
+# Detailed Design
+
+The Astro config definition and types will need to be updated to allow the `'hybrid'` value for `output`.
+
+In `packages/astro/src/core/routing/manifest/create.ts` each route is set up to be `prerender: false` by default. This should be changed to be based on the `output` config option. If `'server'` then it should remain false, if `'hybrid'` it should be interpreted as true.
+
+In `packages/astro/src/vite-plugin-scanner/scanner.ts` it currently throws for falsey values. Since in hybrid rendering users will set `export const prerender = false`, this code will need to be updated to allow the false value when the `output: 'hybrid'`.
```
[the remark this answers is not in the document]
matthewp: Oh ok, so if it's not throwing for `false` currently then we probably don't need to change this behavior at all. No need to add new error messages that did not previously exist.

jamesli2021:
As Opposite to 'serve', using 'hybrid' is confusing, I prefer 'defaultstatic' as it's more meaningful in the docs.

proposals/hybrid-rendering.md
```diff
@@ -0,0 +1,106 @@
+- Start Date: 2023-04-27
+- Reference Issues: https://github.com/withastro/roadmap/issues/539
+- Implementation PR: <!-- leave empty -->
```
MoustaphaDev, line 3:
```suggestion
- Implementation PR: https://github.com/withastro/astro/pull/6991
```

proposals/hybrid-rendering.md
```diff
@@ -0,0 +1,106 @@
...
+
+__astro.config.mjs__
+
+```js
+import { defineConfig } from 'astro/config';
+
+export default defineConfig({
+  output: 'hybrid'
+});
+```
```
MoustaphaDev, line 21: Suggesting this to make it clearer that hybrid requires an adapter
```suggestion
```diff
import { defineConfig } from 'astro/config';
import vercel from "@astrojs/vercel"

export default defineConfig({
-  output: 'static',
+  output: 'hybrid',
+  adapter: vercel()
});
```
```

File: proposals/hybrid-rendering.md [hunk 5]
```diff
@@ -0,0 +1,108 @@
...
+
+Additionally there are a few places in the codebase that assume if `output !== 'static'` that it is server mode. Those might need to be changed, depending on what they are doing with that information. For example, in static mode you cannot access the `Astro.url.searchParams`. However, because this happens before we load a page component we cannot know at the time if the route is prerendered or not, so we cannot continue to enforce this restriction in hybrid rendering. For this reason, and for simplicity, in hybrid rendering these restrictions are lifted for all routes.
+
+# Testing Strategy
+
+Prerendering is currently tested via fixture testing, due to the fact that the build artifacts is what changes. This same strategy will be used to test hybrid rendering as well, only testing for the opposite effect. 
+
+Likely we can use the same test fixtures, but only swap out the `output` when testing `'hybrid'`, which eliminates the need for new fixtures.
```
MoustaphaDev, line 75: In some cases we will need to create new test fixture, like tests in this PR https://github.com/withastro/astro/pull/7150
```suggestion
Likely we can use most of the prerendering test fixtures, and only swap out the `output` when testing `'hybrid'`, which reduces the need for new fixtures.
```

matthewp:
This has been out as experimental since 2.5. I'm moving to have this RFC merged in and experimental label removed. We'll now enter the final comment period. If anyone has issue with this RFC being merged please voice now, thanks!

zadeviggers:
Again, I'm still mainly in favor of not introducing a `"hybrid" ` option, and instead going for an object configuration. But I can see how this way is simpler for the user, if less future-proof and more confusing, so I don't feel too strongly either way.

matthewp:
@zadeviggers Did I miss this idea? Could you link to it? The object configuration I mean.

zadeviggers:
@matthewp here are the relevant comments:
https://github.com/withastro/roadmap/issues/539#issuecomment-1492673941
https://github.com/withastro/roadmap/issues/539#issuecomment-1492895619

matthewp:
Ah yeah. I see what you mean. Personally I would keep `output` as a string and then have a separate option for the default:

```js
output: 'server',
build: {
  prerenderDefault: false
}
```

I think there's still some perception problem with this approach where people see the default being false as being the wrong choice. 

I think we should proceed with the current `'hybrid'` option for now. 3.0 is coming up so we can still make this better in the future. Since this is mostly about bikeshedding the config name and not how exactly it works. Sound fair?

matthewp:
We're going to go with this approach for now. Happy to continue refining it as we approach 3.0.

zadeviggers:
Awesome, that sounds good.

Issue #530, opened 2023-03-22
Referencing data from content collections


bholmesdev:
## Details 
- Accepted Date: 24/03/23
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/477, https://github.com/withastro/roadmap/discussions/525
- Author: @bholmesdev
- Core Champions: @bholmesdev, @tony-sull
- Implementation PR: 

## Summary

Introduce a standard to store data separately from your content (ex. JSON files), with a way to "reference" this data from existing content collections.

## Background & Motivation

Content collections are restricted to supporting `.md`, `.mdx`, and `.mdoc` files. This is limiting for other forms of data you may need to store, namely raw data formats like JSON.

Taking a blog post as the example, there will likely be author information thats reused across multiple blog posts. To standardize updates when, say, updating an author's profile picture, it's best to store authors in a separate data entry, with an API to reference this data from any blog post by ID.

The content collections API was built generically to support this future, choosing format-agnostic naming like `data` instead of `frontmatter` and `body` instead of `rawContent`. Because of this, expanding support to new data formats without API changes is a natural progression.

### Use cases

We have a few use cases in mind considering data collections and data references. We expect this list to grow through the RFC discussion and learning from our community!

- **Blog post meta info.** Common cases include author bios, project contributors, and tags
- **i18n translations.** Many content sites and translation libraries work from key / value pairs stored as JSON. For example, an `i18n/` collection containing `en.json`, `fr.json`, etc.
- **Image asset metadata.** You may want to reference reusable `alt` text or image widths and heights for standard assets. For example, an `images/banner.json` file containing the `src` as a string, alt text, and a preferred `width`

## Goals
- **Introduce JSON collection support,** configurable and queryable with similar APIs to content collections.
- **Determine where data collections are stored.** We may introduce a new `src/data/` directory distinct from `src/content/`, or simply allow data collections within `src/content/`.
- **Introduce an API to reference this data from existing content collections by ID.** This is based on the strongest user need for data collections: referencing metadata (ex. pull in post authors from a blog post).
- **Consider** Both one-to-one and one-to-many relationships between content and data (ex. allow passing a list of author IDs in your frontmatter).

## Non-goals

- User-facing APIs to introduce new data collection formats like YAML or TOML. We recognize the value of community plugins to introduce new formats, and we will experiment with a pluggable API internally. Still, a finalized user-facing API will be considered out-of-scope.

bholmesdev:
[...]

PR #569, opened 2023-05-02
Data collections and collection references

bholmesdev:
# Summary

This introduces two related features to content collections:
- **Data collections:** A way to store JSON files as collections
- **References:** A way to reference collection entries from another collection

# Links
- [Fully rendered proposal 1 - data collections](https://github.com/withastro/roadmap/blob/045bd4958f1aa28228ad0294aa43f3a3922414cd/proposals/0033-data-collections.md)
- [Fully rendered proposal 2 - collection references](https://github.com/withastro/roadmap/blob/045bd4958f1aa28228ad0294aa43f3a3922414cd/proposals/0034-collection-references.md)
- Stage 2 proposal: https://github.com/withastro/roadmap/issues/530
- Stage 1 proposals: https://github.com/withastro/roadmap/discussions/477, https://github.com/withastro/roadmap/discussions/525

https://user-images.githubusercontent.com/51384119/235771035-f48d162c-84f0-424b-997f-45da55252597.mp4


proposals/0034-collection-references.md
```diff
@@ -0,0 +1,214 @@
...
+    title: z.string(),
+    banner: reference('banners'),
+    authors: z.array(reference('authors')),
+    relatedPosts: z.array(reference('blog')),
+  })
+});
+
+const banners = defineCollection({
+  type: 'data',
+  schema: ({ image }) => z.object({
```
FredKSchott, line 115: Forgive me if this was already discussed, but seeing this all laid out now: it feels like a leaky implementation detail that `image` needs to come from a special schema function, but `z` and `reference` don't. There's no reason for this other than "the implementation of image() is special and needs this", correct?

Thoughts on passing `z` and `reference` to the schema function as well, and beginning to push users to always the function signature API instead? Looking at that API in the snippet below and it does seem more consistent to pass all schema helpers the same way. It's also a fairly small change that shouldn't be too disruptive to developers, and we could do it in a backwards-compatible way.


```diff
- import { defineCollection, reference, z } from 'astro:content';
+ import { defineCollection } from 'astro:content';

const blog = defineCollection({
-  schema: z.object({
+  schema: ({z, reference}) => z.object({
    title: z.string(),
    banner: reference('banners'),
    authors: z.array(reference('authors')),
    relatedPosts: z.array(reference('blog')),
  }),
});
const banners = defineCollection({
  type: 'data',
-  schema: z.object({
+  schema: ({z, image}) => z.object({
    src: image(),
    alt: z.string(),
  })
});
const authors = defineCollection({
  type: 'data',
-  schema: z.object({
+  schema: ({z}) => z.object({
    name: z.string(),
  })
});
export const collections = { blog, banners, authors };
```

Not a hill I'll die on, but I think it's worth the change IMO in the name of consistency and removing leaky implementation details!

proposals/0033-data-collections.md
```diff
@@ -0,0 +1,244 @@
...
+
+Early proposals supported single0file data collections. This allows storing _all_ data collection entries as an array in one `.json` file, instead of splitting up entries per-file as we do content collections today. For example, a single `src/content/authors.json` file instead of a few `src/content/authors/[name].json` files.
+
+We want to stick with a single API design for an experimental release. Unlike multi-file, there are some prohibitive reasons against single-file collections: 
+
+- **Big arrays don't scale for complex data entries**, like i18n translation files. Users will likely want to split up by file here, especially where the status quo is `en.json | fr.json | jp.json ...` for this use case.
+- **We'd need to parse the whole array of entries** to determine entry IDs. This could be a performance bottleneck vs. pulling IDs from file names.
+- **It would be different from content collections,** which means a learning curve.
+
+Due to these, we decided against single-file for now. Though we do recognize the convenience of colocation that can be explored in the future.
```
FredKSchott, line 218: +1 to exploring in the future. A solution that I'd love to see explored more in a future RFC that I think would kill a few birds with one stone:

```diff
+ import authorsData from '../authors.json';

const authors = defineCollection({
  type: 'data',
  schema: z.object({
    name: z.string(),
    twitter: z.string().url(),
  }),
+  data: authorsData,
});
```

Just a thought / not at all blocking this RFC which I agree is smart to treat this as out-of-scope!

proposals/0033-data-collections.md
```diff
@@ -0,0 +1,244 @@
...
+
+**Full background:** `slug` was originally introduced alongside the content `id` to have a URL-friendly version of the file path, which can be passed to `getStaticPaths()` for route generation. Data collections are not intended to be used as routes, so we don't want to perpetuate this pattern. `slug` also "slugifies" the file path by removing capitalization and replacing spaces with dashes. If we added this processing to data collection IDs, [collection references](https://github.com/withastro/roadmap/blob/d89e2a4c28379108501aa6bf40d2f8d93d81ad02/proposals/0034-collection-references.md) will be less intuitive to use (i.e. "do I include spaces in the referenced ID here?").
+
+## Implementation
+
+Data collections should following the API design of content collections with a stripped-down featureset. To wire up data collections, we will introduce an internal utility that mirrors our `addContentEntryType()` integration function. This example registers a new `dataEntryType` for `.json` files, with necessary logic to parse data as a JS object:
+
+```ts
+addDataEntryType({
```
FredKSchott, line 177: Just confirming that `JSON` support is built-in to Astro without needing the user to define `addDataEntryType()` themselves, correct? Are any other data types supported out of the box (csv? yaml?) It would be good to make a note of this here in the RFC if this is purely illustrative and not expected that every user copy-pastes this snippet into their project.

proposals/0033-data-collections.md
```diff
@@ -0,0 +1,244 @@
...
+
+<h1>Our Authors</h1>
+<ul>
+  {authors.map(author => (
+    <li><a href={author.data.twitter}>{author.data.name}</a></li>
+  ))}
+</ul>
+```
+
+To retrieve individual entries, `getEntry()` can be used. This receives both the collection name and the entry `id` as described in the [Return type](#return-type). These can be passed as separate arguments or as object keys.
```
FredKSchott, line 131: Huge fan of this new `getEntry` API! It's a clever solution to the ID vs. slug naming complexity you outline below.

proposals/0033-data-collections.md
```diff
@@ -0,0 +1,244 @@
...
+  schema: z.object({
+    name: z.string(),
+    twitter: z.string().url(),
+  })
+});
+
+export const collections = { authors };
+```
+
+These collections can also be queried using the `getCollection()` and `getEntry()` utilities:
```
FredKSchott, line 37: You mention `getEntries()` in the video, but I did a quick search on the page and didn't see that here. Is that included or not? If included, it should be defined here!

proposals/0033-data-collections.md
bholmesdev: Ah yes, I should definitely clarify! JSON is built-in, and this API is only detailed to show how the feature is implemented internally. Nothing else is included at the moment but I think YAML is worth adding before release; this would make copying frontmatter into a data file more direct.

proposals/0033-data-collections.md
bholmesdev: Good point! We should add that 👍

proposals/0034-collection-references.md
matthewp: I think that image is special (it's contextual) *is* a good reason for it to be accessed from a different place. It's not really an implementation detail, we couldn't have implemented it differently. 

Aside from that, given the discussions about "Content Collections v2" it seems like unnecessary churn to move `z` now when defining schemas might look quite different in the near future anyways. `reference` I feel less strongly about. Having it go through the function args and the function args representing "special Astro schema helpers" seems fine to me.

proposals/0033-data-collections.md
matthewp: What's the use-case for importing like this? It won't work with remote data, I'm surprised at the suggestion.

proposals/0033-data-collections.md
matthewp: JSON + YAML seems like reasonable defaults to me.

snimavat: +1 for YAML

proposals/0034-collection-references.md
bholmesdev: I agree with Matthew on moving `z` here. I think allowing `reference()` to come from the arrow function _or_ an import is an okay compromise to meet users where they are

proposals/0033-data-collections.md
bholmesdev: Update: added to the references doc. `getEntries()` is meant for resolving references primarily and probably shouldn't be recommended for manual use. In fact, we should add a documentation note that "if you're calling `getEntries()` on its own, you're probably looking for `getCollection()`"

proposals/0033-data-collections.md
```diff
@@ -0,0 +1,244 @@
...
+
+Content collections are restricted to supporting `.md`, `.mdx`, and `.mdoc` files. This is limiting for other forms of data you may need to store, namely raw data formats like JSON or YAML.
+
+Taking a blog post as the example, there will likely be author information thats reused across multiple blog posts. To standardize updates when, say, updating an author's profile picture, it's best to store authors in a standalone data entry.
+
+The content collections API was built generically to support this future, choosing format-agnostic naming like `data` instead of `frontmatter` and `body` instead of `rawContent`. Because of this, expanding support to new data formats without API changes is a natural progression.
+
+# Goals
+
+- **Introduce JSON and YAML collection support,** configurable and queryable with similar APIs to content collections.
```
ematipico, line 65: I am very skeptical about supporting `YAML` files inside data collections. What are the expectations of data types supported by YAML that don't exist in JSON? E.g. timestamps and dates

 I believe this RFC should address this difference and expectations of how YAML data types are translated to a JavaScript object.

proposals/0033-data-collections.md
```diff
@@ -0,0 +1,244 @@
...
+- **Determine where data collections are stored.** We may allow data collections within `src/content/`, or introduce a new reserved directory.
```
ematipico, line 66: Should we update this line and just say that `src/content/` directory has been chosen?

proposals/0033-data-collections.md
```diff
@@ -0,0 +1,244 @@
...
+})
+
+export const collections = { authors };
+```
+
+## Return type
+
+Data collection entries include the same `id`, `data`, and `collection` properties as content collections:
+
+- `id (string)` - The entry file name with the extension omitted. Spaces and capitalization are preserved.
```
ematipico, line 102: `id` is the name with extension omitted, does the RFC addresses cases where two files have the same name but **different extensions**?

```
src/content/blog/post.json
src/content/blog/post.yaml
```

proposals/0033-data-collections.md
```diff
@@ -0,0 +1,244 @@
...
+
+# Drawbacks
+
+- Data collections could be considered added complexity, when users can create `.md` files storing all data via frontmatter instead. Though true for trivial content, this is restrictive for use cases like i18n lookup files (which are typically JSON).
+
+# Alternatives
+
+## Using multi-file vs. single-file
+
+Early proposals supported single0file data collections. This allows storing _all_ data collection entries as an array in one `.json` file, instead of splitting up entries per-file as we do content collections today. For example, a single `src/content/authors.json` file instead of a few `src/content/authors/[name].json` files.
```
ematipico, line 210:
```suggestion
Early proposals supported single file data collections. This allows storing _all_ data collection entries as an array in one `.json` file, instead of splitting up entries per-file as we do content collections today. For example, a single `src/content/authors.json` file instead of a few `src/content/authors/[name].json` files.
```

ematipico (commented):
First round of question about **data collections**

proposals/0033-data-collections.md
bholmesdev: Probably... could remove the goal entirely?

proposals/0033-data-collections.md
bholmesdev: This should raise an error! Can clarify in the PR. Couldn't find a compelling reason to support this case.

proposals/0033-data-collections.md
bholmesdev: YAML should have the same parsing logic and JS object resolution as frontmatter. Reason in favor: users can simply copy frontmatter into a YAML file without any changes. This should make refactoring shared frontmatter into a data collection + reference a bit cleaner.

matthewp: I think YAML makes sense because it's easier for humans to author than is JSON. Imagine writing an author bio in a JSON string. Or recipe steps.

proposals/0033-data-collections.md
ematipico: Yeah I think so :)

proposals/0033-data-collections.md
```diff
@@ -0,0 +1,243 @@
...
+
+    if (data == null || typeof data !== 'object')
+      throw new Error(`JSON collection entry ${fileUrl.pathname} must be an object.`);
+
+    return { data };
+  },
+});
+```
+
+Then, we will update our type generator to recognize these data-specific file extensions. This should also raise errors when collections are misconfigured (i.e. `type: 'data'` is missing from the config file) and when a mix of content and data in the same collection is detected.
```
tony-sull, line 192: > This should also raise errors ... a mix of content and data in the same collection is detected.

I could see this being a bit of a blocker for the easy upgrade path of a unified `src/content` directory

If I have a large data collection and only need to upgrade one to use body content for some specific use case, I'd end up having to update every item in the collection to `.md` files

I can't think of a *great* example of this at the moment so it may very well be a corner case, feel free to ignore if it isn't a use case worth covering!

bholmesdev: Fair enough! There may be a case to allow content inside _data_ collections, since data collections have a subset of content collections' properties. Though it sounds like you're proposing the other way. In that case, properties like `render()` and `body` would be added to every collection item and stubbed out with `undefined`'s, which I worry about for usability. Also not sure about "mixing" return types based on the entry.

My thinking is this:
- Move forward with collection-level types for launch
- Consider allowing content collections in data collections as a fast follow
- Look into the "singleton" discussion here (https://github.com/withastro/roadmap/discussions/449) for one-off `.md` files as you're describing here. I think there's a use case to allowing one-off entries with a shared schema, though it's something we could tackle separately.

Let me know if that makes sense, or if I'm missing anythign!

bholmesdev:
Thanks for your contributions everyone! We're looking to reach consensus on this RFC by end of the week, so please share any final feedback (blocking or otherwise) that you want us to address. If we don't hear anything, that's our go-ahead to ship 🚢

proposals/0034-collection-references.md
bholmesdev: Discussed with docs and core offline. To avoid thrashing in our API, we decided _just_ making `reference()` a top-level import for launch makes the most sense. We can explore other import options by user demand.

proposals/0033-data-collections.md
bholmesdev: Yep, another case of "defensively make restrictive, open up if people want it" 👍

bholmesdev:
Well, I'll take your stunned silence as a "ship it!" Merging

PR #582, opened 2023-05-09
Specify a final comment period

matthewp:
Updates the language for how consensus works to match what we've been doing. Namely:

- Consensus is reached async during a final comment period.
- Rules here follow the governance voting rules.
- After the comment period the RFC is eligible for merging.


tony-sull (approved):
Yep, I'm a fan of the clarification here

Issue #529, opened 2023-03-22
Custom Client Directives


matthewp:
### Body

- Accepted Date: 2023-03-22
- Reference Issues/Discussions: 
  - https://github.com/withastro/roadmap/discussions/272
  - Legacy https://github.com/withastro/roadmap/pull/212
- Implementation PR: 

# Summary

Provide an API for integrations to implement custom client: directives to provide greater control for when client-side JS is loaded and executed.

# Background & Motivation

The last client directive added to core was the `client:only` directive in [August 2021](https://github.com/withastro/astro/issues/751). Since that time the core team has been hesitant to add new client directives despite the community asking about them.

Allowing custom client directives would both:

- Allow the community to experiment with different approaches to lazy-loading client JavaScript.
- Provide evidence, through telemetry data, on which directives are most used. This data could be used to determine if a directive should be brought into core.

Some examples of custom directives that people have wanted in the past:

- Loading JavaScript on client interactive, such as mouseover or click.
- Loading JavaScript when an element is visible, as opposed to within the viewport as `client:visible` currently does.
- The [Idle Until Urgent](https://philipwalton.com/articles/idle-until-urgent/) pattern which loads on either idle or interaction, whichever comes first.

# Goals

- Provide a way to customize loading of client components.
- Refactor the implementation of `client:` loading to get rid of the precompile step (this is a core repo refactor / improvement).
- Allow integrations to add their own directives.
- Allow integrations to provide type definitions for their new directives.

# Non-Goals 

- Allowing overriding builtin directives.
- Allowing for additional customization via new types of directives outside of `client:`.
- Allowing multiple directives to run at the same time.

# Future

- Solving the `client:only` load vs idle vs visible problem.

# Example

## Usage

To use a custom directive in your app, add an integration like any other.

```js
import { defineConfig } from 'astro/config';
import onClickDirective from '@matthewp/astro-click-directive';

export default defineConfig({
  integrations: [onClickDirective()]
});
```

And then use it in your Astro components:

```astro
---
import Counter from '../components/Counter.jsx';
---

<Counter client:click />
```

## Implementation

An implementer of a custom client directive would do so through the hooks API:

```js
export default function() {
  return {
    hooks: {
      'astro:config:setup': ({ addClientDirective }) => {
        addClientDirective({
          name: '@matthewp/astro-click-directive',
          key: 'click',
          entrypoint: '@matthewp/astro-click-directive/client.js'
        });
      },
    }
  }
}
```

[...]

PR #583, opened 2023-05-12
Custom Client Directives

bluwy:
# Summary

Support custom `client:` directives added by integrations

# Links
* [Full rendered proposal](https://github.com/withastro/roadmap/blob/custom-client-directives/proposals/custom-client-directives.md)
* [Stage 2 proposal](https://github.com/withastro/roadmap/issues/529)
* [Stage 1 proposal](https://github.com/withastro/roadmap/discussions/272)

### DIFF
--- a/proposals/custom-client-directives.md
+++ b/proposals/custom-client-directives.md
@@ -1,0 +1,111 @@
+- Start Date: 2023-05-12
+- Reference Issues/Discussions: 
+  - https://github.com/withastro/roadmap/discussions/272
+  - Legacy https://github.com/withastro/roadmap/pull/212
+- Implementation PR: https://github.com/withastro/astro/pull/7074
+
+# Summary
+
+Provide an API for integrations to implement custom `client:` directives to provide greater control for when client-side JS is loaded and executed.
+
+# Example
+
+```js
+import { defineConfig } from 'astro/config';
+import onClickDirective from '@matthewp/astro-click-directive';
+
+export default defineConfig({
+  integrations: [onClickDirective()]
+});
+```
+
+```js
+export default function onClickDirective() {
+  return {
+    hooks: {
+      'astro:config:setup': ({ addClientDirective }) => {
+        addClientDirective({
+          name: 'click',
+          entrypoint: fileUrlToPath(new URL('./click.js', import.meta.url))
+        });
+      },
+    }
+  }
+}
+```
+
+# Background & Motivation
+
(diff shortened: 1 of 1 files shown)


proposals/custom-client-directives.md
```diff
@@ -0,0 +1,127 @@
...
+
+An e2e test will be setup to make sure the client directive API works and loaded. Typings are a bit hard to test, so I'm doing it manually for now.
+
+# Drawbacks
+
+- Larger API surface area
+- Opens up partial hydration code pattern
+- Future builtin client directives are breaking changes
+- Third-party Astro libraries could rely on non-standard client directives
+
```
matthewp, line 116: The stage 1 proposal mentions file size as a possible drawback (someone could import all of React in a client directive). It would be good to mention that here even if we don't plan on doing anything about it.

bluwy: Updated

bluwy:
I added a new non-goal, which I thought could be a common misconception later:

> Non-goal:
> - Replay interaction on hydrate, e.g. if a `client:click` directive hydrates on clicking the button, Astro doesn't replay the click event to trigger some reaction after it hydrates. The user has to click the (now hydrated) button again to trigger a reaction.

Making this work ootb would be really tricky and I think it's better to leave the directives to handle this themselves if needed, since they should have a reference to the island element and `event.target` to do so.

bluwy:
I've thought about it before, but then we'd have to export a new entrypoint (`astro/directive`) and it's more tooling author-facing that I skipped it for now 😬 But if we're in favour of it, I won't mind supporting that though.

ematipico:
> I've thought about it before, but then we'd have to export a new entrypoint (astro/directive) and it's more tooling author-facing that I skipped it for now 😬 I guess a similar example is the image service API which could perhaps benefit with this pattern too.

You raise a good point. We can leave my suggestion for now, and think about it later! 

Maybe we could merge all these `define*` functions inside a single `astro/define` entry point. Just a suggestion

matthewp:
I'm moving for a call for consensus on this RFC. Custom client directives has been in experimental since 2.5 without incident. No issues have been filed thus far and no feedback on the API. This will be the final comment period (3 days); if there are no objections this will be merged and the feature can move out of experimental in a future release (likely 2.6).

matthewp:
It's been 3 days, this RFC is merged!

Issue #466, opened 2023-02-03
Redirects


natemoo-re:
### Body

- Accepted Date: 2023-02-03
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/319
- Implementation PR: 

# Summary

It would be useful if Astro could support redirects for static builds. Possibly for server builds, too?

Sometimes users want to rename a route or refactor how pages are organized. It's not possible to just rename a page as there may be links pointing to it, so redirects have to be generated. While this is possible to do manually, it's not very convenient.

# Background & Motivation

Redirects for `static` and `server` mode are fairly straight-forward [to implement](https://maxchadwick.xyz/blog/preserving-the-hash-and-query-string-with-jekyll-redirects). Other tools like [Jekyll](https://github.com/jekyll/jekyll-redirect-from) expose this ability, and hosting providers do as well. 

It would be helpful if Astro could generate these automatically as a convenience. This feature could also normalize the implementation across hosting providers and output mode.

# Goals

- Simplify boilerplate by providing a first-party API for redirects.
- Automatically generate markup for redirects in `static` mode
- Maybe: automatically configure redirects for adapters in `server` mode to avoid host-specific redirect configuration?

# Non-Goals 

None at this time.

# Example

API TBD, but this could either be a built-in `redirect` configuration

```
import { defineConfig } from 'astro/config';

export default defineConfig({
  redirects: {
    '/foo/bar': '/unicorn/rainbow'
  }
})
```

or exposed as an official integration

```
import { defineConfig } from 'astro/config';
import redirects from '@astrojs/redirects';

export default defineConfig({
  integrations: [
    redirects({
      '/foo/bar': '/unicorn/rainbow'
    })
  ]
})
```

pilcrowOnPaper:
It would be great if I can define a redirect response inside getStaticPaths, and for a redirect response inside frontmatter to just work like SSR mode.

tony-sull:
This could be a really interesting to host adapters!  The `@astrojs/netlify` integration could automatically build the [`_redirects`](https://docs.netlify.com/routing/redirects/#syntax-for-the-redirects-file) file

lilith:
I have a lot of websites that specify multiple redirects (`Aliases:`) in the frontmatter of each markdown page. Specifically, a web page/file can specify which source paths  should redirect to itself.

[...]

PR #587, opened 2023-05-22
Redirects RFC

matthewp:
# Summary

Provide a `redirects` config option for specifying redirected pages. Provide good integration with adapters.

# Links
* [Full rendered proposal](https://github.com/withastro/roadmap/blob/redirects/proposals/redirects.md)
* [Stage 2 proposal](https://github.com/withastro/roadmap/issues/466)
* [Stage 1 proposal](https://github.com/withastro/roadmap/discussions/319)

### DIFF
--- a/proposals/redirects.md
+++ b/proposals/redirects.md
@@ -106,6 +106,10 @@ export default defineConfig({
 
 In SSG mode this will call the destinations `getStaticPaths` method to get valid static paths. Those paths will be used to generate the HTML files for the redirects.
 
+## Routing priority
+
+Redirects will be given the same priority as file-system routing. This means that if there are conflicts, the same rules applies as with the file-system routes. For example, you could have a file system route `/src/pages/blog/contributing.astro` and a redirect route `/blog/[...slug]`. If these were both filesystem routes you would expect `contributing.astro` to be prioritized over the spread route. This is the case with redirects a well.
+
 ## Static generation
 
 Currently the static generation code throws for any non-200 response. With this change it will now accept any 3xx as a valid response codes. It will generate an HTML doc that looks like:
--- a/proposals/redirects.md
+++ b/proposals/redirects.md
@@ -32,6 +32,21 @@ export default defineConfig({
 });
 ```
 
+You can also specify the status code by using an object notation:
+
+```js
+import { defineConfig } from 'astro/config';
+
+export default defineConfig({
+  redirects: {
+    '/other': {
+      status: 302,
+      destination: '/place'
+    }
+  }
+});
+```
+
 # Background & Motivation
 
 This was original proposed as a stage 1 discussion [here](https://github.com/withastro/roadmap/discussions/319) and was one of the top voted proposals.
@@ -49,26 +64,51 @@ Having a redirects configuration within Astro itself allows a single place to de
 
 # Non-Goals
 
-- Specifying temporary redirects. These are best done at runtime, so SSR is a more appropriate way to do those.
-    - In the future we could expand this API to give more control over what status code is used, once use-cases are known.
+- Dynamic behavior that goes beyond the capabilities of the file-based routing system. So nothing based on the properties of the request, user session, etc. Regular routes still should be used for this scenario.
+- Redirects to pages that do not exist in the Astro project.
+- External redirects.
 
 # Detailed Design
 
 This will be implemented as a feature of the internal routing. The `RouteData` type will be extended to include:
 
 ```js
+interface RedirectConfig = string | {
+  status: 300 | 301 | 302 | 303 | 304 | 307 | 308;
+  destination: string;
+}
+
 export interface RouteData {
   type: 'redirect';
   // ...
-  redirect?: string;
+  redirect?: RedirectConfig;
+  redirectRoute?: RouteData;
 }
 ```
 
-Our core rendering handles routing and will detect this type of route and return a `Response` with a status code of `301` with the `Location` header set to the value of the route's `redirect` property.
+Our core rendering handles routing and will detect this type of route and return a `Response` with a status code in the 3xx range with the `Location` header set to the value of the route's `redirect` property.
+
+When using the object notation `{ destination: string; status: number; }`
+
+## Dynamic routes
+
+Dynamic routes are supported through the same syntax as in the file-based routing system. For example, if a site moved its blog it might set up a redirect like so:
+
+```js
+import { defineConfig } from 'astro/config';
+
+export default defineConfig({
+  redirects: {
+    '/blog/[...slug]': '/team/articles/[..slug]'
+  }
+});
+```
+
+In SSG mode this will call the destinations `getStaticPaths` method to get valid static paths. Those paths will be used to generate the HTML files for the redirects.
 
 ## Static generation
 
-Currently the static generation code throws for any non-200 response. With this change it will now accept `301` as a valid response code. It will generate an HTML doc that looks like:
+Currently the static generation code throws for any non-200 response. With this change it will now accept any 3xx as a valid response codes. It will generate an HTML doc that looks like:
 
 ```html
 `<!doctype html>
--- a/proposals/redirects.md
+++ b/proposals/redirects.md
@@ -1,0 +1,116 @@
+<!--
+  Note: You are probably looking for `stage-1--discussion-template.md`!
+  This template is reserved for anyone championing an already-approved proposal.
+
+  Community members who would like to propose an idea or feature should begin
+  by creating a GitHub Discussion. See the repo README.md for more info.
+
+  To use this template: create a new, empty file in the repo under `proposals/${ID}.md`.
+  Replace `${ID}` with the official accepted proposal ID, found in the GitHub Issue
+  of the accepted proposal.
+-->
+
+- Start Date: 2023-05-22
+- Reference Issues: https://github.com/withastro/roadmap/issues/466
+- Implementation PR: https://github.com/withastro/astro/pull/7067/
+
+# Summary
+
+Add a `redirects` config option to the Astro config which allows you to define redirects in a central place. Allow integrations to read this information and apply it to it's own redirects configuration for deployment.
+
+# Example
+
+New config option used like so:
+
+```js
+import { defineConfig } from 'astro/config';
+
+export default defineConfig({
+  redirects: {
+    '/other': '/place'
+  }
+});
+```
+
+# Background & Motivation
+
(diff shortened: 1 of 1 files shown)


proposals/redirects.md
```diff
@@ -0,0 +1,116 @@
...
+# Goals
+
+- Works in development mode.
+- Writes out `<meta http-equiv="refresh">` tags in static builds.
+- Gives proper hooks to integrations so that they can write to their own configuration.
+
+# Non-Goals
+
+- Specifying temporary redirects. These are best done at runtime, so SSR is a more appropriate way to do those.
+    - In the future we could expand this API to give more control over what status code is used, once use-cases are known.
```
bluwy, line 53: [3xx](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#redirection_messages) supports different sorts of redirects, so I'm not sure if we should limit this as a start. I think the usecase is common that some status codes makes more sense for their business logic.

Maybe for SSG we only support a subset of status codes?

proposals/redirects.md
```diff
@@ -0,0 +1,116 @@
...
+New config option used like so:
+
+```js
+import { defineConfig } from 'astro/config';
+
+export default defineConfig({
+  redirects: {
+    '/other': '/place'
+  }
+});
```
bluwy, line 32: Is there a way to redirect multiple pages at once? e.g. `/blog/*` -> `/articles/*`.

Also, platforms like Cloudflare supports [advanced types of redirects](https://developers.cloudflare.com/pages/platform/redirects/), and if they have to resort to using a custom `_redirects` file for it (which presumably you can't use both the file and `redirects` config at the same time), it puts the `redirects` option in an odd spot.

proposals/redirects.md
matthewp: That's fair. The way we would probably do that is:

```js
{
  '/one': {
    status: 302,
    destination: '/two'
  }
}
```

TypeScript and runtime would restrict to 3xx statuses.

Are you ok with keeping the string destination as a shortcut for 301 (which is the more common one for the use-cases).

proposals/redirects.md
matthewp: The intent is to support our file router's syntax, so you would do:

```js
'blog/[...params]': '/articles/[...params]'
```

I need to implement this before I write it into the RFC as there could be more to it.

proposals/redirects.md
```diff
@@ -0,0 +1,116 @@
...
+
+```js
+export interface RouteData {
+  type: 'redirect';
+  // ...
+  redirect?: string;
+}
+```
+
+Our core rendering handles routing and will detect this type of route and return a `Response` with a status code of `301` with the `Location` header set to the value of the route's `redirect` property.
```
ematipico, line 67: What about API endpoints? Will they be treated as 301 too? Or should we use [308](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308) to keep the method used?

proposals/redirects.md
bluwy: Yeah I'm fine with a `string` shortcut since we won't be merging the settings (which would be complex otherwise).

proposals/redirects.md
bluwy: That sounds like a neat idea! I'll resolve this first then.

proposals/redirects.md
matthewp: Updated to allow the object syntax and to show dynamic routes.

tony-sull:
Notes from May 23 community call:

- what takes priority when `src/pages`, `injectRoute`, and `redirects` collide?
    - currently `redirects` takes the highest priority
    - should routing priority follow the same rules with regards to rest routes?  i.e. will `src/pages/blog/contributing.astro` be overwritten by a redirect rule for `blog/[...slug]`?

proposals/redirects.md
delucis, line 1: Agree that we should figure out priority. It’s not clear to me what’s best. I’d probably land on user route > injectRoute > redirect. That means you can’t redirect away from an existing page — you _must_ remove the page in order to use a redirect — but that mirrors how platforms often work: Netlify for example ignores redirect config for pages that exist in the build output.

In fact, we should double check static output doesn’t interfere with redirects configured in the Netlify adapter. We might need to turn off the static meta tag output there. We have this issue in docs where because we have routes with meta redirects, we lose Netlify redirects, e.g. `/` route — `view-source:https://docs.astro.build/` — despite [having them configured](https://github.com/withastro/docs/blob/ef5a1a791f053c972a6966f072b3d350d1b7fd5e/netlify.toml#L90-L92).

delucis (commented):
Excited for this API — thanks for putting in the work to flesh it out @matthewp!

matthewp:
Note that this was just updated to explain routing priority. The intent is for it to have the same priority as filesystem routes. If there is an *exact* match then the redirect would win. But in the case of, for example, siblings with the redirect being a spread, the filesystem non-spread is prioritized, just as it would be if they were both FS based.

proposals/redirects.md
matthewp: I just updated the text to explain priority. I was able to make it so that it has the same priority as FS routes.

proposals/redirects.md
```diff
@@ -0,0 +1,156 @@
...
+In SSG mode this will call the destinations `getStaticPaths` method to get valid static paths. Those paths will be used to generate the HTML files for the redirects.
+
+## Static generation
+
+Currently the static generation code throws for any non-200 response. With this change it will now accept any 3xx as a valid response codes. It will generate an HTML doc that looks like:
+
+```html
+`<!doctype html>
+<title>OLD_LOCATION</title>
+<meta http-equiv="refresh" content="0;url=LOCATION" />
```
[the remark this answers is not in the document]
matthewp: Correct, we can't apply status codes here. So when deploying to a host without configuration you're always going to get a 200 response from these pages. So no effect on SEO, unfortunately. Yeah we can and should document that.

proposals/redirects.md
```diff
@@ -0,0 +1,156 @@
...
+- Redirects to pages that do not exist in the Astro project.
+- External redirects.
+
+# Detailed Design
+
+This will be implemented as a feature of the internal routing. The `RouteData` type will be extended to include:
+
+```js
+interface RedirectConfig = string | {
+  status: 300 | 301 | 302 | 303 | 304 | 307 | 308;
```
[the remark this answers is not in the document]
delucis: Looks like Cloudflare does this too: allows `200` for [“proxying”](https://developers.cloudflare.com/pages/platform/redirects/#proxying) and `404` for [“rewrites”](https://developers.cloudflare.com/pages/platform/redirects/#create-redirects). Pretty similar to Netlify (I forgot they support 404 too).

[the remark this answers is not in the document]
delucis: On Vercel redirects and rewrites are handled separately and looks like Netlify/Cloudflare’s `200` redirect is what Vercel calls a “rewrite”: https://vercel.com/docs/concepts/edge-network/rewrites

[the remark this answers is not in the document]
delucis: Not really familiar with Deno Deploy and don’t see anything obvious in their docs. I guess there it’s closer to our Node adapter? So up to us how exactly we handle stuff.

Looks like the static site platforms use a “200 redirect” to mean a server-side rewrite where the destination content gets served at the requested URL instead of notifying about a redirect. Probably useful for SPAs in particular.

proposals/redirects.md
delucis: Got it. Couple points there then:

- In the case that a redirect _and_ a file-system route of the same priority type exists (both spread or both not spread routes), which wins is not specified. I’d suggest the redirect loses as per platform behaviour mentioned above.

- We likely still need to have a way to avoid building the static `<meta>` routes if the adapter supports redirect config: Netlify will definitely ignore a config redirect in favour of a matching HTML file, and from reading [Vercel’s docs](https://vercel.com/docs/concepts/projects/project-configuration#rewrites), I suspect they do too: “precedence is given to the filesystem prior to rewrites being applied”

I can do some more research on that if you need.

matthewp: Let's take the point about `<meta>` to another comment, here: https://github.com/withastro/roadmap/pull/587#issuecomment-1563404860

matthewp:
`config.build.redirects = false` wouldn't be terrible either, I guess.

delucis:
Yeah a config flag adapters can toggle off doesn't seem too bad. A fourth option would be to just do this implicitly when an adapter is set on the basis that if there is an adapter, redirects are their responsibility. But I think a flag might be safest for the case where an adapter doesn't handle this.

matthewp:
There are likely adapters that do not have config based redirects, so I wouldn't want to blanket disable the feature on the presence of an adapter. I think I'll go with the config approach. Will update the RFC and implementation.

matthewp:
@delucis Updated to include the config: https://github.com/withastro/roadmap/pull/587/commits/dfd349eefc2251a532b306415ccc320568076e9d

delucis:
Thanks! LGTM 👍

I think the outstanding points from above are:

- Do we widen valid values for `status`? (e.g. to include `200` or `404` or just any number and it’s a users job to pick a good one 😅)

- If a redirect and a file-system route of the same priority type exists (i.e. both spread or both not-spread routes), which wins? I suggested the redirect loses.

matthewp:
- I don't think we should widen the value of `status`. The 200 example seems to be a special case for Netlify and not a widely supported thing. In that case it belongs in the project's `_redirects` file (which gets merged with redirects config). It wouldn't make sense to put this sort of thing into the config because it's not portable; the point of this feature is having a portable place to put redirects.
- Redirects losing sounds good to me, if this is what other systems do. I'll update the RFC to reflect that.

matthewp:
Updated to specify that FS routes win out when there are exact matches.

delucis:
> I don't think we should widen the value of `status`. The 200 example seems to be a special case for Netlify

I’m fine with this. I agree sticking to the common features to maintain cross-platform compatibility is a good idea. For the record, something like this is also supported by Cloudflare and Vercel (see https://github.com/withastro/roadmap/pull/587#discussion_r1205878490).

No more comments from me. Thanks for putting up with all my feedback! 😅

matthewp:
I'd be happy to open this up if we get feedback that people expect this. Will do some more research to see what the support matrix of these codes look like.

jlarmstrongiv:
We’re having trouble with the new experimental redirects in https://github.com/withastro/astro/issues/7322:

```js
{
  experimental: {
    redirects: true
  },
  redirects: {
    '/a': '/b'
  }
}
```

I think this syntax matches the RFC proposal

Issue #697, opened 2023-09-13
Support for partials


matthewp:
- Accepted Date: 2023-09-13
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/266
- Author: @matthewp 
- Champion(s): @matthewp
- Implementation PR: 

# Summary

Allow pages to be marked as being partials, preventing the doctype tag and any head injection from taking place. A partial is defined as a template which belongs in the context of another page and is dependent on that page in order to be fully functional.

# Background & Motivation

Partials are a technique that has been used by web applications for decades, popularized in frameworks such as Ruby on Rails. Frontend oriented JavaScript frameworks have typically not used partials, but instead use JSON APIs and front-end templating in order to dynamically change parts of the page. Nevertheless, partials have remained a niche feature, and with Astro's backend focus we have had interest in support for a long time.

Recently the popularity of projects like [htmx](https://htmx.org/) and [Unpoly](https://unpoly.com/) have revived interest in this technique. Since Astro treats each page request as a request for a full page it automatically attaches the doctype and head elements, making it difficult to simply inject into the page.

# Goals

- The ability to request a URL from Astro that excludes the usual page elements (doctype, head injection).
- Some way to optionally communicate extra dependencies, such as styles and scripts, so libraries that support head-diffing can achieve that.
- The base feature should work the same in SSG and SSR apps. Partials are still output as `.html` files in a static build.

# Non-Goals 

- This isn't an integration specifically for HTMX or any one library. It should work with any library or manual DOM manipulation that does `innerHTML`.
  - We may take into account special headers from popular libraries if there is a way to enhance the base experience.
- No client-side scripts from Astro will be part of this change.
- Routing should mostly be unchanged; this is metadata about a page but doesn't change the location or where routes live (still in the `pages` folder).

lilnasy:
I would like to bring into the conversation Github.com's web-components-based framework, [catalyst](https://github.com/github/catalyst).

The relevant aspect is that associated javascript and css would still need to be fetched and duplicated.

webpro:
Makes me think of RSS feed. Could this maybe also be used/hooked into when rendering RSS content? So there would be no need for a custom Markdown parser/renderer (e.g. `markdown-it`).

ayoayco:
This is amazing. Would unlock so many use cases for .astro files as API endpoints.

aFuzzyBear:
a bear can only dream....

philipschm1tt:
[...]

PR #721, opened 2023-10-03
Partials RFC

matthewp:
# Summary

Add the ability to configure a page in the `pages` directory to be treated as a *fragment*. A fragment is a chunk of HTML that is not interpreted to be a page, so no head content or `doctype` are included.

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/fragments/proposals/partials.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/697)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/266)

### DIFF
--- a/proposals/fragments.md
+++ b/proposals/fragments.md
@@ -1,0 +1,103 @@
+- Start Date: 2023-10-03
+- Reference Issues: https://github.com/withastro/roadmap/issues/697
+- Implementation PR: <!-- leave empty -->
+
+# Summary
+
+Provide a configuration flag for page components to opt-in to *fragment* behavior, preventing head injection of scripts and styles, and the doctype.
+
+# Example
+
+In any component inside of the __pages__ directory set the `fragment` option:
+
+```astro
+---
+export const fragment = true;
+---
+
+<div>This is a fragment!</div>
+```
+
+# Background & Motivation
+
+Partials are a technique that has been used by web applications for decades, popularized in frameworks such as Ruby on Rails. Frontend oriented JavaScript frameworks have typically not used partials, but instead use JSON APIs and front-end templating in order to dynamically change parts of the page. Nevertheless, partials have remained a niche feature, and with Astro's backend focus we have had interest in support for a long time.
+
+Recently the popularity of projects like [htmx](https://htmx.org/) and [Unpoly](https://unpoly.com/) have revived interest in this technique. Since Astro treats each page request as a request for a full page it automatically attaches the doctype and head elements, making it difficult to simply inject into the page.
+
+# Goals
+
+- The ability to request a URL from Astro that excludes the usual page elements (doctype, head injection).
+- The base feature should work the same in SSG and SSR apps. Partials are still output as `.html` files in a static build.
+
+# Non-Goals 
+
+- This isn't an integration specifically for HTMX or any one library. It should work with any library or manual DOM manipulation that does `innerHTML`.
+- No client-side scripts from Astro will be part of this change.
+- Support for integrations or middleware at this time. It could be possible to allow middleware to communicate that a request is for a partial, but that is not part of this RFC.
+
+# Detailed Design
+
+Fragments are opted into on a per-page basis. Any page within the `pages` directory can become a fragment through this config:
+
+```astro
+---
+export const fragment = true;
+---
+```
+
+This value must be either:
+
+- A literal boolean of `true` or `false` (there's no reason to every use false).
+- A configuration value from `import.meta` such as:
+
+    ```astro
+    ---
+    export const fragment = import.meta.env.USE_FRAGMENTS;
+    ---
+    ```
+
+The value *must* be identified statically. This means that a page can't be both a full page and a fragment. If you want to share the same logic and template for a partial and fragment you can do so by putting the common code into a component.
+
+## Implementation
+
+This is a very small change. Internally Astro uses an object known as a `result` that stores intoformation about a request. That result object will add a `fragment` property:
+
+```ts
+interface AstroResult {
+  fragment: boolean;
+}
+```
+
+When rendering the fragment value is taken from the __component module__ and placed on the result object.
+
+This boolean is then used in two places:
+
+- In head injection it will be used to prevent head injection.
+- In page rendering it will be used to prevent doctype prepending.
+
+These are the only changes needed.
+
+# Testing Strategy
+
+We want to verify that fragments do not include head content with tests for:
+
+- A component that contains `<style>` and `<script>` elements.
+- The `doctype`
+
+Additionally we should test both dev mode and the static build, to ensure that the files are still written to disk.
+
+# Drawbacks
+
+- It could be unexpected that a component styles are not included.
+- Fragments need to be styled with global styles. This could mean using `<style is:global>`, or imported CSS, or an atomic CSS library like Tailwind.
+- Use-cases where you want both a page and a fragment in the same file are not possible.
+
+# Alternatives
+
+The other major alternative considered was a file-naming convention such as `todo.fragment.astro`. The `.fragment` would be used to know that the page is for a fragment. This would add extra work to routing, where as the `export const fragment` solution contains the problem to the runtime and build plugin (which already exists).
+
+# Adoption strategy
+
+- This will be introduced as an experimental feature to ensure that it's what developers are after.
+- This would not be possible to do as a 3rd party package or integration so it must be done in core.
+- Based on feedback we might make adjustments to the RFC before it's release.
--- a/proposals/partials.md
+++ b/proposals/partials.md
@@ -4,18 +4,18 @@
 
 # Summary
 
-Provide a configuration flag for page components to opt-in to *fragment* behavior, preventing head injection of scripts and styles, and the doctype.
+Provide a configuration flag for page components to opt-in to *partial* behavior, preventing head injection of scripts and styles, and the doctype.
 
 # Example
 
-In any component inside of the __pages__ directory set the `fragment` option:
+In any component inside of the __pages__ directory set the `partial` option:
 
 ```astro
 ---
-export const fragment = true;
+export const partial = true;
 ---
 
-<div>This is a fragment!</div>
+<div>This is a partial!</div>
 ```
 
 # Background & Motivation
@@ -37,11 +37,11 @@ Recently the popularity of projects like [htmx](https://htmx.org/) and [Unpoly](
 
 # Detailed Design
 
-Fragments are opted into on a per-page basis. Any page within the `pages` directory can become a fragment through this config:
+partials are opted into on a per-page basis. Any page within the `pages` directory can become a partial through this config:
 
 ```astro
 ---
-export const fragment = true;
+export const partial = true;
 ---
 ```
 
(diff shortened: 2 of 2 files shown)


ZebraFlesh:
> No client-side scripts from Astro will be part of this change.

Can you elaborate on the meaning of this line from the Non-Goals section? My use case is that I want to include client side scripts defined by my fragments, and I'm unclear on if this will/won't be supported.

philipschm1tt:
I am interested in fragments coming from the micro-frontend perspective: include a fragment either server-side/edge-side via SSI/ESI or include a fragment client-side via Ajax. I started a roadmap discussion about these use cases at https://github.com/withastro/roadmap/discussions/713
Note: I don't expect these use cases to be covered with this planned partials feature but I think it's desirable to support these use cases in the future and thinking about that future can help inform the current first step towards partials.

The proposal solves some use cases: notably the HTMX/Hotwire-style approach where the server updates the markup for parts of the page.

I want to think one step further about potential future micro-frontend Ajax support.
With Ajax, the "current" page might be rendered by any technologies geared for server-side rendering. Maybe a classic Java + Spring MVC page that uses a Thymeleaf template. Then you might have a button on that page to, for example, get your notification or look into your cart. That could be done via Ajax: the client-side JavaScript could request that fragment from a separate Astro server. The Astro server would then need to deliver not just the markup but also the styles and some JavaScript for that fragment. I believe those stylesheets and scripts are commonly then not added to the head of the page but should work by including the tags "inline" where the fragment is added.

If Astro were to support – in the future – that use case of delivering a fragment to a page rendered by another server, it would need to support styles and scripts for the fragment.
Would you then add another flag like `export const fragmentWithScriptsAndStyles` that is false by default? Or would you include a second flag like `export const fragmentMarkupOnly` that is true by default? Or would you use two independent flags like `export const fragmentMarkupOnly` and `export const fragmentWithStyleAndScript`? 
If you would use two independent flags, the current proposal of just `export const fragment` could become ambiguous and need some work.

matthewp:
@ZebraFlesh No, all I mean by that is that Astro is not opinionated about *how* you load a fragment in the client and are not getting involved in that. So Astro is not going to be adding scripts *of its own*. You can of course use your own scripts to manage fragment insertion.

matthewp:
@philipschm1tt I understand, what i'm saying is that the thing you are describing is *not* a fragment (in the definition of this RFC). It's just a regular page. You might *think* of it as a fragment, from the perspective of how you're using it within another framework, but Astro's perspective a page which includes styles is not a fragment.

Does that make sense? What I'm saying is, you should be able to create a page with just the content in your example and it will include styles, as you have shown. Without needing this feature.

philipschm1tt:
> @philipschm1tt I understand, what i'm saying is that the thing you are describing is _not_ a fragment (in the definition of this RFC). It's just a regular page. You might _think_ of it as a fragment, from the perspective of how you're using it within another framework, but Astro's perspective a page which includes styles is not a fragment.
> 
> Does that make sense? What I'm saying is, you should be able to create a page with just the content in your example and it will include styles, as you have shown. Without needing this feature.

Yes, I understand what use case(s) you want to solve with this proposal.
But multiple people in the original discussion https://github.com/withastro/roadmap/discussions/266 had the micro-frontend use case. What bothered them (and me), was the parts of the page that only work at the page level and not as part of a fragment.

- One Server-Side Include comment: https://github.com/withastro/roadmap/discussions/266#discussioncomment-3874731
- One Edge-Side Include comment: https://github.com/withastro/roadmap/discussions/266#discussioncomment-4942089

These server-side integration use cases (like the Ajax client-side integration approach), still need styles and behavior as part of these kinds of fragment. The fragments in these contexts come from different servers.

matthewp:
@philipschm1tt Can you explain why the existing pages folder behavior does not work for your use-case?

philipschm1tt:
> @philipschm1tt Can you explain why the existing pages folder behavior does not work for your use-case?

I have not personally explored this as much as the commenters in https://github.com/withastro/roadmap/discussions/266 and some of my colleagues.

What I'm looking for is a response like 
```
<link href="fragment.css" rel="stylesheet" />
<script defer src="fragment.js"></script>
<h2>Recommendations</h2>
<div class="recommendations">
  ...
</div>
```

So roughly the same behavior as for a page but without doctype and with the CSS/JS links for that "page" outside of a head tag.

And unlike some of the commenters on that issue, I would like to use SSR and could not use some of the post-processing workarounds that were suggested for static files.

matthewp:
@philipschm1tt If you don't include a `<head>` tag then there will be no head tag, but there will be the scripts and styles at the top, just as you have written.

The only part that is different is that there *will* also be a doctype, which you say that you do not want. 

I would then ask, how are you dealing with these scripts and styles? Are you not extracting them out and putting them into the head? If you do not, you will get FOUC.

ZebraFlesh:
I am inlining those styles for the microfrontend use case. Today, I do the following:

1. manually inline CSS
1. remove DOCTYPE
1. remove head tag (but retain contents)
1. remove body tag (but retain contents)

And now I have a fragment. (I could skip the inlining, but since I'm doing ESI processing at edge on a streaming response, I can't "go back" and put the CSS in the head tag, because that content has already been streamed to the client.)

My hope for this feature was removing the last 3 steps.

matthewp:
Astro does not add a `head` or `body` tag. So those steps you already can skip by simply not adding those tags yourself.

By (1) do you mean you take the link tags from the HTML and move them into the head?

matthewp:
This feature can be tried out by installing the preview release:

```shell
npm install astro@next--fragments
```

philipschm1tt:
> I would then ask, how are you dealing with these scripts and styles? Are you not extracting them out and putting them into the head? If you do not, you will get FOUC.

For the Ajax case the fragment is requested client-side on user interaction, so you won't have the styles and scripts during page load.

For the server-side integration case, this may depend on the specific approach you use.
I'm not aware of a "built-in" way with SSI or ESI to move such links to the head. I don't know if some people might include a "fragment" that includes only these links to the head (at the cost of an additional request). I guess with SSI/ESI you may typically either accept some FOUC or use some workarounds to hide content until CSS is loaded (and then trigger the reveal with JS).

I think this is probably a classic SSI/ESI challenge with some options and trade-offs.
It may not be a big deal since the main part of the page should usually come from the server that adds its styles/scripts to the head. And it may benefit from caching the CSS from the second request on.

ZebraFlesh:
> Astro does not add a `head` or `body` tag. So those steps you already can skip by simply not adding those tags yourself.

Excellent point. It's been a while since I've looked at this problem, so I was quoting from out post-processing script, but this explanation makes sense.
 
> By (1) do you mean you take the link tags from the HTML and move them into the head?

No, I mean we find each `<link rel="stylesheet">`, read the indicated file contents, and replace the `<link rel="stylesheet">` with `<style>{contents_of_file}</style>`.

matthewp:
@MrHBS definitely open to doing that. Should probably run a pole some where.

matthewp:
I've renamed it to Partials in this RFC. I'm working to also update the code. Additionally I've added a section to the RFC describing why the proposal doesn't include scripts/styles: https://github.com/withastro/roadmap/blob/fragments/proposals/partials.md#including-head-content-but-not-doctype

ematipico (commented):
It would be great if we could explain the result with an example. While it's explained that other tools have used partials, some people wouldn't know **how** they should be used.

I would like a few lines that say if a page opt-ins `partial = true`, what's the result and what the user should expect from that page?

matthewp:
@ematipico Added that in 47ab4ba772319a6453429fae40fcb7670861896e, let me know if that's what you were looking for!

ematipico (approved):
Thank you, @matthewp, for addressing my comment. Happy with the RFC!

matthewp:
I'm moving for a call for consensus on this RFC. This will be the final comment period (3 days); if there are no objections this will be merged and the feature can be released in a future release (likely 3.4)

matthewp:
@silent1mezzo I will probably use a `src/pages/partials` convention in my own apps, but I don't think Astro should enforce that. For a few reasons:

- Astro is mostly unopinionated about where you put your files. Instead we lean on conventions; for ex. it's common for layouts to be in `src/layouts`, but they don't have to be! Layouts are really just normal components. So it's not normal for us to have a folder convention *for organizational reasons*. If there was a technical reason for it, we would do it.
- Adding a `src/pages/partials` known folder would be a breaking change, as people might currently have pages in such a folder.
- If you didn't know of this feature, and added a page to `src/pages/partials` it would be confusing to see that the page did not include styles+scripts. Since this is a niche feature, I don't think having people use it by accident is a good thing to do.
- With the current design, integrations can add partials; they would not be able to do so with a folder or file naming requirement.

matthewp:
> Won't adding the `partials` variable also be breaking since someone could have that already?

No, exporting variables is not supported by Astro component outside of the documented ones (`getStaticPaths`, `prerender`), so we can add new variables as needed.

> Hmm, I actually disagree with this. I think the variable is more confusing than a folder. At best they're the same and you need to go to the docs to figure it out. At least the folder has examples from other frameworks people could lean on.

A better explanation for my point here is that in Astro you can create any URL structure you want, and if we give special meaning to pages/partials that means that you *cannot* create a page with the URL `/partials/whatever`.

Conversely we would be enforcing a URL structure for partials themselves. Some people might want to do RESTful URLs where `/todos` is a partial response with the list of todos. I don't think it's Astro's place to tell people what their URLs can be.

silent1mezzo:
Makes sense

lloydjatkinson:
Just wanted to comment I find the name of the feature quite confusing (though I understand how the name was arrived at). Partials is already a fairly well used terminology to describe reusable templates or "views" particularly in MVC-like frameworks. These partials have little in common with this RFC. Essentially partials in this context are components.

[Partial views in ASP.NET Core](https://learn.microsoft.com/en-us/aspnet/core/mvc/views/partial?view=aspnetcore-7.0)
[NodeJs Express EJS Layouts and Partials](https://raddy.dev/blog/nodejs-express-layouts-and-partials/)
[How to use Partials in Rails](https://gorails.com/blog/how-to-use-partials-in-rails)

As the unit of abstraction in Astro is the component, and as I understand it the motivation for this RFC is for endpoints, perhaps *Fragment Pages* is a better feature name?

PeterDraex:
The section [Including head content but not doctype](https://github.com/withastro/roadmap/blob/fragments/proposals/partials.md#including-head-content-but-not-doctype) lists a cons of injecting scripts/styles directly mid-page and concludes that it's never the right approach to include scripts/styles in the partial.

In my case, I am working with an old PHP site with spaghetti code, which I'd like to gradually transition to Astro. During the transition, my plan is to have the Astro partials injected in the HTML. Developing code to manage both the PHP assets and Astro assets together is somewhere between not-worth-it and impossible. On the other hand, I can make sure I develop the Astro components in a manner that repeated injection of styles/scripts would not be an issue. If this RFC goes ahead, I'll have to keep setting `partial = false` and stripping doctype. Unless you guys see any better way to do the transition than server-side injection of Astro code?


As there are many use cases for partials, it might be good to have the option to configure separately:
- if doctype should be stripped
- if scripts/styles should be stripped

What do you think?

matthewp:
@PeterDraex I'd recommend stripping out the doctype using middleware for your case. It should be very simple to write. We don't want a feature that immediate breaks down once you use a partial more than once.

matthewp:
@lloydjatkinson thanks for the reference! We original did call it fragments but switched to partials after some discussion. Your points are very valid! Going to discuss it with the core team and see what they think.

philipschm1tt:
>Just wanted to comment I find the name of the feature unclear (though I understand how the name was arrived at). Partials is already a fairly well used terminology to describe reusable templates or "views" particularly in MVC-like frameworks. These partials have little in common with this RFC. Essentially partials in this context are components.

> @lloydjatkinson thanks for the reference! We original did call it fragments but switched to partials after some discussion. Your points are very valid! Going to discuss it with the core team and see what they think.

I would prefer "partials" over "fragments" in this case.

"Partials" is not uncommon (also used for things like SCSS) and not strictly about re-using – just says that it's only a part of the markup. That usage is not far off from some of the linked examples: the MVC frameworks often amount to parts of only the markup – the content of an Astro partial and something like a Java JSP "partial" would not be big.

By contrast, "fragment" is increasingly popular in the micro-frontend context (see https://engineering.zalando.com/posts/2018/12/front-end-micro-services.html or https://blog.cloudflare.com/better-micro-frontends/ ). Astro does not support that approach. The word refers more to the "outcome" of a piece of the rendered site, which also "implies" the behavior within that "fragment" and not just the markup.

An alternative term that would work: "html fragment".
It's precise: only HTML. Seems fine to me, too.

matthewp:
We are going to stick with partials, but title it as __Page Partials__ in our documentation and communication, to specify that this is about partials that are accessible via a URL. Naming is hard and there's no perfect name. In the future if this seems like a mistake we can always rename, so this is not a blocking sort of decision.

-----

This RFC has gone through the final comment period, I appreciate everyone's feedback!

Issue #701, opened 2023-09-14
i18n routing


ematipico:
- Accepted Date: 2022-09-14
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/600
- Author: @ematipico 
- Champion(s): @ematipico 
- Implementation PR: 

# Summary

First-class support for localized routes, aka i18n routing.

# Background & Motivation

Many websites need to ship support for translated/localised websites for many reasons:
- legal;
- localised content;
- localised market;
- etc.

Nowadays there are [workarounds](https://docs.astro.build/en/recipes/i18n/) in Astro to make it work, although these workarounds have limitations, and because of that many users can't ship their website properly, or they have to work more to compensate the lack of first-class support.

# Goals

- Localised routes with locale prefixes;
- Default locale with no prefix;
- Redirect to default locale if prefix enabled;
- Localise injected routes;
- Domain support, with the help of [Adapter features](https://docs.astro.build/en/reference/adapter-reference/#adapter-features), so this will be bound to the limitations of the hosting provider;
- Provide the necessary APIs for integrations and libraries to request information about the current locales;
- Locale detection via the `Accept-Language` header, so support SSR;
- Provide first-class APIs to users to work around locales (`.astro` components, endpoints, middleware);

# Non-Goals 

- Localised data (dates, numbers, plurals, et.c);
- Dictionaries where users can store translations of pre-defined words;
- SEO optimisations;

SudoCat:
Does this task have someone working on it, or would contributions be welcome?

ematipico:
> Does this task have someone working on it, or would contributions be welcome?

I'll champion it; this means I will manage the project and work on it full-time, but that doesn't mean that third-party contributions aren't welcome. This means that the project will have a boost with a defined release date. You're more than welcome to contribute; we can use Github/Discord as a coordination medium, and I will help with PR reviews and merge them in case there's some delay. Any help is appreciated!

sasoria:
Awesome! This would make my Astro app a lot simpler 🙌

SudoCat:
> > Does this task have someone working on it, or would contributions be welcome?
> 
> I'll champion it; this means I will manage the project and work on it full-time, but that doesn't mean that third-party contributions aren't welcome. This means that the project will have a boost with a defined release date. You're more than welcome to contribute; we can use Github/Discord as a coordination medium, and I will help with PR reviews and merge them in case there's some delay. Any help is appreciated!

[...]

PR #734, opened 2023-10-19
i18n routing RFC

ematipico:
# Summary

i18n routing support baked into Astro.

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/i18n-routing/proposals/0041-i18n-routing.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/701)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/600)

### DIFF
--- a/proposals/0041-i18n-routing.md
+++ b/proposals/0041-i18n-routing.md
@@ -311,7 +311,7 @@ This list is sorted by the highest to the lowest using the [quality value](https
 This information is available through the global object `Astro`: 
 - `Astro.preferredLocaleList: string[] | undefined`
     
-  For example, if `i18n.locals` contains `['pt', 'fr', 'de']`, and the value of the `Accept-Header` value is `en, fr;q=0.2, de;q=0.8, *;q=0.5`, then 
+  For example, if `i18n.locales` contains `['pt', 'fr', 'de']`, and the value of the `Accept-Header` value is `en, fr;q=0.2, de;q=0.8, *;q=0.5`, then 
   `Astro.preferredLocaleList` will be `['de', 'fr']` because `pt` isn't inside the header, and `en` isn't supported by the website. `de` comes first because it has a highest quality value.
   
   The property might be `undefined` if the developer isn't using their site in SSR. When `Accept-Header` is `*`, the list contained in `i18n.locales` is returned. `*` means that no preferences have been set, so all the original locales are supported and preferred. 
--- a/proposals/0041-i18n-routing.md
+++ b/proposals/0041-i18n-routing.md
@@ -271,24 +271,23 @@ The options allow to customise the behaviour of the APIs:
 
 ### Routing strategy
 
-An option called `routingStrategy` that allows to change the behaviour of the routing. The option accepts the following values:
+An option called `routing` that allows to change the behaviour of the routing. This option is an object that accept the following fields:
 
 > **Important**:
 >
 > The routing strategies are only applied to pages. Endpoints and redirects are exonerated.
 
-
-- `prefix-always`: all URLs of the website must have a locale prefix. Astro will return a 404 for any route that doesn't fulfill the requirements.
+- `routing.prefixDefaultLocale`:
+  When `false`, all URLs of the website must have a locale prefix. Astro will return a 404 for any route that doesn't fulfill the requirements.
   Use `example.com/[lang]/content/` for every locale.
   The index `example.com/` will **redirect** to `example.com/<defaultLocale>`.
 
-- `prefix-other-locales`: the URLs of the default locale must not have a prefix, while the rest of locales must have a locale prefix. 
+  When `true`, the URLs of the default locale must not have a prefix, while the rest of locales must have a locale prefix. 
   Use `example.com/content/` for the default locale. Use `example.com/[lang]/content/` for other locales. 
   Trying to access to use `example.com/[defaultLocale]/content/` will result into a 404.
 
-- `domain`: SSR only, it enables support for different domains. When a locale is mapped to domain, all the URLs won't have the language prefix.
-  You map `fr` to `fr.example.com`, if you want a to have a blog page to look like `fr.example.com/blog` instead of `example.com/fr/blog`.
-  The localised folders be must in the `src/pages/` folder.
+- `routing.strategy`: tells Astro where the locales should handled inside a URL 
+  - "pathname": the locales are expected to be in the `pathname` of a URL, meaning after the domain.
 
 ### Fallback system
 
--- a/proposals/0041-i18n-routing.md
+++ b/proposals/0041-i18n-routing.md
@@ -103,7 +103,8 @@ Below the list of additional features that Astro will be provided with the i18n 
 
 ### A new virtual module called `astro:i18n`
 
-> [!NOTE]
+> **Note**:
+> 
 > This feature doesn't require the adapter help
 
 A virtual module called `astro:i18n` will be available to retrieve important information useful to frontend and backed.
@@ -199,12 +200,14 @@ The options allow to customise the behaviour of the APIs:
 
 An option called `routingStrategy` that allows to change the behaviour of the routing. The option accepts the following values:
 
+> **Important**:
+>
+> The routing strategies are only applied to pages. Endpoints and redirects are exonerated.
+
+
 - `prefix-always`: all URLs of the website must have a locale prefix. Astro will return a 404 for any route that doesn't fulfill the requirements.
   Use `example.com/[lang]/content/` for every locale.
   The index `example.com/` will **redirect** to `example.com/<defaultLocale>`.
-  
-  > ![!IMPORTANT]
-  > The routing strategy applies its logic only to pages. Endpoints are excluded by default.
 
 - `prefix-other-locales`: the URLs of the default locale must not have a prefix, while the rest of locales must have a locale prefix. 
   Use `example.com/content/` for the default locale. Use `example.com/[lang]/content/` for other locales. 
@@ -244,7 +247,8 @@ This information is available through the global object `Astro`:
 
   The property might be `undefined` if the developer isn't using their site in SSR, or the highest value of `Accept-Header` is `*`. 
 
-> [!NOTE]
+> **Note**:
+> 
 > This feature is only available in **SSR**
 
 ### `Astro.currentLocale: string | undefined`
@@ -255,7 +259,8 @@ It's `undefined` if the URL doesn't contain a locale that is defined in `i18n.lo
 
 ### Domain support
 
-> [!NOTE]
+> **Note**:
+> 
 > This feature requires adapter support using an Astro feature
 
 A feature that allows to support different domains for certain locales.
--- a/proposals/0041-i18n-routing.md
+++ b/proposals/0041-i18n-routing.md
@@ -128,10 +128,7 @@ export default defineConfig({
     base: '/docs',
     i18n: {
         defaultLocaLe: 'en',
-        locales: ['en', 'es', 'pt', 'fr'],
-        domains: {
-            pt: "https://example.pt"
-        }
+        locales: ['en', 'es', 'pt', 'fr']
     }
 })
 ```
@@ -139,7 +136,7 @@ export default defineConfig({
 ```astro
 ---
 // src/pages/index.astro
-import { getAbsoluteLocaleUrl } from "astro:18";
+import { getRelativeLocaleUrl } from "astro:18";
 console.log(getRelativeLocaleUrl('es')) // will log "/docs/es"
 ---
 ```
@@ -152,7 +149,7 @@ Given a locale, the function will return the **absolute** URL, taking into accou
 ---
 // src/pages/index.astro
 import { getAbsoluteLocaleUrl } from "astro:18";
-console.log(getRelativeLocaleUrl('es')) // will log "http://localhost:4321/es"
+console.log(getAbsoluteLocaleUrl('es')) // will log "http://localhost:4321/es"
 ---
 ```
 
@@ -176,7 +173,7 @@ export default defineConfig({
 ---
 // src/pages/index.astro
 import { getAbsoluteLocaleUrl } from "astro:18";
-console.log(getRelativeLocaleUrl('pt')) // will log "https://example.pt/"
+console.log(getAbsoluteLocaleUrl('pt')) // will log "https://example.pt/"
 ---
 ```
 
@@ -196,6 +193,9 @@ The option `fallbackControl` is a string that can have two values:
 
 - `none`, the **default** Astro should not do anything, and a 404 is rendered;
 - [`redirect`](#redirect), Astro should use a redirect when re-routing to the destination locale; when `defaultLocale` is set, Astro will create a redirect to `defaultLocale` 
+
+The logic is triggered **only** when the path `/` is visited.
+
 
 ### Fallback system
 
@@ -256,29 +256,6 @@ The logic is triggered **only** when the path `/` is visited. When the logic is 
 - the `/<LOCALE>` path, where `<LOCALE>` is the locale matched in `locales` list;
 - the [sub-domain/domain](#domain-support) of `<LOCALE>`, if there's a match;
 
-The adapter needs to have the capabilities to read the `Accept-Language` cookie. An adapter can signal the support of this feature using the new Astro feature:
-
-```js
-export default function createIntegration() {
-  return {
-    name: '@ema/my-adapter',
-    hooks: {
-      'astro:config:done': ({ setAdapter }) => {
-        setAdapter({
-          name: '@ema/my-adapter',
-          serverEntrypoint: '@ema/my-adapter/server.js',
-          supportedAstroFeatures: {
-            i81n: {
-              languageDetection: "experimental",
-            }
-          }
-        });
-      },
-    },
-  };
-}
-```
-
 ### Domain support
 
 > [!NOTE]
@@ -295,7 +272,7 @@ export default defineConfig({
     i18n: {
         defaultLocaLe: 'en',
         locales: ['en', 'es', 'pt_BR', 'pt', 'fr'],
-        domain: {
+        domains: {
             fr: "fr.example.com",
             pt: "example.pt"
         }
@@ -303,17 +280,11 @@ export default defineConfig({
 })
 ```
 
-The following APIs will behave as follow:
+The following APIs will behave as follows:
 - [`getRelativeLocaleUrl`](#getrelativelocaleurllocale-string-string): it won't prefix the locale to the URL. From `/en` to `/`;
 - [`getAbsoluteLocaleUrl`](#getabsolutelocaleurllocale-string-string): it won't have the locale in the URL: From `example.com/fr` to `fr.example.com/`;
 
 Adapters must have the capabilities to redirect a user from one domain to another based on the domains configured.
-
-For example, using the Netlify redirects:
-
-```
-/fr/* /fr.example.com 200
-```
 
 An adapter can signal Astro the feature support using the relative configuration:
 
@@ -391,18 +362,5 @@ Once all the features are deemed stable, the whole i18n routing will be out from
 
 # Unresolved Questions
 
-Optional, but suggested for first drafts.
-What parts of the design are still to be determined?
-
-- language detection
-- domains
-
-For those features that require adapter support (language detection, domains), it's possible that adapters are going to be shipped **after** the feature is implemented in core. And they are, they will be shipped with an `experimental` support until they are stable.
-
-Once all the features are deemed stable, the whole i18n routing will be out from the experimental phase.
-
-
-# Unresolved Questions
-
-Optional, but suggested for first drafts.
-What parts of the design are still to be determined?
+- We are looking at a way to type the APIs of `astro:i18n`, but we don't know if we have the infrastructure to do so;
+- Do we need a configuration to tell Astro **where** the locale folders are? Or, should we enforce that somehow (root folder)?
--- a/proposals/0041-i18n-routing.md
+++ b/proposals/0041-i18n-routing.md
@@ -213,6 +213,10 @@ An option called `routingStrategy` that allows to change the behaviour of the ro
   Use `example.com/content/` for the default locale. Use `example.com/[lang]/content/` for other locales. 
   Trying to access to use `example.com/[defaultLocale]/content/` will result into a 404.
 
+- `domain`: SSR only, it enables support for different domains. When a locale is mapped to domain, all the URLs won't have the language prefix.
+  You map `fr` to `fr.example.com`, if you want a to have a blog page to look like `fr.example.com/blog` instead of `example.com/fr/blog`.
+  The localised folders be must in the `src/pages/` folder.
+
 ### Fallback system
 
 The fallback system is a feature that allows users to re-route users from one locale to another in case a page is missing.
@@ -275,9 +279,10 @@ export default defineConfig({
         defaultLocaLe: 'en',
         locales: ['en', 'es', 'pt_BR', 'pt', 'fr'],
         domains: {
-            fr: "fr.example.com",
-            pt: "example.pt"
-        }
+            fr: "https://fr.example.com",
+            pt: "https://example.pt"
+        },
+        routingStrategy: "domain"
     }
 })
 ```
--- a/proposals/0041-i18n-routing.md
+++ b/proposals/0041-i18n-routing.md
@@ -85,7 +85,7 @@ If a user has a `middleware.ts` file when the i18n routing is enabled, Astro wil
 
 ```js
 pipeline.setMiddlewareFunction(
-    sequence(userMiddleware.onRequest, createI18nMiddleware(config))
+    sequence(createI18nMiddleware(config), userMiddleware.onRequest)
 )
 ```
 
@@ -192,21 +192,17 @@ Same as `getAbsoluteLocaleUrl`, but it will return all the locales supported.
 
 The options allow to customise the behaviour of the APIs:
 
-- `path?: string`: a path that to append to `locale`
 - `prependWith?: string`: a path to prepend to `locale`
 - `normalizeLocale?: boolean`: when `true`, the locale is transformed in lower case and the underscore (`_`) is replaced with dash (`-`) 
 
-### Fallback control
+### Routing strategy
 
-An option that tells Astro what to do in case it doesn't find a page that belongs to one of the `locales` configured.
+An option called `routingStrategy` that allows to change the behaviour of the routing. The option accepts the following values:
 
-The option `fallbackControl` is a string that can have two values:
-
-- `none`, the **default** Astro should not do anything, and a 404 is rendered;
-- [`redirect`](#redirect), Astro should use a redirect when re-routing to the destination locale; when `defaultLocale` is set, Astro will create a redirect to `defaultLocale` 
-
-The logic is triggered **only** when the path `/` is visited.
-
+- `prefix-always`: all URLs of the website must have a locale prefix. Astro will return a 404 for any route that doesn't fulfill the requirements.
+  Use `example.com/[lang]/content/` for every locale.
+- `prefix-other-locales`: the URLs of the default locale must not have a prefix, while the rest of locales must have a locale prefix. 
+  Use `example.com/content/` for the default locale. Use `example.com/[lang]/content/` for other locales.
 
 ### Fallback system
 
@@ -221,51 +217,29 @@ The key is the locale that should benefit from the fallback system and the value
 Astro will throw an error if any locale in `fallback` (keys and values) isn't present in the `locales` list.
 
 
-#### Redirect
+### Browser locales detection 
 
-In the example below, we introduce a `pt_BR` in the `locales` list, and we tell Astro via the `fallback` key that `pt_BR` should fall back to the `pt` locale.  We use the `fallbackControl` option to tell Astro that it must apply a redirect for missing pages that belong to the `pt_BR` locale.
+In SSR, Astro is able to parse the `Accept-Language` header and provide a list of preferred locales by the user **and** supported by the application.
 
-```js
-// astro.config.mjs
-import {defineConfig} from "astro/config"
-export default defineConfig({
-    i18n: {
-        defaultLocaLe: 'en',
-        locales: ['en', 'es', 'pt_BR', 'pt', 'fr'],
-        fallback: {
-            pt_BR: 'pt'
-        },
-        fallbackControl: "redirect"
-    }
-})
-```
+This list is sorted by the highest to the lowest using the [quality value](https://developer.mozilla.org/en-US/docs/Glossary/Quality_values).
 
-Let's suppose a user navigates to `/pt-br/welcome`, and the developer **didn't** create the route `src/pages/pt_BR/welcome.*`, Astro will attempt to **redirect** said user to the `/pt/welcome` URL. If the developer **does** have `src/pages/pt/welcome.*` in their file system, the user will see the page correctly, **otherwise** the user will see the 404 page.
+This information is available through the global object `Astro`: 
+- `Astro.preferredLocaleList: string[] | undefined`
+    
+  For example, if `i18n.locals` contains `['pt', 'fr', 'de']`, and the value of the `Accept-Header` value is `en, fr;q=0.2, de;q=0.8, *;q=0.5`, then 
+  `Astro.preferredLocaleList` will be `['de', 'fr']` because `pt` isn't inside the header, and `en` isn't supported by the website. `de` comes first because it has a highest quality value.
+  
+  The property might be `undefined` if the developer isn't using their site in SSR. When `Accept-Header` is `*`, the list contained in `i18n.locales` is returned. `*` means that no preferences have been set, so all the original locales are supported and preferred. 
 
-### Browser detection
+- `Astro.preferredLocale`
+
+  For example, if `i18n.locals` contains `['pt', 'fr', 'de']`, and the value of the `Accept-Header` value is `en, fr;q=0.2, de;q=0.8, *;q=0.5`, then
+  `Astro.preferredLocale` will be `de` because it has a highest quality value.
+
+  The property might be `undefined` if the developer isn't using their site in SSR, or the highest value of `Accept-Header` is `*`. 
 
 > [!NOTE]
-> This feature requires adapter support using an Astro feature
-
-An opt-in feature that allows to detect the `Accept-Langauge` header sent by the browser, and redirect users the corresponding locale. If a locale isn't supported - AKA isn't present in the `locales` list - Astro won't execute any redirect.
-
-The user can enable the feature using the `detectBrowserLanguage` configuration:
-
-```js
-// astro.config.mjs
-import {defineConfig} from "astro/config"
-export default defineConfig({
-    i18n: {
-        defaultLocaLe: 'en',
-        locales: ['en', 'es', 'pt_BR', 'pt', 'fr'],
-        detectBrowserLangauge: true
-    }
-})
-```
-
-The logic is triggered **only** when the path `/` is visited. When the logic is triggered, Astro attempts to redirect the user to:
-- the `/<LOCALE>` path, where `<LOCALE>` is the locale matched in `locales` list;
-- the [sub-domain/domain](#domain-support) of `<LOCALE>`, if there's a match;
+> This feature is only available in **SSR**
 
 ### Domain support
 
--- a/proposals/0041-i18n-routing.md
+++ b/proposals/0041-i18n-routing.md
@@ -1,0 +1,408 @@
+- Start Date: 2023/10/19
+- Reference Issues:
+- Implementation PR:
+
+# Summary
+
+First-class support for localized routes, aka i18n routing.
+
+
+# Background & Motivation
+
+Many websites need to ship support for translated/localised websites for many reasons:
+- legal;
+- localised content;
+- localised market;
+- etc.
+
+Nowadays, there are [workarounds](https://docs.astro.build/en/recipes/i18n/) in Astro to make it work, although these workarounds have limitations, and because of that many users can't ship their website properly, or they have to work more to compensate the lack of first-class support.
+
+# Goals
+
+- Localised routes with locale prefixes;
+- Default locale with no prefix;
+- Redirect to default locale if prefix enabled;
+- Localise injected routes;
+- Domain support, with the help of [Adapter features](https://docs.astro.build/en/reference/adapter-reference/#adapter-features), so this will be bound to the limitations of the hosting provider;
+- Provide the necessary APIs for integrations and libraries to request information about the current locales;
+- Locale detection via the `Accept-Language` header, so support SSR;
+- Provide first-class APIs to users to work around locales (`.astro` components, endpoints, middleware);
+
+# Non-Goals
+
+- Localised data (dates, numbers, plurals, et.c);
+- Dictionaries where users can store translations of pre-defined words;
+- SEO optimisations;
+
+This gives the reader the correct context on what is intentionally left out of scope.
+It is okay to leave this section empty.
+
+# Detailed Design
+
+## Opt-in configuration
+
+The feature is opt-in, to avoid disrupting existing websites. To enable the feature, a new configuration called `i18n` is available:
+
+```js
+// astro.config.mjs
+import {defineConfig} from "astro/config"
+export default defineConfig({
+    i18n: {
+        
+    }
+})
+```
+
+The feature requires **two required** fields, where the user needs to store the default locale via `defaultLocale`, and a list of available locales via `locales`:
+
+```js
+// astro.config.mjs
+import {defineConfig} from "astro/config"
+export default defineConfig({
+    i18n: {
+        defaultLocaLe: 'en',
+        locales: ['en', 'es', 'pt', 'fr']
+    }
+})
+```
+
+Astro will throw an error if the `defaultLocale` value is not present in the list of `locales`.
+
+## File system based
+
+The feature is file system based and doesn't force users to have `/en`, `/es`, etc. directories. Astro doesn't force users to have these directories in a certain folder, other than being at least in the `/pages` directory.
+
+Missing translated pages will render a 404 by default.
+
+## Logic via middleware
+
+Most of the logic will leverage the [middleware](https://docs.astro.build/en/guides/middleware/) system.
+
+If a user has a `middleware.ts` file when the i18n routing is enabled, Astro will place its i18n middleware right after the one of the user:
+
+```js
+pipeline.setMiddlewareFunction(
+    sequence(userMiddleware.onRequest, createI18nMiddleware(config))
+)
+```
+
+By placing the middleware **after** the one of the user, Astro allows users to apply their business logic to the emitted `Response` of the i18n middleware.
+
+## Adapters and Astro features
+
+Some features can be supported only with the help of the adapter.
+
+A new [Astro features](https://docs.astro.build/en/reference/adapter-reference/#astro-features) will be introduced. The features will be progressively presented when talking about the features.
+
+## Features
+
+Below the list of additional features that Astro will be provided with the i18n routing.
+
+### A new virtual module called `astro:i18n`
+
+> [!NOTE]
+> This feature doesn't require the adapter help
+
+A virtual module called `astro:i18n` will be available to retrieve important information useful to frontend and backed.
+
+Here's a list of APIs available to users to retrieve information:
+
+#### `getRelativeLocaleUrl(locale: string): string`
+
+Given a locale, the function will return the **relative** URL, without the website at the beginning. The function respects the configurations `base`, `trailingSlash` and `build.format`.
+
+```astro
+---
+// src/pages/index.astro
+import { getRelativeLocaleUrl } from "astro:18";
+console.log(getRelativeLocaleUrl('es')) // will log "/es"
+---
+```
+
+Another example, using `base`:
+
+```js
+// astro.config.mjs
+import {defineConfig} from "astro/config"
+export default defineConfig({
+    base: '/docs',
+    i18n: {
+        defaultLocaLe: 'en',
+        locales: ['en', 'es', 'pt', 'fr'],
+        domains: {
+            pt: "https://example.pt"
+        }
+    }
+})
+```
+
+```astro
+---
+// src/pages/index.astro
+import { getAbsoluteLocaleUrl } from "astro:18";
+console.log(getRelativeLocaleUrl('es')) // will log "/docs/es"
+---
+```
+
+#### `getAbsoluteLocaleUrl(locale: string): string`
+
+Given a locale, the function will return the **absolute** URL, taking into account the [domain](#domain-support) supported. The function respects the configurations `base`, `site`, `trailingSlash` and `build.format`.
+
+```astro
+---
+// src/pages/index.astro
+import { getAbsoluteLocaleUrl } from "astro:18";
+console.log(getRelativeLocaleUrl('es')) // will log "http://localhost:4321/es"
+---
+```
+
+With domain support enabled for a `locale`, the returned value will be slightly different:
+
+```js
+// astro.config.mjs
+import {defineConfig} from "astro/config"
+export default defineConfig({
+    i18n: {
+        defaultLocaLe: 'en',
+        locales: ['en', 'es', 'pt', 'fr'],
+        domains: {
+            pt: "https://example.pt"
+        }
+    }
+})
+```
+
+```astro
+---
+// src/pages/index.astro
+import { getAbsoluteLocaleUrl } from "astro:18";
+console.log(getRelativeLocaleUrl('pt')) // will log "https://example.pt/"
+---
+```
+
+#### `getRelativeLocaleUrlList(locale: string): string[]`
+
+Same as `getRelativeLocaleUrl`, but it will return all the locales supported.
+
+#### `getAbsoluteLocaleUrlList(locale: string): string[]`
+
+Same as `getAbsoluteLocaleUrl`, but it will return all the locales supported.
+
+### Fallback control
+
+An option that tells Astro what to do in case it doesn't find a page that belongs to one of the `locales` configured.
+
+The option `fallbackControl` is a string that can have two values:
+
+- `none`, the **default** Astro should not do anything, and a 404 is rendered;
+- [`redirect`](#redirect), Astro should use a redirect when re-routing to the destination locale; when `defaultLocale` is set, Astro will create a redirect to `defaultLocale` 
+
+### Fallback system
+
+The fallback system is a feature that allows users to re-route users from one locale to another in case a page is missing.
+
+The fallback system is an opt-in feature.
+
+The fallback system is configured using `fallback`. `fallback` is an object where both keys and values must be locales.
+
+The key is the locale that should benefit from the fallback system and the value is the locale where Astro should re-route.
+
+Astro will throw an error if any locale in `fallback` (keys and values) isn't present in the `locales` list.
+
+
+#### Redirect
+
+In the example below, we introduce a `pt_BR` in the `locales` list, and we tell Astro via the `fallback` key that `pt_BR` should fall back to the `pt` locale.  We use the `fallbackControl` option to tell Astro that it must apply a redirect for missing pages that belong to the `pt_BR` locale.
+
+```js
+// astro.config.mjs
+import {defineConfig} from "astro/config"
+export default defineConfig({
+    i18n: {
+        defaultLocaLe: 'en',
+        locales: ['en', 'es', 'pt_BR', 'pt', 'fr'],
+        fallback: {
+            pt_BR: 'pt'
+        },
+        fallbackControl: "redirect"
+    }
+})
+```
+
+Let's suppose a user navigates to `/pt-br/welcome`, and the developer **didn't** create the route `src/pages/pt_BR/welcome.*`, Astro will attempt to **redirect** said user to the `/pt/welcome` URL. If the developer **does** have `src/pages/pt/welcome.*` in their file system, the user will see the page correctly, **otherwise** the user will see the 404 page.
+
+### Browser detection
+
+> [!NOTE]
+> This feature requires adapter support using an Astro feature
+
+An opt-in feature that allows to detect the `Accept-Langauge` cookie sent by the browser, and redirect users the corresponding locale. If a locale isn't supported - AKA isn't present in the `locales` list - Astro won't execute any redirect.
+
+The user can enable the feature using the `detectBrowserLanguage` configuration:
+
+```js
+// astro.config.mjs
+import {defineConfig} from "astro/config"
+export default defineConfig({
+    i18n: {
+        defaultLocaLe: 'en',
+        locales: ['en', 'es', 'pt_BR', 'pt', 'fr'],
+        detectBrowserLangauge: true
+    }
+})
+```
+
+The logic is triggered **only** when the path `/` is visited. When the logic is triggered, Astro attempts to redirect the user to:
+- the `/<LOCALE>` path, where `<LOCALE>` is the locale matched in `locales` list;
+- the [sub-domain/domain](#domain-support) of `<LOCALE>`, if there's a match;
+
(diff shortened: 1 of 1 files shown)


ematipico:
@SudoCat @jlarmstrongiv @maddsua @delucis 

I am looping you in because you gave excellent support and advice, so it would be great if we could shape this RFC together and answer all the questions.

maddsua:
I might've missed it, but do we have a function that just returns language code?

A usecase for it would be like follows:

```typescript
//  frontmatter
import { getLanguage } from 'whatever';
const lang = getLanguage();
```
```jsx
//  template
<p>
  { {
    uk: 'Щось українською',
    en: 'Something in English',
    es: 'Algo en español',
    fr: 'Quelque chose en français'
  }[lang] }
</p>
```

This is how I currently do localization for pages that can't be shoved into separate md file for each separate language.

This approach works well in both SSR mode, where we pass language code in the fly, and SSG, where this variable gets passed through getStaticPath's props:
```typescript
//  src/pages/[...lang].astro

import { localesList } from '@/data/locales.json';

const mkIntlPath = (langs: string[] | 'auto', dynamicRouteSlug: string) => {
  const pageLanguages = langs === 'auto' ? localesList.map(item => item.code) : langs;
  return pageLanguages.map(lang => ({
    params: { [dynamicRouteSlug]: (lang === localesList[0].code ? undefined : lang) },
    props: {  lang  }
  }));
};

export function getStaticPaths() {
  return mkIntlPath('auto', 'index');
}
const { lang } = Astro.props;
...
/* 

Will render a page for each language in localesList:
  --> index.html
  --> en.html
  --> fr.html
  --> es.html
  and so...

Call mkIntlPath with and array of language codes
to use something different than app-global language list, like so

mkIntlPath(['es', 'fr'], 'index');

*/
```

SudoCat:
This looks like a great start! I think I'll need a lil more time to gather some more detailed thoughts, but at first glance:

## Default Locale and Prefixes
- I see `Redirect to default locale if prefix enabled;` is mentioned in goals, but there doesn't seem to be a configuration specification designed. From some brief research, the approaches in other frameworks vary. Next always removes the default prefix, and has some nasty workarounds instead - we should plan ahead to avoid this. Nuxt on the other hand has a "strategy" configuration option, with numerous values. Something like this probably gives the best future compatibility, as we can use an enum which can be extended if new edge cases arise in the future. 
- If i18n is enabled, should all routes be localised? for instance, Astro server endpoints may not want to be localised if their content is not localisation specific. Is this a situation that we should consider in the design (by making it possibly to have routes not localised), or should this just be an edge case we document for awareness?
  - if we are to handle this, my suggested solution would be a rollup extension similar to how SSG/SSR/Hybriad works, with an `export const localize: boolean;`. This could also make it easier for teams migrating to the new i18n routing from a different solution, as they could opt-in or out as appropriate.

## Locale Standardisation and edge cases
- Should we support any valid string as a locale, or should we strictly only support [valid UTS locale identifiers](https://www.unicode.org/reports/tr35/tr35-59/tr35.html#Identifiers) or some other such standard?
- If we are standardising them, should it be possibly to rewrite them in the URLs by some means? It's quite common for websites in to use `/uk` for their `en-gb` websites. (I personally think this is an edge case, and if it is required, a developer can handle it with URL rewriting on their http server/hosting provider/CDN)

## Domain Routing
- If a domain for a locale is defined in the configuration, will Astro still create `/[locale]/` routes for that locale on the root domain? 
- Likewise, will the subdomain support the other locales?

I suppose a lot of this is dependent on the adapters?

given:
```
site: 'https://example.com'
i18n: {
  defaultLocale: 'en',
  locales: ['en', 'fr']
  domains: {
    fr: 'https://fr.example.com'
  }
}
```
What will the user see if they visit `https://example.com/fr`?
What will the user see if they visit `https://fr.example.com/en`?

##  Middleware
I'm unsure about the middleware setup
- Is Astro now capable of injecting middleware into a users middleware choices? (I was not aware of this capability)
- Would this middleware be responsible for determining which locale the user is on? Surely this would need to be executed before the user's middleware, so that their middleware could also read the locale 

## Usage
I believe we'll need to specify how users can access the Locale, and what information should be available. It will need to be available to pages, endpoints and middleware. 

With this in mind, I think this should be added as a separate key on the Astro global, likely named `i18n` for consistency. 

Something like this:

```tsx
type i18nGlobal = {
  locale: string; // the locale identifier, as defined
  locales: string[] // the list of all available locales
}
```

I don't believe anything else should be needed on this object immediately. This should provide what is needed to render localised content, and show information relating to other locales if required.

## Misc
- I'm somewhat confused reading the [file system based](https://github.com/withastro/roadmap/blob/i18n-routing/proposals/0041-i18n-routing.md#file-system-based) section. It says that the user will not be required to have locale-folders, and then goes on to say these folders should at least be in the pages folder. I think this section needs rewording.

lilnasy:
I don't understand the motivation behind [`getRelativeLocaleUrl`](https://github.com/withastro/roadmap/blob/i18n-routing/proposals/0041-i18n-routing.md#getrelativelocaleurllocale-string-string) and related functions. What use case do they enable? What would an example application of it look like?

SudoCat:
It's quite common to only use relative URLs, especially on the frontend - you don't tend to define your navigation using absolute URLs with domains. Most websites I've worked on use absolute pathnames for their navigation links, without defining a hostname.

However you raise a good point - relativeLocaleUrl might be the wrong term - the URL returned is still an absolute pathname, but without a protocol and domain. Perhaps we should rename these to `getLocalePathname` and `getLocaleHref` to closer conform to the [URL ](https://developer.mozilla.org/en-US/docs/Web/API/URL) object behaviour.

lilnasy:
Thanks, I think that explains it well - you would use these functions to manually construct links.
```astro
---
import { getRelativeLocaleUrl } from "astro:18";
import { removeLocaleBase } from "../helper-functions.ts"
---
<a href={getRelativeLocaleUrl('es') + '/' + removeLocaleBase(Astro.request.url)}>Read this article in Spanish</a>
<a href={getRelativeLocaleUrl('it') + '/' + removeLocaleBase(Astro.request.url)}>Read this article in Italian</a>
<a href={getRelativeLocaleUrl('fr') + '/' + removeLocaleBase(Astro.request.url)}>Read this article in French</a>
```

jlarmstrongiv:
@lilnasy I created several [helpers](https://github.com/jlarmstrongiv/astro-i18n-aut/tree/main/src/edge-runtime) in [`astro-i18n-aut`](https://www.npmjs.com/package/astro-i18n-aut).  I found that several settings affect building urls:
- build output file / directory
- trailing slash always or never
- base url
- `defaultLocale` vs other locales
- is a valid and configured locale or not
- root url or not
- url concatenation (avoiding `/es//path/joined//together`)

There’s so many options and config variations that I decided to build these helpers into the package.  

The only case I haven’t had the time yet to implement is the relative urls.  But, my goal was to avoid manual concatenation to avoid mistakes in joining paths.

I think it would be nice for Astro to provide helpers like the proposal described because there are so many cases and it’s very easy to get it wrong.  It’d be nice to use translated urls without thinking about it every time.

---

I’m traveling this week and haven’t had a chance to go over the full proposal or questions above.  I’ll check back this weekend.

SudoCat:
hmmm your code examples raises a really good point that perhaps we should have a method that can return a specific link in a specific locale, that will automatically handle replacing the locale if found with the one requested, with handling for default locale prefix removal and everything. Something like `getRelativeLocaleUrl(locale: string, pathname: string)` and `getAbsoluteLocaleUrl(locale: string, pathname: string)`.

As a stretch goal, we should maybe consider handling Astro route urls: `getRelativeLocaleUrl(locale: string, pathname: string, params: Record<string, string>)`. E.g. `getRelativeLocaleUrl("en", "products/[sku]", { sku: 'foo' })` :eyes:

Tc-001:
IMO it currently seems way too repetitive to be used all across a project (even more if migrating to it) 

Maybe something like `prependLocale(url, locale?)`
`prependLocale("/foo") // /docs/en/foo` 

A a nice approach could actually be a special i8n `<A>` tag that handle everything for you if you just give it the href (without the base or language)

ematipico:
@maddsua 

> I might've missed it, but do we have a function that just returns language code?

We don't, but we can definitely add it

___

@SudoCat 

> I see Redirect to default locale if prefix enabled; is mentioned in goals, but there doesn't seem to be a configuration specification designed.

Yeah, I want to strike a solution in the middle. I just merged a PR where we implement the Next.js behaviour. Although, I would like to understand the needs and shortcomings. 

I like the Nuxt approach, but there are too many options IMHO and their documentation isn't the best because it lacks examples in the documentation. 

> If i18n is enabled, should all routes be localised?

No, they should not. The idea is that the user decides where to have the localized folders/routes. Astro helps in doing proper routing. 

> if we are to handle this, my suggested solution would be a rollup extension similar to how SSG/SSR/Hybriad works, with an export const localize: boolean;.

How would this help? You might be onto something here 🤔 

> I suppose a lot of this is dependent on the adapters?

Yes, and it's possible that domain support needs to make various assumptions in order to work. 

> What will the user see if they visit https://example.com/fr?
> What will the user see if they visit https://fr.example.com/en?

It's possible that the adapters should give the ability to do so, Astro will provide the proper information to do so. However, to set expectations right, I believe that `https://example.com/fr` needs to be available **at least** in development and production. `https://fr.example.com/en` should not be available.


>    Is Astro now capable of injecting middleware into a users middleware choices? (I was not aware of this capability)

This was always a possibility. With the `sequence` API you can compose middleware functions.

>    Would this middleware be responsible for determining which locale the user is on? Surely this would need to be executed before the user's middleware, so that their middleware could also read the locale

Can you expand on this? The user can already read the locale from the headers. The RFC explains why the middleware is set after the user middleware, although we can put it before if you think it's best, although we need to have a stronger argument.

ematipico:
The reason why I created `getRelativeLocaleUrl` is mostly to **help** users to get the base URL right. Astro provides **many** options that make links very difficult to get right:
- `base`
- `trailingSlash`
- `build.format`

These are a lot combinations, as [you might see from the tests](https://github.com/withastro/astro/blob/feat/i18n-routing/packages/astro/test/units/i18n/astro_i18n.js). Astro already solved the issue internally, so we should take advantage and help users to create URLs with locales. 

Although I am not entirely sold on this utility, because this utility has some assumptions that I don't like. Maybe `prependLocale(url, locale?)` is what we need?

SudoCat:
@ematipico 

> I like the Nuxt approach, but there are too many options IMHO and their documentation isn't the best because it lacks examples in the documentation.

Yeah I definitely think we can do a little better than that. I think for starters, we should just add a `routingStrategy` configuration, with an enum for values. Just stick to two simple values for now, with room to expand if needed. This could potentially be expanded upon for any domain routing configuration requirements.

```
type RoutingStrategy = {
  PrefixExceptDefault,
  PrefixAlways,
}
```

If PrefixExceptDefault is selected, (the default option), then the default locale will not be prefixed, and all other locales shall be prefixed.
If PrefixAlways is selected, then all locales, including the default will use a prefix.

If we need later on, we could consider options like `PrefixExceptDomain` or something.

> No, they should not. The idea is that the user decides where to have the localized folders/routes. Astro helps in doing proper routing.

Right okay, I'd like to see some examples of how this should work to better understand it. I'm not sure it's clear right now how this would work. Would this mean a user would create a folder like `src/pages/[locale]` to put their localised pages/endpoints in? 

> How would this help? You might be onto something here 🤔

It would allow users to opt specific pages in to localisation, regardless of where the file lived. It would avoid needing to restructure your project around localisation.

It's not without flaws though - potentially more complicated and could increase build time overheads.

This could also be extended to allow more complicated configuration objects - such as localising static routes or even folder names via `index.astro` files:

```tsx
export const localization = { en: 'about', fr: 'à-propos' }
```

> This was always a possibility. With the sequence API you can compose middleware functions.

I've used Sequence before, but I wasn't aware Astro core could inject middleware into the user's defined sequence. I presumed the user's would be required to manually include the middleware in their own sequence. I remember that integrations could not inject middleware into a sequence, so didn't expect there to be an internal API for it.

> Can you expand on this? The user can already read the locale from the headers. The RFC explains why the middleware is set after the user middleware, although we can put it before if you think it's best, although we need to have a stronger argument.

Would this be via reading the Accept-Language header, or by parsing the hostname manually to detect the locale? 

I guess my question hinges on how Astro handles parsing the locale, and providing that information to the developer, and what exactly the middleware will be doing. My understanding is that most of the i18n implementation will need to happen within Astro before the request even reaches the middleware, at the routing layer. What else would the middleware do?

I might be misunderstanding middleware here, but aren't they processed in order of definition, not reverse? So if `createI18nMiddleware` is the last middleware in the sequence, then the i18n middleware response will not be available to user-defined middleware. If this i18n middleware is responsible for detecting and parsing the current locale, redirecting routes, wouldn't other middleware need to be aware of these changes first?

For example, in the website I recently worked on, we used middleware for creating our CMS API Client instance, so it was available to all pages. This API Client needed to be created with the locale as a parameter. 

Sorry for a lot of questions here, but I think I need to better understand the underlying implementation of the middleware/localisation to know where the middleware should sit in the chain.

ematipico:
@SudoCat don't worry about the questions, that's what the RFC is for.

I really like the proposal strategy, I will update the RFC accordingly.

> Right okay, I'd like to see some examples of how this should work to better understand it. I'm not sure it's clear right now how this would work. Would this mean a user would create a folder like src/pages/[locale] to put their localised pages/endpoints in?

Users can have their localised routes wherever they want, e.g. `src/pages/en_AU`, or `src/pages/blog/en_AU`. Conversely, I also understand the need for a possible opt-out feature with an `export const localize: boolean`. I will think about it.

> I guess my question hinges on how Astro handles parsing the locale and providing that information to the developer, and what exactly the middleware will be doing

I see your point now, and it makes sense. 

So here's how your middleware functions work. Suppose you have three middleware functions: `hello`, `validation` and `auth`, in this order. The journey of `Request` and `Response` happens as follows:

```markdown
		Request				Request
hello --------> validation --------> auth ┐
                                          |
                                          | Rendering, create Response
                                          |
hello <-------- validation <-------- auth ┘
		Response			Response
```

`auth` is the last one the get the `Request` and the first one to get the `Response`. If we swap `auth` with our i18n middleware, the reason why I put it there is because I want to give the chance to the users to handle the `Response` emitted by the i18n middleware, in case it's needed. 

However, your argument that we need to calculate the current locale is very important.

itsmatteomanf:
> We don't, but we can definitely add it

This is a must for implementing `<select>` to allow for language selection. A list of locales and the current locale are a must...

Ideally, it would allow for adding a name to each locale, so that it can be shown and fetched without relying on separate config options, but that is not a big issue to work around.

ematipico:
The signature of the APIs has been updated here [`4718adb` (#734)](https://github.com/withastro/roadmap/pull/734/commits/4718adb5ad0a9128ba20805b53c533de9eaf2e7a)

itsmatteomanf:
@ematipico i have seen the changes. I get what you are going for with the path argument... but I'd add a simple `getCurrentLocale()` and `getAllLocales()` or similar, you can do it easily with these, but it's an additional step.

Maybe return an array of objects with code, domain, root URL, etc.? As with the domain support it gets complicated if you just return the ISO code strings.

ematipico:
I'll add `getCurrentLocale()` to the RFC when I have more technical information I can share. I am exploring the API and how it will work both in the backend and frontend. I don't feel comfortable adding an API if I can't explain the internals.

About `getAllLocales()`, can you give me more info about it? Like use cases

itsmatteomanf:
Understood, thank you. It's the first RFC I'm following, so I don't know exactly how they work here on Astro 😊

Regarding the `getAllLocale()`, the main use case would be a page (or component, like a `<select>`) that lists all locales with links. The first can be done currently relatively easily, the second one it means the values of each option would be a path, which is not the worst (but may be problematic with long URLs). The issues arise when I want to get a name of the language from something like `i18next`, which wants the language code... what do I do then? Do I manually keep a separate list? Do I strip the URL of extra stuff from values with `path = ""`?

This is of course less of a problem when you start having domains, as you need that in the front end to redirect, but when you have just a path which includes the language code, the JS is a lot simpler if you just keep it to the language code. Especially if you have *very* long URLs as I do...

proposals/0041-i18n-routing.md
```diff
@@ -0,0 +1,377 @@
...
+The user can enable the feature using the `detectBrowserLanguage` configuration:
+
+```js
+// astro.config.mjs
+import {defineConfig} from "astro/config"
+export default defineConfig({
+    i18n: {
+        defaultLocaLe: 'en',
+        locales: ['en', 'es', 'pt_BR', 'pt', 'fr'],
+        detectBrowserLangauge: true
```
matthewp, line 261: bikeshed: Should it be `detectLanguage` or something else? Not necessarily coming from a browser.

jlarmstrongiv:
@itsmatteomanf and @ematipico I had a similar function called [`getAllLocaleUrls`](https://github.com/jlarmstrongiv/astro-i18n-aut/blob/main/src/edge-runtime/getAllLocaleUrls.ts) for that exact purpose of a language picker.

---

Another helpful addition may be to expose parts of the config in `astro:18` for users and custom middleware.  For example, accessing the config below would be helpful to build my own utilities and handle my own edge cases:

```js
{
  i18n: {
        defaultLocaLe: 'en',
        locales: ['en', 'es', 'pt', 'fr'],
        domains: {
            pt: "https://example.pt"
        }
    }
}
```

---

I like `export const localize: boolean` and `export const localization = { en: 'about', fr: 'à-propos' }`.  It’s important to be able to opt-out and having a way to translate the urls would be amazing!  

---

> Users can have their localised routes wherever they want, e.g. src/pages/en_AU, or src/pages/blog/en_AU. Conversely, I also understand the need for a possible opt-out feature with an export const localize: boolean. I will think about it.

In [`astro-i18n-aut`](https://www.npmjs.com/package/astro-i18n-aut), I assumed every page was localized unless the globs excluded it.  I like the idea of using the default pages directory with `export const localize: boolean` if it’s not localized, similar to the `export const prerender: boolean`.  

Though, I remember a previous conversation that emphasized that some content would be region locked, meaning some translations and product pages would be available in some locales, but not in others. 

The benefit of `export const localization = { en: 'about', fr: 'à-propos' }` would allow for both localizing urls and listing only partial translations, if a page is supported in English and French, but not Spanish for example.  The `getAllLocaleUrls` would also need to take into account the translated urls and partial page translations too.

Just something to think about

proposals/0041-i18n-routing.md
```diff
@@ -0,0 +1,377 @@
...
+```
+
+The feature requires **two required** fields, where the user needs to store the default locale via `defaultLocale`, and a list of available locales via `locales`:
+
+```js
+// astro.config.mjs
+import {defineConfig} from "astro/config"
+export default defineConfig({
+    i18n: {
+        defaultLocaLe: 'en',
```
matthewp, line 68:
```suggestion
        defaultLocale: 'en',
```

proposals/0041-i18n-routing.md
```diff
@@ -0,0 +1,377 @@
...
+        locales: ['en', 'es', 'pt', 'fr']
+    }
+})
+```
+
+Astro will throw an error if the `defaultLocale` value is not present in the list of `locales`.
+
+## File system based
+
+The localized directories must inside the `pages/` directory. Other than this restriction, the user is free to place the localized folders anywhere.
```
matthewp, line 78: I don't think I understand this answer, it doesn't really matter, right? The implementation doesn't know about / care about where the underlying pages are coming from, just that they answer with a `200`?

matthewp:
> I like export const localize: boolean and export const localization = { en: 'about', fr: 'à-propos' }. It’s important to be able to opt-out and having a way to translate the urls would be amazing!



In the past we've avoided similar types of APIs (someone suggested something like this for redirects) because Astro doesn't read all of your files at startup. It only reads them when requested. This makes our startup time faster. So I don't think we should add this kind of feature that would require first compiling every page. Even though I agree it's a nice sort of API.

artursopelnik:
Hello everyone. I have an advanced workaround that uses content collection, trailing slash and basic support. Maybe you can learn something for this RFC here or use it in the meantime.

**Repo:**
https://github.com/artursopelnik/astro-translate-routes-example

**Demo:**
https://artursopelnik.github.io/astro-translate-routes-example/

**Article:**
https://www.webdesign-sopelnik.de/en/blog/translate-routes-for-astro-content-collections-or-subpages-with-trailingslash-and-base-support/

itsmatteomanf:
@ematipico re: root page redirect.

If I have the routing strategy set as `prefix-other-locales` the root page (`pages/index.astro`) will be the default language home page. And that is fine and expected.

If I were to set the routing strategy to `prefix-always` the `pages/index.astro` page would be empty, or a redirect. It'd be nice to have the option to select if you wanted to have the redirect or not, so one could do a landing page to select a language, or anything the user wants. I know it's currently not implemented, so it's perfectly fine and allows both use cases, but I'd like for there to be an opt-out to that behaviour if it ever were to become standard :)

itsmatteomanf:
An additional, and important, thing to consider is how to handle the `pages/404.astro` and related files.

Some services (Cloudflare Pages, for example) expect a `404.html` file in the root of the public folder to define if an app is a SPA or not. It will not be an issue for `prefix-other-locales`, but not sure how it currently is handled for `prefix-always`. Does it keep that file or causes issues? Or is it overwritten with a redirect?

lilnasy:
I think there needs to be a story for error pages in general.

delucis:
We discussed this in [early planning](https://www.notion.so/i18n-routing-7a10fe1b77c44501ae691f401ac02802?d=ffc027893b22406197cede30b22261e0&pvs=4#52e1a2add0864fd490ebe743d01bd7ef), but @ematipico and I were just chatting and I realised this wasn’t covered in this RFC: currently `locales` lets you set the URL path segments to be treated as a specific locale, but there’s no explicit mapping from that to a language:

```js
locales: ['en', 'it']
// Enables: example.com/en and example.com/it to be treated as separate locales
```

The current design implicitly assumes that `'en'` and `'it'` in this case are ALSO the current _language_ for features such as `Astro.preferredLanguage` or redirecting according to `Accept-Language` headers. However, languages must be specified in the restrictive [BCP-47 format](https://www.w3.org/International/questions/qa-choosing-language-tags) in order to be interoperable with things like `Accept-Language` headers or `Intl` JavaScript APIs. Because the current design makes a 1:1 mapping between URL path and language, it places a restriction on how people design their URLs: they **must** use a valid BCP-47 tag for their locales.

Some example use cases:

- I have a site serving US and UK customers and want to serve them as `example.com/usa` and `example.com/uk`. Neither `usa` nor `uk` are valid BCP-47 tags, their equivalents would be `en-US` and `en-GB`.
-  I have a site with `lang="pt-BR"` content, but my host doesn’t support uppercase characters in URLs (e.g. Netlify), so must serve at `example.com/pt-br`.

To offer one possibility, the config could look something like

```ts
interface PartialI18nConfig {
	locales: Array<string | { path: string, lang: string }>
}
```

So for the examples above config could look like:

```js
{
	locales: [{ path: 'usa', lang: 'en-US' }, { path: 'uk', lang: 'en-GB' }]
}
```
```js
{
	locales: [{ path: 'pt-br', lang: 'pt-BR' }]
}
```

You could still user the string shorthand when your locale and lang are the same, but have a way to explicitly set the locale => language mapping when needed.

Another option would be switching to a map of locale keys to language values. This would mirror what [`@astrojs/sitemap` does in its i18n config](https://docs.astro.build/en/guides/integrations-guide/sitemap/#i18n).

itsmatteomanf:
@delucis one variation not covered in you post is that one may want content to be the same for multiple variations.

For example Italian has 4 BCP-47 options:
- `it-CH`
- `it-IT`
- `it-SM`
- `it-VA`

All these span a geographic area which almost never brings different content, there are cases, of course, but not many.
So a mapping that allows multiple `Accept-Language` values to map to a single `it`, both as main locale (which HTML support as `lang="it"`) and path would be good. Especially if this RFC ever gets expanded to a more fully fledged i18n implementation that offers a way to get the current locale for the build.

delucis:
# Routing strategy config

_Picking up some discussion around the shape of the routing config started in https://github.com/withastro/astro/pull/9006#discussion_r1383559633 so it doesn’t get lost._

## Current design

Currently, the RFC proposes a `routingStrategy` option with an enum of potential values:

```ts
interface PartialI18nConfig {
	routingStrategy: 'prefix-always' | 'prefix-other-locales';
}
```

The intention is to expand these values in the future if/when we add support for subdomains (e.g. `fr.example.com` instead of `example.com/fr`).

I [raised the concern](https://github.com/withastro/astro/pull/9006#discussion_r1383559633) that the current options are not very clear:

- “prefix” doesn’t tell me what it prefixes (a subdomain could be a prefix, a subpath could be a prefix)
- “other” is particularly unclear: other than what?

The configuration options do not make it easy to understand what they do.

## Aligning with user thinking

Given the control users are currently being given (ignoring subdomains for a minute), we can boil down the user question to:

> Do I want to use a subpath for my default locale or not? (`example.com/en/` vs `example.com`)

This is an easy question to answer for most users, but the current design requires you to answer `prefix-always` for “Yes, please prefix my default locale” or `prefix-other-locales` for “No, please don’t prefix my default locale”.

One option to align the configuration with the question a user will ask themselves would be:

```ts
interface PartialI18nConfig {
	routing: {
		prefixDefaultLocale: boolean; // default: false
	}
}
```

This gives users a simple yes/no toggle that aligns with their answer to the question above.

## Subdomains

This brings us to the future and subdomains. IIUC, the string enum solution was chosen to help leave room for subdomains in the `routingStrategy` picture. For example, by adding `'subdomain-always'` and `'sudomain-other-locales'` options (assuming something similar to the current design).

I would argue that this is combining two distinct user questions into one option.

The second user question is:

> Do I want to use subdomains for my locales or subpaths? (`en.example.com` vs `example.com/en`)

This is a different question entirely from “Do I want to use a subpath/subdomain for my default locale?”

Given this, I think it’s ideal to separate the config into two options aligned with these two user questions, for example:

```ts
interface PartialI18nConfig {
	routing: {
		prefixDefaultLocale: boolean; // default: false
		strategy: 'pathname' | 'subdomain'; // default: 'pathname'
	}
}
```

@ematipico raised [a concern](https://github.com/withastro/astro/pull/9006#discussion_r1384787568) here:

> I would like to avoid breaking down the feature in too many flags because they are a footgun, and they become more difficult to maintain.

I can’t comment on the maintenance difference between these setups, but if it is harder to maintain, it may still be worth the trade-off IMO, if the result is a clearer experience for Astro users. If we really prefer the 4-way enum, Zod can happily help us out under the hood:

```js
type RoutingStrategies = 'pathname-always' | 'pathname-non-default' | 'subdomain-always' | 'subdomain-non-default';
const RoutingStrategySchema = z.object({
	prefixDefaultLocale: z.boolean().default(false),
	strategy: z.enum(['pathname', 'subdomain']).default('pathname'),
}).transform(routing =>
	`${routing.strategy}-${routing.prefixDefaultLocale ? 'always' : 'non-default` as RoutingStrategies
);
```

One counter proposal in the original discussion uses `suburl` where I initially proposed `subpath`:

> I used "URL" because that's what the users see, and "path" is misleading because it can be used for files too.

In my examples here I updated the suggestion to `pathname` to match what the standard `URL` object calls that portion of the URL.

proposals/0041-i18n-routing.md
```diff
@@ -0,0 +1,353 @@
...
+export default function createIntegration() {
+  return {
+    name: '@ema/my-adapter',
+    hooks: {
+      'astro:config:done': ({ setAdapter }) => {
+        setAdapter({
+          name: '@ema/my-adapter',
+          serverEntrypoint: '@ema/my-adapter/server.js',
+          supportedAstroFeatures: {
+            i81n: {
```
lilnasy, line 288:
```suggestion
            i18n: {
```
minor typo

proposals/0041-i18n-routing.md
ematipico: No, the implementation doesn't care about it, we care remove this paragraph if you think it can mislead.

tysian:
Hey! I've finished reading docs for experimental i18 features in Astro 3.5.0 and thought this place will be good for my question.

Most of websites I'm working on are translated in couple different languages and data is fetched from CMS with `language` parameter.

So let's say I need to create 10 pages in 10 languages.
Basing on current documentation... I have to manually create (or rather copy/paste) 100 astro files?

Is it (or will be) achievable to render `/[lang]/example` route using `src/pages/example.astro` file?

## Example/Proposal
### Files structure
* `src/pages/index.astro`
* `src/pages/about.astro`
* `src/pages/blog/[slug].astro`

### Config
```js
import { defineConfig } from "astro/config";

export default defineConfig({
  experimental: {
    i18n: {
      defaultLocale: "en",
      locales: ["es", "fr"],
    },
  },
});
```

### Output
This will generate routes for multiple languages:
* `index.astro` -> `/`, `/fr/`, `/es/`, etc.
* `about.astro` -> `/about/`, `/fr/about/`, `/es/about/`, etc.
* `blog/[slug].astro` -> `/blog/slug/`, `/fr//blog/slug/`, `/es//blog/slug/`, etc.

We can get current language as prop or using some helper function from `astro:i18n`

```astro
---
const { lang } = Astro.props;

const data = await fetchFromCMS({language: lang})
---
```

Is it achievable with current state of this feature? Or is this planned for the future?

itsmatteomanf:
@ematipico 

> Do you think it is vital to this feature? I believe this can be created in userland by creating a middleware. We will soon release the feature under the experimental flag, and you can let us know how's the experience.

Vital? No. Nice to have? Definitely. To me it's a simple config option: `redirectRootToDefaultLocale: true`, which I guess does allow for unprefixed routes to go to the default locale, if they exist there.

> I believe - I need to check the code because I don't remember - that when we render the 404, we don't trigger the middleware, which means that `404.astro` is exonerated from the i18n routing logic. Although, we should test it and see if that is correct.

I'm gonna try and do a test... whoever does first will let this thread know. Might be true, if I remember correctly there was some other thing that this behaviour causes in a completely different topic, but I'm gonna check anyway.

> Considering @itsmatteomanf comment, I believe we should go even further and have something like this, too:
> 
> ```js
> {
> 	locales: [
> 		{ path: 'it', lang: [ "it", "it-CH", "it-IT",  "it-SM",  "it-VA"] }
> 	]
> }
> ```
> 
> Or to have `lang` as an array to begin with.

Yes, agreed there. But at this point you really should offer a way to get all these config read from the pages, I know I could do a separate file and load that in both places, but a way to get all languages I configured, with a default language string (something needs to be put in the `lang` HTML attribute...), though.

So maybe something like?

```js
{
	locales: [
		{
			path: 'it',
			default: "it", // or anything, this is the one you want whenever you want to chose things
			lang: [ "it", "it-CH", "it-IT",  "it-SM",  "it-VA"] // these you use for like canonicals, matching routes, etc.
		}
	]
}
```

This would then, ideally, offer a `Astro.i18n` that exposes these settings? So that I can do things simpler next, without actually going into dictating how the user should setup things?

---

@tysian

That is already possible, without this PR.

Just enclose your language pages in a `[lang]/` folder, and use that param to generate all of them.
That'd be just `{ lang } = Astro.params;`

jlarmstrongiv:
In hindsight, I was expecting what @tysian described.  I got close with [`astro-i18n-aut`](https://www.npmjs.com/package/astro-i18n-aut), but still run into problems with astro core in dev.  It’s not possible to have a `defaultLocale` and render all your pages using a template file.  It would be great for Astro to support that use-case

tysian:
@itsmatteomanf 

Yeah, this is the way I was doing it before. I was just creating `src/pages/[lang]/example.astro` and were doing it this way.

But this wasn't possible, as you mentioned, when I want to skip lang prefix for default locale, so I had to copy/paste all my routes pages at least twice (for default lang, and other languages).

The `[...lang]` is not the best solution also, as if pages need to have different tamplates, I need to make conditionals to match exact template needed.

My final workaround was to create `views` directory and using its components instead. The `pages` in that scenario is only for routing purposes, it imports required template from `views` directory and shows is passing all the props, etc.

So and the end of the day, this can't be done without weird workarounds and bad developer experience 😕

It would be pretty cool, to have some solution for it. I've tried to use `fallback` feature, but it redirected me to fallback page and I couldn't get `lang` code from URL (it returned object of fallbacked page instead).



@jlarmstrongiv 

This library looks pretty cool, I'll take a deeper dive into it in my next project, thank you 😉

delucis:
Congrats on the first public release of the experimental i18n API @ematipico! 🎊 
Really excited to see what everyone builds with it and discovers trying it out.

---

Re: locale mapping, I’m not sure I understand what that suggested config would do:

```js
{ path: 'it', lang: [ "it", "it-CH", "it-IT",  "it-SM",  "it-VA"] }
```

If you only have a single path (`it`), wouldn’t it always serve with the same `lang`?

Or is the proposal here only about resolving preferred locale (e.g. via `Accept-Language`)? So that Astro knows that for a request with `Accept-Language: it-SM`, it’s valid to respond with the `/it/` route? Wondering if that could already be handled by `fallback` without needing to complicate this 🤔

itsmatteomanf:
> The `[...lang]` is not the best solution also, as if pages need to have different tamplates, I need to make conditionals to match exact template needed.

No, you don't need any complex conditional logic there.

You can do `src/pages/[...lang]/product/4321.astro`, and inside you do a `getStaticPaths()` which returns the `lang` param to be either the language or `''` if it's the default locale. The only limit here is that if you needed another param *after* `lang` (the other way round works) you can't have it, but it will be incorporated into `lang` itself.

Personally I'd keep it to all pages with the language in the url, much simpler to do, much cleaner as it clearly states the current locale, and way easier to handle locale changes. Of course, though, not all projects can or want this.

I then use Astro locals to make sure all components have the locale set. I could probably do a prop on the layout or something, but it's easier given you don't need to make sure it's passed through multiple layers. There it's available everywhere.

tysian:
@itsmatteomanf 

Woah, I've never thought of that this way. Thank you for all explanations, I'll definitely check this out 😄 

Thanks!

itsmatteomanf:
@ematipico

I tested the 404 page on a `prefix-always` build. It gets ignored, as do any other routes/pages without a prefix.
I have `src/pages/index.astro`, `src/pages/404.astro`, `src/pages/browserconfig.xml.ts` and it ends up with just a `index.html`.

This won't work with Cloudflare Pages, for example.

itsmatteomanf:
@ematipico re: message above.

It seems to ignore also any other folder, I had a folder with routes which created `.js` files for analytics purposes and it gets removed from the build, too.
I guess if it's not in a `/[lang]/` path it gets ignored if the setting is `prefix-always`.

fatbobman:
Currently, when i18n is enabled and the routingStrategy is set to prefix-always, Astro automatically redirects the homepage to the index.astro file in the corresponding defaultLocale directory.

However, in some cases, developers may want to customize this redirection logic, such as based on the browser language or by creating a static language selection page on the root index.

I believe giving developers the option to enable or disable automatic redirection will provide them with greater flexibility.

Could we implement something similar to the following approach:
```mjs
experimental: {
    i18n: {
      defaultLocale: 'en',
      locales: ['en', 'zh'],
      routingStrategy: 'prefix-always',
      enableRedirect: false,      // enable or disable
    }
 },
```

fatbobman:
> @ematipico
> 
> > Do you think it is vital to this feature? I believe this can be created in userland by creating a middleware. We will soon release the feature under the experimental flag, and you can let us know how's the experience.
> 
> Vital? No. Nice to have? Definitely. To me it's a simple config option: `redirectRootToDefaultLocale: true`, which I guess does allow for unprefixed routes to go to the default locale, if they exist there.
> 
> > I believe - I need to check the code because I don't remember - that when we render the 404, we don't trigger the middleware, which means that `404.astro` is exonerated from the i18n routing logic. Although, we should test it and see if that is correct.
> 
> I'm gonna try and do a test... whoever does first will let this thread know. Might be true, if I remember correctly there was some other thing that this behaviour causes in a completely different topic, but I'm gonna check anyway.
> 
> > Considering @itsmatteomanf comment, I believe we should go even further and have something like this, too:
> > ```js
> > {
> > 	locales: [
> > 		{ path: 'it', lang: [ "it", "it-CH", "it-IT",  "it-SM",  "it-VA"] }
> > 	]
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Or to have `lang` as an array to begin with.
> 
> Yes, agreed there. But at this point you really should offer a way to get all these config read from the pages, I know I could do a separate file and load that in both places, but a way to get all languages I configured, with a default language string (something needs to be put in the `lang` HTML attribute...), though.
> 
> So maybe something like?
> 
> ```js
> {
> 	locales: [
> 		{
> 			path: 'it',
> 			default: "it", // this is the one you want whenever you have to chose like a single value for the specific locale 
> 			lang: [ "it", "it-CH", "it-IT",  "it-SM",  "it-VA"] // these you use for like canonicals, matching routes, etc.
> 		}
> 	]
> }
> ```
> 
> This would then, ideally, offer a `Astro.i18n` that exposes these settings? So that I can do things simpler next, without actually going into dictating how the user should setup things?
> 
> @tysian
> 
> That was always possible, without this proposal. This proposal currently allows you to simplify routing between languages.
> 
> Just enclose your language pages in a `[lang]/` folder, and use that param to generate all of them. That'd be just `{ lang } = Astro.params;` and handling like any other param, be it in static paths or SSR.
> 
> I guess the only case in which you may need duplicated files, unless you want to do the `[...lang]/` folder, but it's a mess if you have a need for that in other things, is if you want the default locale without a prefix.

Definitely! At least for me, this feature is very important, and I hope to be able to easily implement the logic based on the root index.

such as based on the browser language or by creating a static language selection page on the root index.

itsmatteomanf:
I did manage a workaround for the problems in the prefixing, but I'd love for that to be solved in core.

You can use `x-default` as the root language, which is the correct language code to use for unspecified language, and then set it as `prefix-other-locales`. There it would work, but of course when you ask for all links you get a unprefixed URL which you need to ignore.

ematipico:
> @ematipico cont: message above.
> 
> It seems to ignore also any other folder, I had a folder with routes which created `.js` files for analytics purposes and it gets removed from the build, too. I guess if it's not in a `/[lang]/` path it gets ignored if the setting is `prefix-always`.

@itsmatteomanf 

Yeah, that's the idea of the routing strategy `prefix-always`. Even endpoints fall under the same logic.

Are there use cases where 404 pages need to be translated? If not, we could safely assume that 404 pages and 500 pages can always be rendered at the root.

itsmatteomanf:
@ematipico 

> So, how do you think we should make sure that all routes have the prefix in their prefix while having **some** routes be exonerated from the routing logic?

I'd personally avoid getting involved with any of the page routing decisions, bar the redirects which are added according to the choices in the config. You could expose a warning for the pages without a path locale, which can be silenced with some config option...

I presume given this discussion you can't do just a single nested localised folder.

ematipico:
Thank you @itsmatteomanf, for your answer and your point of view! Yeah, routing is a soft subject and not easy to get right. 

As for now, we internally decided to allow-list all endpoints: https://github.com/withastro/roadmap/pull/734/commits/6954fb224f162911bdd2f46fc885344572c3669f

In case something changes, I will update the RFC. 

I am not sure this is the right way to go regarding your suggestion. You seem to be an experienced developer, and you know your way around Astro and the subject so a warning can be enough, although I fear that this won't be sufficient for beginners or less experienced on the subject.

itsmatteomanf:
@ematipico yeah, you could also go the other way around, default to ignoring all root things and then allow users to chose with a config option. Easier for beginners, but must be documented, as not all root things get ignored depending on where they are put and when they are generated in the Astro build process.

sandstrom:
The new i18n stuff is great!

I've read all the docs, and this thread. We're considering switching to Astro, and a few i18n-related things that I still haven't figured out how to do in Astro are:

- Support a nested directory structure with translations (arbitrary groups, or namespaces). See example below.
  - Some of them would just plainly ignore the structure and simply glob all `*.yml` files, and then inside every file you'd have a top-level key for the locale, so e.g. `pages/about/en.yml` would contain the YAML-key `"en.about.heading": "Hello World"`. That way, the directory structure is just an arbitrary organisation mechanism (a very useful one though!).
- Controlling which translations go into the JS layer, to be available to e.g. a `t('my.key')` translation helper in a JS component. Most should be server-side only, when used only for static content.
- The dance of `let lang = getLangFromUrl(Astro.url); let t = useTranslations(lang);` seems complicated, would be nicer if I could simply call `t` in my templates, and it would "figure things out".
- Other frameworks have the pattern of `pages/index.html` and `pages/index.dk.html` where `index.html` would have to use the `t` helper, but the file `index.dk.html` would be "pre-translated". You can then mix and match, so most files might use locale-data and `t` helper, but some might be pre-translated (for example a long privacy policy document).


### Locales folder example

```
// example structure for our `src/locales` folder
.
├── components
│   ├── de.yml
│   ├── en.yml
│   └── fr.yml
├── dict
│   ├── de.yml
│   ├── en.yml
│   └── se.yml
├── layout
│   ├── de.yml
│   ├── en.yml
│   └── fr.yml
└── pages
    ├── about
    │   ├── careers
    │   │   ├── en.yml
    │   │   └── fr.yml
    │   └── press
    │       ├── de.yml
    │       ├── en.yml
    │       └── fr.yml
    ├── benefits
    │   ├── de.yml
    │   ├── en.yml
    │   └── fr.yml
    ├── schedule
    │   ├── en.yml
    │   └── fr.yml
    ├── features
    │   ├── users
    │   │   ├── de.yml
    │   │   ├── en.yml
    │   │   └── fr.yml
    │   ├── calendar
    │   │   ├── de.yml
    │   │   ├── en.yml
    │   │   └── fr.yml
```

ematipico:
Added `currentLocale`: https://github.com/withastro/roadmap/pull/734/commits/f934880ab5835492e56ddbb69e3f32662310c8bf

ematipico:
We landed to the conclusion that routing strategies should be applied only to **pages**: https://github.com/withastro/roadmap/pull/734/commits/c93015d44a8dc9e0794c4067899d8f587aaabf47

itsmatteomanf:
@ematipico 

> We landed to the conclusion that routing strategies should be applied only to **pages**: [c93015d](https://github.com/withastro/roadmap/commit/c93015d44a8dc9e0794c4067899d8f587aaabf47)

This would solve most of the issues, but the routing is still pretty messed up which makes it hard to check. Especially the fallback part... I bet you saw the issues I opened.

ematipico:
Added more information about domain support: https://github.com/withastro/roadmap/pull/734/commits/713fce74bb742f18ad985055758c3818b5eed9aa

ematipico:
> @ematipico
> 
> > We landed to the conclusion that routing strategies should be applied only to **pages**: [c93015d](https://github.com/withastro/roadmap/commit/c93015d44a8dc9e0794c4067899d8f587aaabf47)
> 
> This would solve most of the issues, but the routing is still pretty messed up which makes it hard to check. Especially the fallback part... I bet you saw the issues I opened.

I'll get to those issues, don't worry. It's completely fine to have issues during the experimental phase, and I thank you very much for being so adventurous to try this new feature.

itsmatteomanf:
> > @ematipico
> > > We landed to the conclusion that routing strategies should be applied only to **pages**: [c93015d](https://github.com/withastro/roadmap/commit/c93015d44a8dc9e0794c4067899d8f587aaabf47)
> > 
> > 
> > This would solve most of the issues, but the routing is still pretty messed up which makes it hard to check. Especially the fallback part... I bet you saw the issues I opened.
> 
> I'll get to those issues, don't worry. It's completely fine to have issues during the experimental phase, and I thank you very much for being so adventurous to try this new feature.

I was mentioning those as they prevent a lot of testing as they fail in many cases, but I know you'll get there :)

itsmatteomanf:
Oh, I might be adding a feature request or consideration at least.

If you want to add these [tags](https://developers.google.com/search/docs/specialty/international/localized-versions#html), which are the i18n standard way of doing things you need all links, which you can get with the current APIs, but it becomes a pain to match those to the language that generated it.

Returning an array of value pairs or an object would be great. Mostly so you know which link corresponds to which locale.

proposals/0041-i18n-routing.md
```diff
@@ -0,0 +1,373 @@
...
+### Browser locales detection 
+
+In SSR, Astro is able to parse the `Accept-Language` header and provide a list of preferred locales by the user **and** supported by the application.
+
+This list is sorted by the highest to the lowest using the [quality value](https://developer.mozilla.org/en-US/docs/Glossary/Quality_values).
+
+This information is available through the global object `Astro`: 
+- `Astro.preferredLocaleList: string[] | undefined`
+    
+  For example, if `i18n.locals` contains `['pt', 'fr', 'de']`, and the value of the `Accept-Header` value is `en, fr;q=0.2, de;q=0.8, *;q=0.5`, then
```
itsmatteomanf, line 242: Small typo :)

```suggestion
  For example, if `i18n.locales` contains `['pt', 'fr', 'de']`, and the value of the `Accept-Header` value is `en, fr;q=0.2, de;q=0.8, *;q=0.5`, then 
```

jlarmstrongiv:
I’ve seen several comments ([1](https://github.com/jlarmstrongiv/astro-i18n-aut/issues/34#issuecomment-1817425747), [2](https://github.com/withastro/roadmap/pull/734#issuecomment-1804816701), [3](https://discord.com/channels/830184174198718474/1174065571256606720), [4](https://github.com/withastro/astro/issues/8241#issuecomment-1817458843), etc.) requesting that an Astro file render all locales 

```
.
└── pages/
    ├── about.astro
    └── index.astro
```

instead of having Astro files for every locale

```
.
└── pages/
    ├── about.astro
    ├── index.astro
    ├── es/
    │   ├── about.astro
    │   └── index.astro
    └── fr/
        ├── about.astro
        └── index.astro
```

to avoid code duplication and fetch data from APIs and content collections

ematipico:
@itsmatteomanf One question about your [comment](https://github.com/withastro/roadmap/pull/734#issuecomment-1804920792):

```js
{
	locales: [
		{
			path: 'it',
			default: "it", // this is the one you want whenever you have to chose like a single value for the specific locale 
			lang: [ "it", "it-CH", "it-IT",  "it-SM",  "it-VA"] // these you use for like canonicals, matching routes, etc.
		}
	]
}
```

What's the different between `path` and `default`? How should I use `default` and in which context?

itsmatteomanf:
@ematipico 

> What's the different between `path` and `default`? How should I use `default` and in which context?

`path` is the actual URL path `example.com/[lang]/blog`, `default` is the default locale, it could be like `it-CH`, in that example.

It's like that so that you know which is the canonical (if it were just one language) or which is the locale you set your HTML to, e.g. `<html lang="<default>">...</html>`.

ematipico:
Thank you @itsmatteomanf. Now that I understand what's meant for, maybe we can directly use the `lang` array. Since it's an array, we can leverage the order (from left to right), so we can also deliver a sense of priority of the languages needed by the user. This would be inline with how the `Accept-Language` header works.

itsmatteomanf:
> Since it's an array, we can leverage the order (from left to right), so we can also deliver a sense of priority of the languages needed by the user.

Fine by me! I'm almost sure I wrote it somewhere, but I guess I didn't ahah

Maybe do expose that with a specific endpoint, just to make it clearer and easier, but not sure what scope and how big the APIs will be. It's kinda easy to do, anyway.

sandstrom:
Some more feedback (you may have thought about it already).

### Count and plural forms

English has two plural forms, but many languages have several. This is generally handled via a `count` param to the translation helper, e.g. `t('login.status', { count: myInt });` and the following translation data:

```yaml
login:
  heading: "Login statistics"
  status: 
    # some languages would have additional plural forms here
    one: "Last login ${count} day ago"
    other: "Last login ${count} days ago"
    
```

### File formats

Instead of having `.astro` files, I'd recommend something like `json` or `yaml`.

Take some inspiration from the large localisation platforms:

- https://support.phrase.com/hc/en-us/articles/5709604147100#UUID-839bf492-c753-2f54-2062-abb1d360aa14
- https://help.transifex.com/en/articles/6223189-yaml#h_9ad434762f
- https://localazy.com/docs/cli/yaml-format

stefanprobst:
how would a language-switcher look like with the current `astro:i18n` api?

ematipico:
> how would a language-switcher look like with the current `astro:i18n` api?

Depending on your environment, use the list APIs: https://docs.astro.build/en/guides/internationalization/#getrelativelocaleurllist

Then map the result into a new array where you add the label of each locale. Then you should have all the information for a drop-down.

ematipico:
In https://github.com/withastro/roadmap/pull/734/commits/089253d317e4327ee4e3b93eb303ddc28859952b, I removed `routingStrategy` in favour of:

```js
{
	routing: {
		strategy: "pathname",
		prefixDefaultLocale: true
	}
}
```

As discussed and proposed in https://github.com/withastro/roadmap/pull/734#issuecomment-1799126771

Also, myself and Chris had a chat, we changed the shape slightly to align it with the current one, and take advantage of the zod `.transform` function.

itsmatteomanf:
Looks good to me, @ematipico.

The only thing I don't see, but may be there just not explained fully, is a way to get alla locales, in the more granular configuration option, with URLs, even if they are repeated. 

This is very useful for the link alternate i18n configurations (and possibly even for the sitemap plugin).

ematipico:
> Looks good to me, @ematipico.
> 
> The only thing I don't see, but may be there just not explained fully, is a way to get all locales, in the more granular configuration option, with their URLs, even if they are repeated.
> 
> This is very useful for the link alternate i18n configurations (and possibly even for the sitemap plugin).

I'll be happy to add such APIs, could you propose something with the expected result? It would help to get the API right easily

itsmatteomanf:
> > Looks good to me, @ematipico.
> > The only thing I don't see, but may be there just not explained fully, is a way to get all locales, in the more granular configuration option, with their URLs, even if they are repeated.
> > This is very useful for the link alternate i18n configurations (and possibly even for the sitemap plugin).
> 
> I'll be happy to add such APIs, could you propose something with the expected result? It would help to get the API right easily

Sorry, was on mobile...

Given a config file like this:

```js
// astro.config.mjs
import {defineConfig} from "astro/config"
export default defineConfig({
    i18n: {
        defaultLocale: 'en',
        locales: ['en', 'es', 'fr', {
            path: "portugues",
            codes: ["pt", "pt-BR", "pt-AO"]
        }]
    }
})
```

Return something like this:

```json
[
    ["en", "/blog"],
    ["es", "/es/blog"],
    ["fr", "/fr/blog"],
    ["pt", "/portugues/blog"],
    ["pt-BR", "/portugues/blog"],
    ["pt-AO", "/portugues/blog"]
]
```

Two issues I see, doing this. The current `getRelativeLocaleUrlList()` and `getAbsoluteLocaleUrlList()` don't associate a locale with the URL, making it hard to use in a select with a value or assign a label to the URL (how do I know which is first? I know it's probably the order I defined them, but it's not stored anywhere, apart from the config file), and that would apply here, too.

Maybe an array of objects here, e.g. `{ locale: "pt-BR", path: "portugues", url: "/portugues/blog" }`, is better?

And probably the same or similar in the other cases? 🤔

ematipico:
I'll have to think about it, and we can definitely land a new API after we merge the RFC, there's nothing to prevent us from doing so.

zanhk:
@ematipico Would be nice to have a feedback about code duplication (https://github.com/withastro/roadmap/pull/734#issuecomment-1817750144) as from a quick look on the RFC it's not clear to me how, after the implementation, the projects files should be structured,  (I probably missed between the comments) is there a sample repo to showcase how the page structure will look like?

Thanks

ematipico:
> @ematipico Would be nice to have a feedback about code duplication ([#734 (comment)](https://github.com/withastro/roadmap/pull/734#issuecomment-1817750144)) as from a quick look on the RFC it's not clear to me how, after the implementation, the projects files should be structured, (I probably missed between the comments) is there a sample repo to showcase how the page structure will look like?
> 
> Thanks

The RFC is meant for **routing** and fallback, not content (labels, dictionaries, etc.). We wanted to keep the first round of implementation small and provide as many utilises as possible so the community can build around them.  Let's remember that nothing prevents us from adding more features. In fact, we already have a few features that we will add after this first RFC is merged,

The fallback system is one of the most significant changes we added because it adds patterns that weren't possible in Astro before. Once we deem that feature **stable** enough, we could use the same logic to generate pages to avoid code duplication.

proposals/0041-i18n-routing.md
```diff
@@ -0,0 +1,389 @@
...
+Here's a list of APIs available to users to retrieve information:
+
+#### `getRelativeLocaleUrl(locale: string, path: string, options?: Options): string`
+
+Given a locale, the function will return the **relative** URL, without the website at the beginning. The function respects the configurations `base`, `trailingSlash` and `build.format`.
+
+```astro
+---
+// src/pages/index.astro
+import { getRelativeLocaleUrl } from "astro:18";
```
Princesseuh, line 140:
```suggestion
import { getRelativeLocaleUrl } from "astro:i18n";
```

proposals/0041-i18n-routing.md
```diff
@@ -0,0 +1,389 @@
...
+console.log(getRelativeLocaleUrl('es', "")) // will log "/docs/es"
+---
+```
+
+#### `getAbsoluteLocaleUrl(locale: string, path: string, options: Options): string`
+
+Given a locale, the function will return the **absolute** URL, taking into account the do supported. The function respects the configurations `base`, `site`, `trailingSlash` and `build.format`.
```
Princesseuh, line 169: > taking into account the do supported

Might be missing a word perhaps?

bluwy:
Oops didn't notice this is merged, but anyways my comments aren't urgent, just some questions that I wonder if it's a typo or not.

proposals/0041-i18n-routing.md
```diff
@@ -0,0 +1,389 @@
...
+
+In SSR, Astro is able to parse the `Accept-Language` header and provide a list of preferred locales by the user **and** supported by the application.
+
+This list is sorted by the highest to the lowest using the [quality value](https://developer.mozilla.org/en-US/docs/Glossary/Quality_values).
+
+This information is available through the global object `Astro`: 
+- `Astro.preferredLocaleList: string[] | undefined`
+    
+  For example, if `i18n.locales` contains `['pt', 'fr', 'de']`, and the value of the `Accept-Header` value is `en, fr;q=0.2, de;q=0.8, *;q=0.5`, then 
+  `Astro.preferredLocaleList` will be `['de', 'fr']` because `pt` isn't inside the header, and `en` isn't supported by the website. `de` comes first because it has a highest quality value.
```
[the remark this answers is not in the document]
ematipico: Just noticed now, the description isn't correct, I will fix it.

eric-burel:
Hey folks,

I am quite late to the party but I am currently discovering Astro, and I got a good knowledge of Next.js behavior and recent changes regarding i18n, I'd like to send some feedback after spending ~1h playing with Astro i18n setup and reading the [guide](https://docs.astro.build/en/guides/internationalization/) :

- the routing strategy is not super clearly documented yet, though the [browser redirection section](https://docs.astro.build/en/guides/internationalization/#browser-language-detection) seems to implicitly mention that it's based on the Accept-Language header, and URL. So Astro doesn't seem to handle cookie based language selection with no URL param (which is ok but would deserve to be said explicitly)

- the routing seems to rely on middlewares to be able to redirect users, yet doesn't seem to imply per-request SSR as user-land middleware do => it seems to have been mentioned above, but it feels like i18n routing is using an internal feature that could useful to other use cases.

- can we set a generic [locale] parameterrather than explicit folders en, fr etc.? My site is highly dynamic so I only have a big virtual [...path] route that handles all routes of the site. I'll dig again, that's perhaps an issue on my side, but I couldn't make it work in a few dozens minutes

- can we access Astro.preferredLocale without setting up i18n routing? This would allow to easily setup a client-side redirection, without having to explicitely read cookies and activate per-request rendering

Astro made a quite opinionated choice here, with a solution quite similar how Next.js worked before version 12. 

However since that, Next.js has started considering i18n routing as a part of the broader family of personnalization via redirection. User-land middlewares, which are meant to run also for static pages, allow devs to get full control of i18n. 

My feeling is that this is the inevitable direction that all frameworks are meant to take at some point. In Astro, to implement i18n routing, you seem to use an internal feature allowing to have middleware that can run for each HTTP request but also preserve the static nature of the page. Making this feature available to end user would allow them to take full control over i18n instead of relying on the framework provided implementation, but also cover similar personalization use cases.

Issue #709, opened 2023-09-19
Integration hooks for middleware


matthewp:
- Accepted Date: 2023-09-19
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/692
- Author: @matthewp
- Champion(s): @matthewp
- Implementation PR: 

# Summary

A brief, one or two sentence explanation of the proposal.

Add an API to integrations which allows them to define middleware so that users do not need to go through the boilerplate of creating a middleware file and importing the functionality.

# Background & Motivation

When we built the middleware API we anticipated a need for integrations to hook into it, but we decided not to make that a goal for v1 due to the complexity of ordering. Yet the need keeps coming up for the same reasons as before, integrations want to transparently add middleware.

# Goals

- The ability to get middleware into an Astro project simply by adding an integration as normal.
- Allow an integration to place itself either at the front or end of the stack in terms of ordering.

# Non-Goals 

- Not aiming for full control over middleware ordering at this stage. Middleware ordering is determined dynamically as the root middleware module is defined, so an API that allowed integration control over that would be difficult and more broad than in this proposal. There is no known use-case for this regardless.

jlarmstrongiv:
Great!  I currently [generate](https://github.com/jlarmstrongiv/astro-i18n-aut/blob/main/src/integration/ensureValidConfigs/ensureValidRedirectMiddleware.ts#L47) the middleware file if it doesn’t exist.  It would mean users no longer have to [set up](https://github.com/jlarmstrongiv/astro-i18n-aut/tree/main#configure) their own middleware and sequence either.  Looking forward to this update 👍

florian-lefebvre:
@matthewp can this issue be closed?

matthewp:
Yes it can, thanks for the reminder!

jlarmstrongiv:
Thank you @matthewp I just updated [`astro-i18n-aut`](https://www.npmjs.com/package/astro-i18n-aut) to use the new integration hooks—it works great 🎉

PR #735, opened 2023-10-19
Integration hooks for middleware

matthewp:
# Summary

Add the ability for integrations to define middleware on behalf of the application developer.

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/integration-middleware/proposals/middleware-in-integrations.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/709)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/692)

### DIFF
--- a/proposals/middleware-in-integrations.md
+++ b/proposals/middleware-in-integrations.md
@@ -98,6 +98,12 @@ export const onRequest = sequence(
 );
 ```
 
+### Runtime debugging
+
+Existing middleware users might find integrations hooking into their pipeline to be unexpected, therefore we will add a runtime debugging logging when we detect that both integrations and the application developer have added middleware.
+
+The debugging logging will include which other the integration was added to help with debugging unexpected behavior.
+
 # Testing Strategy
 
 The existing middleware tests, which are mostly fixture based, can be reused here. We test other integrations in this way and it works well. Just create some integrations for different scenarios such as:
--- a/proposals/middleware-in-integrations.md
+++ b/proposals/middleware-in-integrations.md
@@ -1,0 +1,122 @@
+- Start Date: 2023-10-19
+- Reference Issues: https://github.com/withastro/roadmap/issues/709
+- Implementation PR: https://github.com/withastro/astro/pull/8869
+
+# Summary
+
+Allow integration to add middleware without the application developer needing to define a `src/middleware.{js,ts}` file.
+
+# Example
+
+This adds a new integration API, a function in the `astro:config:setup` hook:
+
+```js
+export function myIntegration() {
+  return {
+    name: 'my-integration',
+    hooks: {
+      'astro:config:setup': ({ addMiddleware }) => {
+        addMiddleware({
+          entrypoint: '@my-package/middleware',
+          order: 'pre'
+        });
+      }
+    }
+  };
+}
+```
+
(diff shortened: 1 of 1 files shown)


natemoo-re:
Very excited for this one! Proposed API looks good. Any thoughts on using [**`enforce`**](https://vitejs.dev/guide/api-plugin.html#plugin-ordering) instead of **`order`** to match Vite's existing naming convention?

matthewp:
@natemoo-re Yeah, could do that. I went with `order` to match Rollup actually: https://rollupjs.org/plugin-development/#build-hooks but I'm fine with either one. ccing @bluwy the naming expert. Will also put this in the `alternatives` section.

natemoo-re:
Ah good call. I wasn't aware that Rollup went with a different name. Does Vite plan to move towards `order`?

ematipico:
Overall, this is a good addition, and I like it very much. Some concerns that I voiced already in private, it would be great if we could at least add some logging **when** an integration **attempts** to add a middleware. This can be a `logger.debug` (no need to flood the console).

bluwy:
> Ah good call. I wasn't aware that Rollup went with a different name. Does Vite plan to move towards `order`?

Yeah Vite would move towards `order` eventually, but it's a big breaking change at the moment. I'd also recommend following Rollup's convention.

And the proposal looks good to me 👍

natemoo-re (approved):
LGTM! Excited for this one to land.

proposals/middleware-in-integrations.md
```diff
@@ -0,0 +1,128 @@
...
+Once the `settings` object contains the ordered list of middleware, a Vite plugin will be used to create a single middleware `onRequest` function using the `sequence` API. The pseudo-code of what this plugin produces looks like so:
+
+```js
+import { sequence } from 'astro:middleware';
+import { onRequest as userOnRequest } from '/src/middleware';
+
+import { onRequest as _pre_0 } from 'one';
+import { onRequest as _pre_1 } from 'two';
+
+import { onRequest as _post_0 } from 'one';
```
lilnasy, line 91: The same middleware runs twice per request. When might this be necessary?

lilnasy: Is "both" a valid value for "order"?

lilnasy: Seems like the example might be unintentional.

matthewp: No, both is not a valid value. This is just showing that an integration adds both a pre and a post. You have to add one or the other. I can make this be a different integration to make that clearer.

lilnasy: Do you think the `astro:config:setup` hook example might be a better place to show `addMiddleware` being called multiple times?

ematipico (approved):

proposals/middleware-in-integrations.md
matthewp: This example is implementation details. The idea is to lay out how you think the feature should be implemented so that someone could take the information and implement it. So it needs to be hear to convey the expected compile output.

matthewp:
I'm moving for a call for consensus on this RFC. This will be the final comment period (3 days); if there are no objections this will be merged and the feature can be released in a future release (likely 3.5)

bluwy (approved):

Issue #754, opened 2023-11-01
Support prefetch in core


bluwy:
- Accepted Date: 2023-11-01
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/749
- Author: @bluwy
- Champion(s): @bluwy
- Implementation PR: https://github.com/withastro/astro/pull/8951

# Summary

Deprecate `@astrojs/prefetch` and provide first-class prefetch support in Astro.

# Background & Motivation

With the introduction of View Transitions, it includes partial prefetching code for snappy navigation between pages. We can take this opportunity to support prefetching in core, and share the prefetch behaviour with View Transitions. 

I've started an implementation before an RFC as the initial plan was to simply move `@astrojs/prefetch` to core. However, it would also be a good time to polish up and extend the API.

# Goals

- An option to enable prefetching
- Enable prefetching via an attribute/hint
- Enable prefetching for all links by default (required by View transitions)
- Disable prefetching if all links are enabled by default
- Different prefetching strategies (click, hover, viewport, etc)
- Only add JS if using prefetching

# Non-Goals

- Prefetch cache invalidation (Browser relies on cache control)
- Prefetch external links

NOTE: Stage 2 does not mention solution details, but I've written one in the discussion: https://github.com/withastro/astro/pull/8951

bluwy:
Closing as the stage 3 RFC has been merged.

PR #755, opened 2023-11-01
Support prefetch in core

bluwy:
# Summary

Deprecate `@astrojs/prefetch` and provide first-class prefetch support in Astro.

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/prefetch-core/proposals/0043-prefetch.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/754)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/749)


proposals/0043-prefetch.md
```diff
@@ -0,0 +1,142 @@
...
+- Only add JS if using prefetching
+
+# Non-Goals
+
+- Prefetch cache invalidation (Browser relies on cache control)
+- Prefetch external links
+
+# Detailed Design
+
+A prefetch script for client-side is required. It should only be included if `prefetch` is truthy. The script can be injected through the `injectScript` integration API.
```
matthewp, line 79: > It should only be included if `prefetch` is truthy.

Do you mean the config option? I think this isn't true if so, it can also be included if someone uses the programmatic API, right?

bluwy: Ah yup. Will clarify that. Now that you mention about the programmatic API, the user would have to enable the `prefetch` config first before they can import it.

I've not covered this error handling if not, so I should also fix that in the implementation PR.

bluwy: Actually, ignore the latter 😅  You can import `prefetch` from `astro:prefetch` without the `prefetch: true` config with the current implementation I think, since it's standalone manual prefetch and will not setup any listeners.  But I think it's better to only allow importing `astro:prefetch` if `prefetch: true` is enabled so it's consistent with our other APIs before.

proposals/0043-prefetch.md
```diff
@@ -0,0 +1,147 @@
...
+
+```ts
+export declare function prefetch(url: string, opts?: { with?: 'link' | 'fetch' }): void
+```
+
+This module can only be imported if the `prefetch` config is enabled.
+
+# Testing Strategy
+
+End-to-end tests, make sure user tap/hover/viewport all workks.
```
matthewp, line 127:
```suggestion
End-to-end tests, make sure user tap/hover/viewport all works.
```

proposals/0043-prefetch.md
```diff
@@ -0,0 +1,147 @@
...
+Different prefetch strategies have different behaviours and opinions:
+- `tap`: Calls `fetch()` on `touchstart` or `mousedown` events (they are called before `click` event)
+- `hover`: Calls `fetch()` on `focusin`+`focusout` or `mouseenter`+`mouseleave` events. Hover detection kicks in after 80ms delay.
+- `viewport`: Creates a `<link rel="prefetch">` on enter viewport. It has lower priority than `fetch()` to not clog up the request. Intersection detection kicks in after 300ms.
+
+Notes: `hover` and `viewport` only works on `<a />` tags on initial page load (and view transition page load) due to limitations of the events. Unless we watch the entire DOM with MutationObserver but it's not performant.
+
+---
+
+For programmatic usage (only if `prefetch` config is enabled):
```
matthewp, line 49: Should we have a warning if someone attempts to use the programmatic API without having enabled it in their config?

matthewp: Should make sure this doesn't happen for the ViewTransitions case.

proposals/0043-prefetch.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Programmatic API
+
+The client script would have an internal `prefetch` function that we can expose to the `astro:prefetch` module:
+
+```ts
+export declare function prefetch(url: string, opts?: { with?: 'link' | 'fetch' }): void
+```
+
+This module can only be imported if the `prefetch` config is enabled.
```
ematipico, line 123: It's unclear how the API works and how `link` and `fetch` change the behaviour of said API. 
For example, what the `url` stand for? Does it support external links? I'd like to see an example and a brief explanation of what happens:

```js
prefetch("/path");
prefetch("http://example.com");
prefetch("http://example.com", { with: "link" });
prefetch("http://example.com", { with: "fetch" });
```

FYI I don't know how the the current integration `@astrojs/prefetch` works.

proposals/0043-prefetch.md
```diff
@@ -0,0 +1,147 @@
...
+- Double prefetching could happen as View Transitions prefetches too
+
+# Alternatives
+
+Continue supporting `@astrojs/prefetch`
+
+# Adoption strategy
+
+We should document a migration path for `@astrojs/prefetch` users to use the new `prefetch` option.
+
```
ematipico, line 142: Don't we plan to have it as experimental first?

proposals/0043-prefetch.md
bluwy: Good callout. I actually documented it better in the implementation jsdoc now, but will port the explanation here.

proposals/0043-prefetch.md
bluwy: I think the plan is to ship it stable immediately. Especially that Astro 4.0 is near if we need to make any change, and that View Transitions that uses this internally is also stable too.

bluwy:
Calling for consensus for this RFC. While it's already shipped in Astro 3.5, we're treating it as an unflagged experimental feature for now and are still open for final feedback around the API. The period will last for 3 days and if there's no objections, this RFC will be merged.

bluwy:
Merging this as it has passed 3 days! Thanks for the discussion everyone.

Issue #698, opened 2023-09-13
Incremental Builds


natemoo-re:
- Accepted Date: 2023-09-13
- Reference Issues/Discussions:
  - https://github.com/withastro/roadmap/discussions/237
- Author: @natemoo-re 
- Champion(s): @natemoo-re 
- Implementation PR: 

# Summary

Incremental Build support in Astro aims to significantly speed up repeated runs of `astro build`. Initial builds will populate some kind of cache, which will allow subsequent builds to bypass Rollup for unchanged trees of the module graph.

# Background & Motivation

The original [incremental build support](https://github.com/withastro/roadmap/discussions/237) proposal is one of our oldest and most highly upvoted issues. Unfortunately, users have not shared many concrete examples in that thread. However, from the extensive performance profiling we've done, we know that Rollup is the main bottleneck in Astro's build process. 

Now that our rendering, Markdown, and MDX performance has been optimized about as far as we can take it, now is the time to explore new options. The best option we have is to move _as much of our build process out of Rollup_ as possible, likely with some sort of build cache. That is the goal of introducing **incremental builds**.

Why now? The reason is straightforward—caching is notoriously difficult to get right. We did not want to take on the additional complexity until our API surface was stable and our internal build process was easier to reason about. Thanks to significant effort from @bluwy and @ematipico in the past few months, we're now in a good place to tackle this.

# Goals
 
- Improve `astro build` performance
- Avoid as much repeated processing during `astro build` as possible
- Implement a generic API that enables incremental builds
- Eventually, enable incremental builds by default
- Support existing `static`, `hybrid`, and `server` outputs

# Non-Goals 

- Vendor lock-in. Incremental builds will be implemented generically, supporting our existing ecosystem of deployment platforms where possible. If a host caches assets between builds, it is likely that they will support incremental builds automatically.
- Incremental Static Regeneration (also known as [ISR](https://nextjs.org/docs/pages/building-your-application/data-fetching/incremental-static-regeneration) or [DPR](https://www.netlify.com/blog/2021/04/14/distributed-persistent-rendering-a-new-jamstack-approach-for-faster-builds/)). The proposal for [supporting ISR](https://github.com/withastro/roadmap/discussions/228) is an entirely different topic, not covered by this accepted proposal. **These features are not mutually exclusive.** Implementing incremental build support **benefits every user of Astro** and does not prevent Astro from potentially introducing ISR in the future.
[...]

PR #763, opened 2023-11-08
Incremental Build RFC

natemoo-re:
# Summary

Built-in incremental build support

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/incremental-builds/proposals/0044-incremental-build.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/698)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/237)

### DIFF
--- a/proposals/0044-incremental-build.md
+++ b/proposals/0044-incremental-build.md
@@ -1,0 +1,95 @@
+- Start Date: 2023-11-08
+- Reference Issues: https://github.com/withastro/roadmap/issues/698
+- Implementation PR: https://github.com/withastro/astro/pull/8854
+
+# Summary
+
+Incremental Build support in Astro aims to significantly speed up repeated runs of `astro build`. Initial builds will populate a local cache in the [`cacheDir`](https://docs.astro.build/en/reference/configuration-reference/#cachedir), which will allow subsequent builds to bypass Rollup for unchanged trees of the module graph.
+
+# Example
+
+This propsal introduces minimal public API surface changes, but requires many internal implementation details to be updated to support caching in some way.
+
+### Adds a new `cache` option in the Astro config
+
+```js
+// astro.config.mjs
+export default defineConfig({
+  build: { cache: true },
+});
+```
+
+### Adds a new `--force` flag to the CLI to bypass the build cache
+
+```js
+astro build --force
+```
+
+# Background & Motivation
+
+The original [incremental build support](https://github.com/withastro/roadmap/discussions/237) proposal is one of our oldest and most highly upvoted issues. The corresponding [roadmap issue](https://github.com/withastro/roadmap/issues/698) has likewise recieved a lot of attention and positive feedback. From the extensive performance profiling we've done, we know that Rollup is the main bottleneck in Astro's build process.
+
+Now that our rendering, Markdown, and MDX performance has been optimized about as far as we can take it, now is the time to explore new options. The best option we have is to move _as much of our build process out of Rollup_ as possible, with as much of the build as possible being restored from a cache. That is the goal of introducing **incremental builds**.
+
+Why now? The reason is straightforward—caching is notoriously difficult to get right. We did not want to take on the additional complexity until our API surface was stable and our internal build process was easier to reason about. Thanks to significant effort from @bluwy and @ematipico in the past few months, we're now in a good place to tackle this.
+
+# Goals
+
+- Improve `astro build` performance
+- Restructure our
+- Avoid as much repeated processing during `astro build` as possible
+- Implement a configurable config flag that enables incremental builds
+- Eventually, enable incremental builds by default
+
+# Non-Goals
+
+- Vendor lock-in. Incremental builds will be implemented generically, supporting our existing ecosystem of deployment platforms where possible. If a host caches assets between builds, it is likely that they will support incremental builds automatically.
+- Incremental Static Regeneration (also known as [ISR](https://nextjs.org/docs/pages/building-your-application/data-fetching/incremental-static-regeneration) or [DPR](https://www.netlify.com/blog/2021/04/14/distributed-persistent-rendering-a-new-jamstack-approach-for-faster-builds/)). The proposal for [supporting ISR](https://github.com/withastro/roadmap/discussions/228) is an entirely different topic, not covered by this accepted proposal. **These features are not mutually exclusive.** Implementing incremental build support **benefits every user of Astro** and does not prevent Astro from potentially introducing ISR in the future.
+- Future: Adapter API. Some adapters perform a significant amount of processing and may also want some form of incremental build support. To reduce the scope of this proposal, we are not considering exposing a public Adapter API for this. This may be implemented in the future as an enhancement.
+
+# Detailed Design
+
+TODO
+
+# Testing Strategy
+
+- Integration tests will be needed to determine that the cached behavior is consistent with the existing build process
+- Unit tests might be relevant for certain parts of this feature, such as the module graph or how individual modules are invalidated
+
+# Drawbacks
+
+- Caching, especially cache invalidation, is really hard. Invalid cache hits could lead to hard to debug problems and ultimately to a negative impression of Astro.
+- There could be some integrations that unknowingly expect a full build to run. This could break implicit behavior in subtle ways.
+- Code complexity. Our build process is already complex, and adding a caching layer will certainly add additional complexity and consideration for future features.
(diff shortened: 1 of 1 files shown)


montsamu:
One small datapoint in favor of the currently experimental content cache:

Audiobookaneers.com build:

21:10:55 [build] 5904 page(s) built in 170.55s

21:13:36 [build] 5904 page(s) built in 118.22s

Second build was more than 50 seconds (about 30%) faster. I'm hoping to see similar improvements on the Cloudfront side now that they have a "beta" build cache as well, but so far the gains are a bit more minimal (from 2m24s to 1m54s, though this DID include changes to 3 posts content, though not new posts/tags/etc).

In the "best case scenario" when I add a post to /src/content/posts, the following pages would need to be rebuilt:

1. /2024/01/11/some-title/index.html -- the actual new post page
2. /2024/01/11/ index.html -- index of the day's posts
3. /2024/01/11/page/2... -- in unlikely event of more than a page (10) worth of posts for the day, all page/2 page/3 etc. would need to be updated
4. /2024/01/index.html -- index of the month's posts
5. /2024/01/page/2... -- as above, the new post would bump all pages of this paginated set
6. /2024/index.html -- index of the year's posts
7. /2024/page/2... -- as above
8. /index.html -- front page, always shows newest posts
9. /page/2/index.html, /page/3/index.html... on through /page/70/index.html as the new post bumps through the pagination
10. /rss.xml
11. /author/myname/index.html -- archive by author
12. /author/myname/page/2...
13. /category/somecategory/index.html -- archive by category
14. /category/somecategory/page/2...
15. /tag/sometag/index.html -- archive by tag(s)
16. /tag/sometag/page/2...
17. /tag/anothertag...

I tell you what, this has discouraged me from implementing a 'recent posts' sidebar widget into the static build, as that would mean that *every single page on the site* would need to be regenerated. I'll definitely just use client-side JS for that! And it makes me think about investigating the "hybrid" approach for the pagination as this would cut down a lot of what needs to be rebuilt.

Mark-RD:
Build caching idea:

```js
//Return a page hash, unique for each page variation
export function getHash() {
  //return CMS.pageVersion;
  //return CMS.lastUpdateTime;
  //return null; //Always rebuild - Sitemap etc
  return Astro.props; //Default
}

//OR

  return [
    { hash : /* required for partial building, null by default */ ,params: { /* required */ } ,props: { /* optional */ } },
  ];
}
---
TEMPLATE
```

If every page had a function (similar to getStaticPaths) called getHash that produced a unique hash for a particular page, then on a new build a list of the previous hashes and new hashes could be compared and only the ones with different hashes would be built.

This assumes that the output HTML is deterministic based of its inputs, for example the default implementation of getHash would return: sha256(Astro.props).

* Where to get the last array of page hashes? _Build caches native to platforms_
* Where to download the last full build to add the incremental files? _Build caches native to platforms, or adaptors to S3 compatible APIs e.g. Cloudflare R2_
* What if there are no build cache adaptors? _Rebuild the full site_
* What if the build cache is empty or expired? _Rebuild the full site_
* What if the hash array is empty? _Rebuild the full site_

Mark-RD:
* What to do if the SSG is so large it times out current static host builds? _Allow the build cache to be hydrated on a local machine before pushing to host_

proposals/0044-incremental-build.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Adds a new `--force` flag to the CLI to bypass the build cache
+
+```js
+astro build --force
+```
+
+# Background & Motivation
+
+The original [incremental build support](https://github.com/withastro/roadmap/discussions/237) proposal is one of our oldest and most highly upvoted issues. The corresponding [roadmap issue](https://github.com/withastro/roadmap/issues/698) has likewise recieved a lot of attention and positive feedback. From the extensive performance profiling we've done, we know that Rollup is the main bottleneck in Astro's build process.
```
jcayzac, line 30:
```suggestion
The original [incremental build support](https://github.com/withastro/roadmap/discussions/237) proposal is one of our oldest and most highly upvoted issues. The corresponding [roadmap issue](https://github.com/withastro/roadmap/issues/698) has likewise received a lot of attention and positive feedback. From the extensive performance profiling we've done, we know that Rollup is the main bottleneck in Astro's build process.
```

proposals/0044-incremental-build.md
```diff
@@ -0,0 +1,95 @@
...
+
+Now that our rendering, Markdown, and MDX performance has been optimized about as far as we can take it, now is the time to explore new options. The best option we have is to move _as much of our build process out of Rollup_ as possible, with as much of the build as possible being restored from a cache. That is the goal of introducing **incremental builds**.
+
+Why now? The reason is straightforward—caching is notoriously difficult to get right. We did not want to take on the additional complexity until our API surface was stable and our internal build process was easier to reason about. Thanks to significant effort from @bluwy and @ematipico in the past few months, we're now in a good place to tackle this.
+
+# Goals
+
+- Improve `astro build` performance
+- Restructure our
```
jcayzac, line 39:
```suggestion
- Restructure our (something?)
```

proposals/0044-incremental-build.md
```diff
@@ -0,0 +1,95 @@
...
+- Reference Issues: https://github.com/withastro/roadmap/issues/698
+- Implementation PR: https://github.com/withastro/astro/pull/8854
+
+# Summary
+
+Incremental Build support in Astro aims to significantly speed up repeated runs of `astro build`. Initial builds will populate a local cache in the [`cacheDir`](https://docs.astro.build/en/reference/configuration-reference/#cachedir), which will allow subsequent builds to bypass Rollup for unchanged trees of the module graph.
+
+# Example
+
+This propsal introduces minimal public API surface changes, but requires many internal implementation details to be updated to support caching in some way.
```
ferparmur, line 11:
```suggestion
This proposal introduces minimal public API surface changes, but requires many internal implementation details to be updated to support caching in some way.
```

matthewp:
The experimental feature was removed in Astro 5. With the introduction of Content Layer the internals of how Content Collections have changed quite a lot, so how an incremental build could work needs to be rethought.

dmitryshostak:
Incremental builds on demand would be such a game-changer for many large companies and organizations with tens of thousands of static pages.

delucis:
> Do you think its possible to achieve simple ISR using [app.render](https://docs.astro.build/en/reference/adapter-reference/#apprender) in a custom adapter? Would someone be willing to provide guidance for implementing that?

Here are a couple of guides to achieving something like ISR even with the regular adapters: 

- https://logsnag.com/blog/implementing-isr-in-astro
- https://developers.netlify.com/guides/how-to-do-advanced-caching-and-isr-with-astro/

KL13NT:
Thanks for the guides, but that's not exactly what I meant. I've already gone over the articles and they either use cache headers with SSR or stale-while-revalidate, or some Netlify specific stuff. 

What I wish to explore is to build a node adapter (extension? Idk) that would look at incoming requests, determine the requested page, and render it if it has been flagged as revalidate by my revalidate api route. Is this even possible using the app.render() or did I misunderstand the utility?

delucis:
Ah I see. Yeah, I guess that might work? It might also not be needed if you follow a model more like the first guide link I shared. That shows invalidating the cache based on headers, but adding an endpoint that invalidates the cache for a page instead would also work in that model (and might be a bit easier to implement than rebuilding the adapter?)

KL13NT:
That works when you can have a CDN in front of your server, which I believe is enough for the majority of users at this stage of Astro. Thanks for the suggestions. Can't wait for Astro-level ISR tho! 👀

delucis:
> I was wondering if there are any recent updates on this RFC, or if it’s still being actively explored?

This RFC was specific to our old Markdown/MDX build process prior to Astro 5, so no longer applies.

These couple of articles are good ones for how to approach ISR with Astro:

- [“How to Implement Incremental Static Regeneration (ISR) in Astro”](https://logsnag.com/blog/implementing-isr-in-astro) (LogSnag)
- [“How to do ISR and advanced caching with Astro”](https://developers.netlify.com/guides/how-to-do-advanced-caching-and-isr-with-astro/) (Netlify Blog)

Vizards:
> This RFC was specific to our old Markdown/MDX build process prior to Astro 5, so no longer applies.

Really appreciate you pointing me to those articles—I’ll definitely give them a thorough read. In the meantime, I’ve actually been experimenting with leveraging Content Collections together with Server Islands to try and achieve something similar to on-demand ISR. It seems like a promising approach, and I can’t wait to dive deeper and see where it leads.

PR #798, opened 2023-12-21
rfc: i18n routing domain support

ematipico:


proposals/0046-i18n-domain-support.md
```diff
@@ -0,0 +1,62 @@
...
+have localised content under different subdomains or domains. 
+
+# Goals
+
+- Support different domains, all driven by the same Astro project;
+- Mix locales that require a different domain, with locales that don't require a different domain;
+
+# Non-Goals
+
+- Force a redirect to users
```
matthewp, line 23: Can you explain this more? I think what you mean is that it's not going to redirect users from the domain to a path prefix, is that right?

proposals/0046-i18n-domain-support.md
```diff
@@ -0,0 +1,62 @@
...
+})
+```
+
+The following APIs will behave as follows:
+- [`getRelativeLocaleUrl`](#getrelativelocaleurllocale-string-string): it won't prefix the locale to the URL. From `/en` to `/`;
+- [`getAbsoluteLocaleUrl`](#getabsolutelocaleurllocale-string-string): it won't have the locale in the URL: From `example.com/fr` to `fr.example.com/`;
+
+Adapters must have the capabilities to redirect a user from one domain to another based on the domains configured.
+
+An adapter can signal Astro the feature support using the relative configuration:
```
matthewp, line 56: What configuration is that? Are you saying that a feature flag is needed in the adapter?

ematipico:
> In addition to the inline comments, one question I have is whether you have considered it makes sense to support domains and path prefixes at the same time.

If the users use the utilities that we provide to create the URLs, we won't have problems with that.

proposals/0046-i18n-domain-support.md
ematipico: Oh, good catch. This is outdated! We don't need any configuration, although this feature requires the hosting platform to set some specific headers. So I still think we should add an astro feature to the list. If the platform can't provide those headers, the relative adapter can't support domains

proposals/0046-i18n-domain-support.md
ematipico: Yes. Also, if a user lands on a domain different from their browser language, we don't plan to do a redirect.

proposals/0046-i18n-domain-support.md
```diff
@@ -0,0 +1,62 @@
...
+        routingStrategy: "domain"
+    }
+})
+```
+
+The following APIs will behave as follows:
+- [`getRelativeLocaleUrl`](#getrelativelocaleurllocale-string-string): it won't prefix the locale to the URL. From `/en` to `/`;
+- [`getAbsoluteLocaleUrl`](#getabsolutelocaleurllocale-string-string): it won't have the locale in the URL: From `example.com/fr` to `fr.example.com/`;
+
+Adapters must have the capabilities to redirect a user from one domain to another based on the domains configured.
```
Princesseuh, line 54: I'm not sure I understand this line, it's previously said in the non-goals that redirecting users is not part of this?

Princesseuh:
Matthew perfectly covered all the questions I had, so looking forward to the answers to his comments! Looks great, I'm excited to see this land in Astro, our i18n support is really starting to look quite complete, it's an awesome story.

matthewp:
@ematipico 

> If the users use the utilities that we provide to create the URLs, we won't have problems with that.

So if this is supported, would that still fall under the `routingStrategy: "domain"`? If so could you maybe add this to the RFC and explain that it is supported.

proposals/0046-i18n-domain-support.md
ematipico: Yeah, that line is really confusing, and it doesn't add much information. I will remove it. 

> Additionally, will adapters have any specific way of getting access to those domains, or will they read from the config?

The configuration is more than enough, which is an information that they already have. Astro won't do anything fancy with this information

proposals/0046-i18n-domain-support.md
```diff
@@ -0,0 +1,79 @@
...
+
+# Goals
+
+- Support different domains, all driven by the same Astro project;
+- Mix locales that require a different domain, with locales that don't require a different domain;
+
+# Non-Goals
+
+- Redirect users from a path website to a domain website 
+- Change website/domain based on the language of the user's browser
```
bluwy, line 24: In this RFC, do we cover how multi-domain sites get deployed? (Goal/non-goal)

For example, the Vercel adapter could work and redirect to another subdomain, but does that mean Vercel needs to run another build for that subdomain? Or would a single build be able to share for multiple subdomains? (I don't know if Vercel supported something like this ootb)

Also, will this work for static builds? And if so, is there a specific deployment guide for it?

ematipico: No, we don't cover it, and it shouldn't be part of the proposal. Each hosting platform has its own quirks to setup multiple domains (and sub-domains). And redirects too.

The adapters **could** decide to emit the configuration needed to support subdomains, redirects, other domains, etc. although Astro doesn't need to expose anything that comes out from this proposal because it already offers everything that an adapter needs: the new configuration fields and the check for the request headers.

ematipico: > I'm not really sure if the user can easily deploy i18n-subdomain sites on Vercel yet. (or i could be wrong)

You're right! Probably the integration that will work out of the box will be the Node.js adapter

ematipico:
The commit [`cc268a6` (#798)](https://github.com/withastro/roadmap/pull/798/commits/cc268a62b955791131e4504994823b2796e0ce85) adds some restrictions and clarifications based on everybody's feedback

proposals/0046-i18n-domain-support.md
```diff
@@ -0,0 +1,87 @@
...
+- Support for static/hybrid output
+- Offer the means to configure multiple domains for serverless hosts (Vercel, Netlify, etc.)
+
+
+# Detailed Design
+
+There are some restrictions about how the feature will initially work:
+- the feature works only with `output: "server"`, hence only in SSR;
+- the presence of at least one pre-rendered route will cause a failure in the build;
+- the option `base` is required, because Astro needs to know how to create absolute URLs for locales that aren't mapped to a domain;
```
delucis, line 35: Is this required? Couldn’t the default `base` of `'/'` be used for this purpose? (Also noting that the code examples below do not include `base`, so would presumably invalid if this is true.)

proposals/0046-i18n-domain-support.md
```diff
@@ -0,0 +1,87 @@
...
+            fr: "https://fr.example.com",
+            pt: "https://example.pt"
+        },
+        routingStrategy: "domain"
+    }
+})
+```
+
+The following APIs will behave as follows:
+- [`getRelativeLocaleUrl`](#getrelativelocaleurllocale-string-string): it won't prefix the locale to the URL. From `/en` to `/`;
```
delucis, line 59: Wondering if this should even be an error or something in this case? Seems like kind of a footgun if you use `getRelativeLocaleUrl()` hoping to navigate to a different locale, but get a same-domain random path instead.

Also to make sure I follow this. With the example config above, would you get the following behaviour?

```js
getRelativeLocaleUrl('en', '') // => '/en'
getRelativeLocaleUrl('es', '') // => '/es'
getRelativeLocaleUrl('fr', '') // => '/'
```

And is that consistent across domains? i.e. for a page on `fr.example.com`, does `getRelativeLocaleUrl('en', '')` still return `'/en'`? That would resolve to `fr.example.com/en`, which seems kind of useless.

Should sites using multiple domains perhaps _always_ use `getAbsoluteLocaleUrl()`? At the most extreme end, we could even change the type of `astro:i18n` and remove `getRelativeLocaleUrl()` entirely in these cases.

proposals/0046-i18n-domain-support.md
```diff
@@ -0,0 +1,87 @@
...
+- [`getAbsoluteLocaleUrl`](#getabsolutelocaleurllocale-string-string): it won't have the locale in the URL: From `example.com/fr` to `fr.example.com/`;
```
delucis, line 60: Again, for clarification, 

> it won't have the locale in the URL

This only applies to subdomain locales, right? So

```js
getAbsoluteLocaleUrl('en', '') // => 'https://example.com/en'
getAbsoluteLocaleUrl('es', '') // => 'https://example.com/es'
getAbsoluteLocaleUrl('fr', '') // => 'https://fr.example.com/'
```

delucis (commented):
Saw the PTAL @ematipico, so left a few comments/questions! 🙌

natemoo-re (approved):
I'm conceptually onboard with all of this! Hopefully we can lift the prerendering restriction at some point, but I know that's a whole other topic. This is definitely #NWTWWHB!

proposals/0046-i18n-domain-support.md
ematipico: Yes, this only applies to locales configured to have domains. Locales that don't have any domain should still be the same.

proposals/0046-i18n-domain-support.md
ematipico: > And is that consistent across domains? i.e. for a page on fr.example.com, does getRelativeLocaleUrl('en', '') still return '/en'? That would resolve to fr.example.com/en, which seems kind of useless.

No, it's not consistent and it's impossible to make it consistent. 

> Should sites using multiple domains perhaps always use getAbsoluteLocaleUrl()? At the most extreme end, we could even change the type of astro:i18n and remove getRelativeLocaleUrl() entirely in these cases.

Is it something that we can do? 😮 If so, I think we should go this way, although it should make the transition from a non-domain website to a domain website.

Maybe we should do something in the middle; if the user uses a `getRelativeLocale*` function, we should fall back to use `getAbsoluteLocale*` instead. And we could add a warning.

proposals/0046-i18n-domain-support.md
ematipico: Oh, shoot, the wrong option! The required options is `site`! :D

proposals/0046-i18n-domain-support.md
ematipico: Also, I realised that `getRelativeLocaleUrl` should not change its behaviour. I removed that line

delucis: > Maybe we should do something in the middle; if the user uses a `getRelativeLocale*` function, we should fall back to use `getAbsoluteLocale*` instead. And we could add a warning.

Yes, that would be a nice compromise I think and wouldn’t need the type to change (I assume? The two functions have the same signature right?) while still giving feedback that the function used is not intended for this context.

proposals/0046-i18n-domain-support.md
```diff
@@ -0,0 +1,87 @@
...
+    site: "https://example.com",
+    output: "server",
+    i18n: {
+        defaultLocaLe: 'en',
+        locales: ['en', 'es', 'pt_BR', 'pt', 'fr'],
+        domains: {
+            fr: "https://fr.example.com",
+            pt: "https://example.pt"
+        },
+        routingStrategy: "domain"
```
delucis, line 55: I realised that this option may not be needed. Instead, the “flag” for switching to domain mode would simply be providing some `i18n.domains`. I’d guess that’s nicer for users than weird combos like defining `domains` then getting a “wrong routing strategy error” if they didn’t also set that.

(It’s also `routing: { strategy: 'domain' }` in the implementation PR I think?)

delucis (commented):
Thanks for addressing my comments @ematipico! Sounds good to me. 

Posting one more bit of feedback that I already shared on Discord.

ematipico:
FYI, we **won't** need a new `i18n.routing.strategy` variant. I updated the implementation PR.

matthewp (approved):

ematipico:
Call for consensus starts now

matthewp:
The final comment period has ended and this RFC is accepted! It can now be unflagged in Astro.

Issue #811, opened 2024-01-10
 CSRF Protection #471 


ematipico:
- Accepted Date: 2024-01-10
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/471
- Author: @FredKSchott 
- Champion(s): @ematipico 
- Implementation PR: 

# Summary

Provide the infrastructure to protect Astro websites from CSRF attacks

# Background & Motivation

Most background is available here: https://owasp.org/www-community/attacks/csrf

Astro should provide some level of security to users.

# Goals

- Add the required checks to prevent CSRF, probably via an option

# Non-Goals 

- Give the users the possibility to customise the implementation of the protection

ematipico:
Closing. Stage 3: https://github.com/withastro/roadmap/pull/879

PR #879, opened 2024-04-02
rfc: CSRF protection

ematipico:
# Summary

CSRF - Cross-site Request forgery -  protection

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/feat/csrf-protection-rfc/proposals/0046-csrf-protection.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/811)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/471)

### DIFF
--- a/proposals/0046-csrf-protection.md
+++ b/proposals/0046-csrf-protection.md
@@ -24,11 +24,12 @@ Astro should provide some level of security to users.
 
 # Goals
 
-- Add the required checks to prevent CSRF, probably via an option
+- Add the required checks to prevent CSRF, by checking the routes that can be called via `POST`, `PUT`, `PATCH`, or `DELETE` requests.
+- Add the required checks to prevent CSRF in case the application behind a reverse proxy.
 
 # Non-Goals
 
-- Give the users the possibility to customise the implementation of the protection
+- Give the users the possibility to customise the implementation of the protection at route level.
 
 # Detailed Design
 
--- a/proposals/0046-csrf-protection.md
+++ b/proposals/0046-csrf-protection.md
@@ -33,6 +33,8 @@ Astro should provide some level of security to users.
 
 # Detailed Design
 
+The solution should work only on on-demand applications (SSR), and applications with `server: "hybrid"` for routes that opt-out prerendering. 
+
 The solutions proposed should work only on request meant to **modify** data: `POST`, `PUT`, `PATCH`, or `DELETE`. Other requests should be exempt from the CSRF check.
 
 We will call these requests as **known requests**.
--- a/proposals/0046-csrf-protection.md
+++ b/proposals/0046-csrf-protection.md
@@ -60,9 +60,9 @@ The cookie solution is an alternative way to provide CSRF protection. It will be
 
 When the **first** `GET` request is sent to the application, Astro will create a token that will be saved inside a cookie named `Astro-csrf-token`. This token will be read in the **known requests** and it doesn't match the expected value, it means that the request isn't genuine.
 
-The token must be unique for each user and must be verifiable by the server; his prevents the client from making up its own tokens.
+The token must be unique for each user and must be verifiable by the server. This prevents the client from making up its own tokens.
 
-This solution should fit more esoteric scenarios, where applications are behind reverse proxies.
+This solution should fit more esoteric scenarios, where applications are behind reverse proxies, because ...
 
 A user can provide options to configure the name of the cookie and the name of the token:
 
--- a/proposals/0046-csrf-protection.md
+++ b/proposals/0046-csrf-protection.md
@@ -12,10 +12,7 @@ Provide the infrastructure to protect Astro websites from CSRF attacks
 // astro.config.mjs
 export default defineConfig({
     security: {
-        csrfProtection: {
-            origin: true,
-            cookie: true
-        }
+        checkOrigin: true
     }
 })
 ```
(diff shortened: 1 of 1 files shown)


florian-lefebvre:
I wonder if it would be better to make the option available under `security`, in case we want to add more security features (eg. honeypot) to core. So
```js
// astro.config.mjs
export default defineConfig({
  security: {
    csrfProtection: ["origin", "cookie"]
  }
})
```

natemoo-re:
@florian-lefebvre that's a good suggestion. We do try to avoid nesting too early since we don't necessarily know if we'll ever add something like a `security.honeypot` option, but if we already see lots of potential for expanding the APIs, adding a top-level category is a good suggestion.

florian-lefebvre:
I think CORS could be supported in the future as well but i can't see anything else rn

ematipico:
@lilnasy do you envision some option for the CSP support? If so, we could add a top-level `security` option.

proposals/0046-csrf-protection.md
```diff
@@ -0,0 +1,74 @@
...
+
+This solution should fit most websites.
+
+## `"cookie"` option
+
+The cookie solution is an alternative way to provide CSRF protection. It will be heavily inspired from [Angular](https://angular.io/guide/http-security-xsrf-protection).
+
+When the first `GET` request to the application is sent, Astro will create a token that will be saved inside a cookie named `Astro-csrf-token`. This token will be read in the **known requests**.
+
+This solution should fit more esoteric scenarios, where applications are behind reverse proxies.
```
natemoo-re, line 55: This seems more like background than implementation. Can we automatically match the `origin` or does it require a user to configure the `site`? How would the cookie be generated? Is the cookie static or refreshed for each build?

lilnasy:
Not necessarily. It's too soon to say because it's only stage 1, but it's possible a solution could just work. I only envision a flag for backwards compatibility, and that's only _**if**_ the picked solution can't be retrofitted.

proposals/0046-csrf-protection.md
```diff
@@ -0,0 +1,74 @@
...
+
+Astro should provide some level of security to users.
+
+# Goals
+
+- Add the required checks to prevent CSRF, probably via an option
+
+# Non-Goals
+
+- Give the users the possibility to customise the implementation of the protection
```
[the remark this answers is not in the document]
lilnasy: The threat of CSRF is only relevant to data mutations, so I would think that this is okay.

proposals/0046-csrf-protection.md
```diff
@@ -0,0 +1,74 @@
...
+  csrfProtection: ["origin", "cookie"]
+})
+```
+
+# Background & Motivation
+
+
+Most background is available here: https://owasp.org/www-community/attacks/csrf
+
+Astro should provide some level of security to users.
```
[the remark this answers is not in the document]
ematipico: I am not good at writing, so if you have a better suggestion, I would really appreciate it

ematipico:
In this commit [`4d37d51` (#879)](https://github.com/withastro/roadmap/pull/879/commits/4d37d51e5c5cf39a6d6d26be3a82dfce23c276bf) I changed the shape of the configuration due to some requirements that are needed for the cookie/token strategy.

proposals/0046-csrf-protection.md
```diff
@@ -0,0 +1,99 @@
...
+The header `origin` is a header that is preset and set by all modern browsers and there's no way to temper it. If the header `origin` won't match
+the origin of the URL, Astro will return a 403.
+
+This solution should fit most websites.
+
+## `"cookie"` option
+
+The cookie solution is an alternative way to provide CSRF protection. It will be heavily inspired from [Angular](https://angular.io/guide/http-security-xsrf-protection).
+
+When the **first** `GET` request is sent to the application, Astro will create a token that will be saved inside a cookie named `Astro-csrf-token`. This token will be read in the **known requests** and it doesn't match the expected value, it means that the request isn't genuine.
```
matthewp, line 61: What is the `**known requests**`? Which requests are these that it's checked?

proposals/0046-csrf-protection.md
```diff
@@ -0,0 +1,99 @@
...
+
+The token must be unique for each user and must be verifiable by the server; his prevents the client from making up its own tokens.
```
lilnasy, line 63: I'm not sure how angular's approach works.

- From the sound of it, it seems stateful. Who writes the state, who reads the state?
- The angular docs make mention of setting a custom header, implying they manage the client-side code that makes the request. Is astro going to offer a client-side `fetch()` wrapper?

A few implementation details may help communicate the approach.

proposals/0046-csrf-protection.md
```diff
@@ -0,0 +1,99 @@
...
+
+This solution should fit more esoteric scenarios, where applications are behind reverse proxies.
```
lilnasy, line 65: It's not apparent to my _why_ that would be the case. Could you elaborate?
```suggestion
This solution should fit more esoteric scenarios, where applications are behind reverse proxies, because ...
```

proposals/0046-csrf-protection.md
ematipico: They are mentioned [in this](https://github.com/withastro/roadmap/blob/feat/csrf-protection-rfc/proposals/0046-csrf-protection.md#detailed-design), third sentence.

proposals/0046-csrf-protection.md
```diff
@@ -0,0 +1,99 @@
...
+
+A user can provide options to configure the name of the cookie and the name of the token:
+
+```js
+// astro.config.mjs
+export default defineConfig({
+    security: {
+        csrfProtection: {
+            cookie: ["cookie-name", "token-name"]
```
lilnasy, line 74: It comes across as odd that the token "name" is configurable because I don't think the token was introduced as a key-value pair.

> Astro will create a token that will be saved inside a cookie named `Astro-csrf-token`.

If "Astro-csrf-token" is the cookie name, the token is the value of that cookie, what's a token name?

proposals/0046-csrf-protection.md
lilnasy, line 63:
```suggestion
The token must be unique for each user and must be verifiable by the server. This prevents the client from making up its own tokens.
```

septatrix:
This will break form submissions when using `curl` for example or pretty much any other tool which is not a modern browser such as http clients in most programming languages. Another example would be load-testing tools. While some like k6 support setting custom headers this is cumbersome and not supported by everything.

A cookie based implementation would at least work with everything which has a sense of session. So not curl but at least the other stuff I listed.

_Originally posted by @septatrix in https://github.com/withastro/astro/issues/10678#issuecomment-2051402692_

ematipico:
@septatrix: have you read the RFC? The RFC proposes a cookie solution, it would be great if you could share your thoughts about that.

matthewp:
@septatrix can you explain why this will break curl usage? Does curl not allow you to sit the origin header?

septatrix:
Curl does allow setting headers but this has to be done manually, existing automation might fail and - as said above - this is not the case for everything. E.g. many load testing tools are very simplistic and do not allow setting headers.

The cookie solution would still require changes to existing curl scripts and require one additional initial request to fetch the cookie. However, cookies are automatically handled by many other tools, especially client libraries in many programming languages establish session which store cookies between requests.

ematipico:
We decided to reduce the scope of the protection to the origin check only, hence the proposed configuration is also reduced. Changed applied in [`bc87ea7` (#879)](https://github.com/withastro/roadmap/pull/879/commits/bc87ea79e82a1200e8739d325540a2d6531b0fc4)


Also, this is a call for consensus! The RFC is ready to be shipped.

florian-lefebvre:
Is it possible to opt-out of CSRF on a route basis? For example for webhooks like stripe

ematipico:
> Is it possible to opt-out of CSRF on a route basis? For example for webhooks like stripe

No, it isn't possible at the moment. Do you have webhooks that send posts like they were forms?

florian-lefebvre:
Not sure I understand what you mean. In the case of stripe, a webhook is a post route that accepts some data (through `request.text()`). Stripe docs recommend [disabling cors](https://docs.stripe.com/webhooks#csrf-protection) for the webhook specifically

ematipico:
I understand now. I think it's a legitimate request, however, It can be implemented after this RFC lands

matthewp:
The final comment period has ended and this RFC is accepted.

Issue #837, opened 2024-02-09
Improved env handling


florian-lefebvre:
- Accepted Date: 2024-02-09
- Reference Issues/Discussions: #822
- Author: @florian-lefebvre
- Champion(s): @florian-lefebvre (with some help!)
- Implementation PR: N/A

# Summary

This RFC aims to improve DX (and eventually security) around env variables.

# Background & Motivation

Env variables are an important part of any web application. You need to store sensitive data (think API secrets, tokens etc) without being leaked inside your git repo. But that's only the 1st part of the story. It's easy to leak this data by importing in the wrong place, eg. the frontend like [Resend a few weeks ago](https://resend.com/blog/incident-report-for-january-10-2024).

Other JS frameworks (eg. [SvelteKit](https://kit.svelte.dev/docs/modules#$env-dynamic-private)) are handling env pretty well. From my understanding, the env story is currently a bit tricky in Astro. According to the [docs](https://docs.astro.build/en/guides/environment-variables/), here is how env variables are currently handled:

- Astro uses Vite’s built-in support for environment variables
- Static variables (ie. replaced statically at build time) are accessible via `import.meta.env`
- `import.meta.env` includes some [default variables](https://docs.astro.build/en/guides/environment-variables/#default-environment-variables) like `SSR`, `BASE_URL`...
- A public variable key has to be prefixed by `PUBLIC_`
- A non public variable accessed using `import.meta.env` on the client side will be `undefined` (value will be accessible server side)
- Env variables can be loaded through `.env` (or `.env.production`, `.env.development`) and CLI
- Any non built-in variable can be [manually typed]([.env.production](https://docs.astro.build/en/guides/environment-variables/#intellisense-for-typescript))
- Runtime variables should be access using `process.env`, or following the used runtime (eg. `Deno.env.get()` for the deno adapter)
- I'm not sure if `process.env` has any protection against client-side usage (like `import.meta.env`), but I guess it doesn't

# Goals

- Provide a fully type-safe experience for environment variables, without [manual type definitions](https://docs.astro.build/en/guides/environment-variables/#intellisense-for-typescript)
- Allow users to mark environment variables as required, providing hints when a variable is missing
- Reduce user confusion between inlined, static variables and dynamic, runtime variables
- Allow adapters to specify how runtime env variables should be handled

# Non-Goals 

- Complex validation or type casting of env variables. We might want to enable this at some point, but there is likely a performance cost for this at runtime. We should punt on this if possible!
[...]

PR #894, opened 2024-04-15
Astro Env

florian-lefebvre:
# Summary

`astro:env` - Improve DX and security around environment variables in Astro

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/feat/astro-env-rfc/proposals/0046-astro-env.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/837)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/822)


proposals/0046-astro-env.md
```diff
@@ -0,0 +1,366 @@
...
+# Example
+
+```js
+// astro.config.mjs
+import { defineConfig, envField } from "astro/config"
+
+export default defineConfig({
+  env: {
+    schema: {
+      API_URL: z.static().private().string(),
```
matthewp, line 30: Where does `z` come from? It's not shown in the import. For content collections it is inside of `astro:content`, would it be in `astro/config` here?

bholmesdev: I've wondered where `z` should come from as we use it from more utilities. I could see `z` being an expected utility for form actions as well. Since it could be used from three places, I wonder if we could:
1. Require Zod be explicitly installed in your project
2. Encourage importing `z` from the `astro/zod` module. This module exists today for our virtual modules to access

bholmesdev: Update: just noticed `z.static()` is referenced here, which is _not_ a core zod property. I would suggest a custom name like `envVar` or `envField` instead of `z`.

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,366 @@
...
+# Non-Goals
+
+- Complex validation or type casting of env variables. We might want to enable this at some point, but there is likely a performance cost for this at runtime. We should punt on this if possible!
+- Future: allow adapters to customize build-time variable handling. Better runtime handling is the most important problem this proposal aims to solve, but if we find an API that enables this as well, that's great.
+
+# Detailed Design
+
+## Terminology
+
+- `Static variable`: variable replaced by its value at build time. Uses `import.meta.env` under the hood
```
matthewp, line 77: These are not code so..

```suggestion
- **Static variable**: variable replaced by its value at build time. Uses `import.meta.env` under the hood
```

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,366 @@
...
+- `Dynamic variable`: variable retrieved at runtime, never part of the bundle. Uses runtime specific features, like `process.env` or `Deno.env.get()`
```
matthewp, line 78:
```suggestion
- **Dynamic variable**: variable retrieved at runtime, never part of the bundle. Uses runtime specific features, like `process.env` or `Deno.env.get()`
```

Same as above.

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,366 @@
...
+  export const FOO: string
+  export const PUBLIC_BAR: boolean
+} 
+```
+
+There's no protection to put in place regarding client-side usage, since `import.meta.env` already handles it.
+
+## Dynamic variables
+
+Dynamic variables are an adapter feature, meaning it's not supported by purely static site (ie. no adapter set).
```
natemoo-re, line 193: That's a bit of an unexpected limitation that would make migrating between `static <> hybrid <> server` more difficult. Any reason _not_ to support it for static builds and use similar behavior to the Node adapter?

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,366 @@
...
+  name: "supabase-integration",
+  hooks: {
+    "astro:config:setup": (params) => {
+      params.updateConfig({
+        env: {
+          schema: {
+            PUBLIC_SUPABASE_URL: envField.static().public().string()
+          }
+        }
+      })
```
natemoo-re, line 161: 👨‍🍳💋 Very nice!

proposals/0046-astro-env.md
florian-lefebvre: > use similar behavior to the Node adapter

What's the current behavior? Haven't used it much myself! But I guess runtime features could still be available and just default to node handling, so `process.env`. Is that what you are talking about?

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,366 @@
...
+
+Unit tests will be made for:
+- Custom validators
+
+E2e test will be made for:
+- Testing public/private
+- Testing static/dynamic
+- Integrations
+
+Features will be implemented incrementally using feature flags:
```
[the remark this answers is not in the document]
florian-lefebvre: Actually there's not really a reason 😅 I just remembered Ema doing something similar with i18n, but it was in a different context (ie. follow up rfc). I'll remove the weird flags.

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,366 @@
...
+
+```ts
+export default defineConfig({
+  experimental: {
+    env: {
+      schema: {},
+      // then
+      staticVariables: true,
+      // then
+      dynamicVariables: true
```
florian-lefebvre, line 240:
```suggestion
      schema: {}
```

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,366 @@
...
+
+## Dynamic variables using `Astro.env`
+
+The issue restrict dynamic variables usage inside `.astro` files (or endpoints `context`). It's common to be able to use it outside, eg. in `.ts`. That's why it uses an ALS.
+
+## Using zod
+
+Using zod in the public interface isn't great as `.env` files remain strings, so it requires more work to get right ([especially for booleans](https://env.t3.gg/docs/recipes#booleans)) and we only support a tiny subset of zod APIs.
+
+We've also considered using it under the hood. While it's not an issue for static variables (build time, not part of the bundle), it would increase the bundle for runtime usage significantly.
```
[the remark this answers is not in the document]
florian-lefebvre: I didn't put all the details here but I've tried many things with zod. One of the biggest problem is serialization and types, you need 2 libraries (from and to json schema) + some tricks, and that's only for static variables. And yeah for dynamic variables, Bjorn said we would import basically zod all the time for not so many features. I guess adding custom simple validators are fine, as long as there are enough tests

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,366 @@
...
+
+Note using such convention is still possible using the current proposal:
+
+```ts
+// env.ts
+import { envField } from "astro/config"
+
+export const envSchema = {
+  FOO: envField.static().public().string()
+} as const
```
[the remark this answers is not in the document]
florian-lefebvre: I think so because if you have a look at the detailed types above, public variables must start with `PUBLIC_` also at the type level. The `as const` allows users to catch the error earlier. An alternative here would be to provide a `defineEnvSchema` that allows to show errors directly.

That being said, it depends on what we want to promote: do we want to set this convention when there are too many variables or should we let people basically figure it out? (I'm in favor of such convention when it starts to grow significantly but not everyone was agreeing on the Stage 2 RFC)

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,366 @@
...
+    schema: envSchema
+  }
+})
+```
+
+# Adoption strategy
+
+- **If we implement this proposal, how will existing Astro developers adopt it?**
+  - They should not use [manual typing](https://docs.astro.build/en/guides/environment-variables/#intellisense-for-typescript) anymore
+  - Any custom env var (ie. not built-in like `SSR`) used with `import.meta.env` should be added to `env.schema` and imported through `astro:env/static`
```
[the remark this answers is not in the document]
florian-lefebvre: Basically, `import.meta.env` direct access would now be _kinda_ reserved for built-in variables, so nothing has to change here. We just need to make sure the user don't try to reuse reserved names

florian-lefebvre:
> I just wonder: with the separation between static and dynamic, are devs supposed to know and keep track of which env variables will be static or dynamic? Or can all env variables be accessed by both astro:env/static and astro:env/dynamic and basically it's just under the hood works differently?

No variables won't be available as both static and dynamic, it's either one or another. I think it's great this way for technical reasons and also so that devs are aware where their variables end up (eg. static public can end up on the client). Do you see a possible downside to such approach?

> Is there a reason astro uses the PUBLIC_ prefix for public variables, rather than the VITE_ prefix from Vite itself? Is it just to abstract Vite away a little bit more?

I don't know the reason behind this, I just reused the existing convention. IMO it's more explicit that it can end up on the client than having `VITE_` as a prefix

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,359 @@
...
+} 
+```
+
+There's no protection to put in place regarding client-side usage, since `import.meta.env` already handles it.
+
+## Dynamic variables
+
+Dynamic variables are an adapter feature, meaning it's not supported by purely static site (ie. no adapter set).
+
+> TODO: see what api makes most sense, probably an entrypoint but it also needs to be request dependent for Cloudflare
```
ematipico, line 195: Since this is an adapter feature, we will have to add a [new feature in the relative page](https://docs.astro.build/en/reference/adapter-reference/#astro-features), so adapters can signal Astro if they support it or not.

As for the implementation, I also think an entry point would make sense because it's in line with all the other adapters' APIs.

I suppose `getEnv` will call that entrypoint internally, right?

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,359 @@
...
+```js
+// astro.config.mjs
+import { defineConfig, envField } from "astro/config"
+
+export default defineConfig({
+  env: {
+    schema: {
+      API_URL: envField.static().private().string(),
+      PUBLIC_FOO: envField.static().public().string({ default: "bar" }),
+      STRIPE_KEY: envField.dynamic().private().string()
```
bluwy, line 32: 1. How do we support detecting leaks for `.static().private()`? If all static env vars are from `"astro:env/static"`, how to we know if it's being used in client code?
  
   I know you mention this below:

    > **Static variables** ... There's no protection to put in place regarding client-side usage, since `import.meta.env` already handles it.

    Does that mean `import.meta.env.API_URL` will be replaced as `undefined` if it's imported in client code? Otherwise it will error in runtime.


2. Should this be `.string().default("bar")` instead to mimic zod?
3. Like zod, are these fields all required by default unless you use `.optional()` or `.default()`?

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,359 @@
...
+  | Record<string, (PrivateStaticEnvFieldMetadata | PrivateDynamicEnvFieldMetadata) & EnvFieldType>
+
+type AstroUserConfig = {
+  env?: {
+    schema: EnvSchema
+  }
+}
+```
+
+Since there are various combinations possible that can make choosing confusing, we provide a `envField` helper that uses chaining, mirroring other APIs like Astro DB or zod:
```
bluwy, line 131: I wonder if it's cleaner to rename `envField` as `env` instead, but not a priority 😄

proposals/0046-astro-env.md
florian-lefebvre: > Does that mean import.meta.env.API_URL will be replaced as undefined if it's imported in client code? Otherwise it will error in runtime.

Yes I thought we could just return `undefined`. I guess it will trick a few people, so idk if the client can make the server trigger an `AstroError` in development (maybe using `import.meta.hot`?)

> Should this be .string().default("bar") instead to mimic zod?

I don't mind tbh!

> Like zod, are these fields all required by default unless you use .optional() or .default()?

I thought about this and I guess it depends how strict with want to be by default! So we can either provide `.optional()` or `.required()`

proposals/0046-astro-env.md
florian-lefebvre: Actually I've been thinking about this. The only issue I see with calling it `env` is that people could use it (because suggested by auto-complete) thinking it allows to access `env`. That's why I originally wanted to provide the helper from a function argument (see alternatives below) but ended up just renaming it to `envField`. I'm open to any naming tho! I think we can manipulate the ordering in the vscode extension too? Could be something to explore

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,359 @@
...
+Dynamic variables are an adapter feature, meaning it's not supported by purely static site (ie. no adapter set).
+
+> TODO: see what api makes most sense, probably an entrypoint but it also needs to be request dependent for Cloudflare
+
+Variables are accessible through the `astro:env/dynamic` virtual module:
+
+```ts
+import { getEnv } from "astro:env/dynamic"
+
+getEnv("FOO")
```
florian-lefebvre, line 202: I've been thinking about supporting variables that have not been declared in the astro config and I think we should. Basically:
1. If declared in the astro config, properly type it
2. If not, allow any string but
	1. adapters implementation must make sure the request env is of type string (eg. cloudflare with bindings)
	2. the function would return `string | undefined` in this case

proposals/0046-astro-env.md
bluwy: I also think we should support this for static builds. It would be kinda like the node adapter and source from `process.env.*` in build time. That means it kinda works like static but we don't prevent still using as dynamic so they can switch the output freely.

proposals/0046-astro-env.md
bluwy: 1. If we want a better error, we'd need to separate out as `"astro:env/static/public"` and `"astro:env/static/private"`, so that we can detect during import time. I'm not sure we can detect on usage time since it's dynamically used.

    Perhaps another idea is to export `getEnv()` like `"astro:env/dynamic"` so we can check if it's referencing a private env var at usage time. I'm not sure otherwise if we leave it as undefined since it can be hard to catch what's wrong. I am curious to hear what others think to though.
---
3. I think making it required by default may be better, so that we can warn if an env var is missing and the user can then decide whether it's intentional.

florian-lefebvre: 1. I like that approach
2. I'm also in favor of making it required by default

matthewp:
I don't think this idea of static/dynamic and public/private is quite right. Any "static" variable, meaning one that is set at build time, is always "public", in the sense that anyone with access to the source code can see the value. I would recommend a new distinction here:

- **public** is any variable that is replaced at build-time into the code.
- **private** is any variable that is retrieved at run-time through the `process.env[name]` etc. API.
- **static** is any variable that is defined in the config schema.
- **dynamic** is variables that are not, ie `getEnv(foo + bar)` is dynamic.

With these definitions, I don't think the `.dynamic()` and `.static()` helpers are needed at all.

Additionally, I don't think the `/static` and `/dynamic` subpaths are needed at all. They can all be part of `astro:env`, but the exports of `astro:env` will differ in the client where **only** the public items will be available.

If we want to restrict some public environment variables to only be available in the server or client I would recommend `.client()` and .server()` helpers.

florian-lefebvre:
I think that makes sense, let me summarize how the api looks and what it means:
- `.server().private()`:
	- `process.env`
	- Supports static (ie. key and value typed) and dynamic access (ie. any key, returns `string | undefined`)
- `.server().public()`:
	- `import.meta.env`, any key except built-in ones (eg. `import.meta.env.SSR`)
	- Only supports static access
- `.client().private()`:
	- Not supported because it would require passing some possibly sensible data to the client
- `.client().public()`:
	- `import.meta.env`, key prefixed by `PUBLIC` except built-in ones (eg. `import.meta.env.SSR`)
	- Only supports static access

Now let's say we have such config:

```ts
import { defineConfig, envField } from "astro/config"

export default defineConfig({
	env: {
		schema: {
			API_PORT: envField.server().private().number(),
			API_URL: envField.server().public().string(),
			PUBLIC_FLAG: envField.client().public().boolean()
		}
	}
})
```

If we have a single module like below, there's no issue on the server but on the client because we have no way to update types based on where it's called:
```ts
import { API_URL, PUBLIC_FLAG, getSecret } from "astro:env"

// On the server, no issues
API_URL // string
PUBLIC_FLAG // boolean
getSecret("API_PORT") // number
getSecret("UNKNOWN") // string | undefined

// But on the client...
API_URL // typed as string, returns undefined
PUBLIC_FLAG // boolean
getSecret("API_PORT") // typed as number, returns undefined
getSecret("UNKNOWN") // typed as string | undefined, returns undefined
```

I think it could make sense to at least have 2 imports like `astro:env/server` and `astro:env/client`:
```ts
import { API_URL, getSecret } from "astro:env/server"
import { PUBLIC_FLAG } from "astro:env/client"
```

Let me know what you think, and if I summarized correctly at the top!

bluwy:
I think that's also an interesting way to look at it. I would prefer chaining `.public()`/`.private()` first, and then chain `.client()`/`.server()`, as the bonus with this is that we can say `.client()`/`.server()` is only available after chaining `.public()`. `.private()` variables are always to be used in servers only. Personally I quite like this API too.

About combining `astro:env`, I also think we need a better way to warn misused env var rather than undefined (I also touched at [here](https://github.com/withastro/roadmap/pull/894#discussion_r1568226312)). I think what Florian suggested with two imports seems like the best way to warn in that case, but maybe it should be `astro:env/public` and `astro:env/private` instead? Because you can import public env vars in server code. If we stick with `astro:env/client`, it's technically safe to import in server code, but the name implies otherwise.

proposals/0046-astro-env.md
ematipico: I think `envField` is sound. I see `env` too close to `process.**env**` 😅

bholmesdev:
I agree with Blu on the ordering! I also have a question on API design. Astro DB uses a base function with an object parameter (`column.number({ optional: true })`), while Astro Env is using a chain (`envField.private().static()`). Ideally, Astro's custom validators would be aligned. Astro DB is pre-1.0, so that API choice is not set in stone. I'm curious what reasons y'all have for preferring one format over another?

florian-lefebvre:
My initial proposal was to have chaining first for the important stuff (so server/client and public/private) and then object syntax per type to match astro db:
```ts
envField.private().static().string({ optional: true })
```
I think Blu suggested to match zod api with full chaining
```ts
envField.private().static().string.optional()
```
I'm in favor to keep at least server/client and public/private as chained calls and then the type as an object

matthewp:
I agree with @bholmesdev that we should align the APIs. I think I tend to prefer making public/private/client/server properties of the options object rather than chaining. This mostly comes down to the discussion above about what order chaining should happen. 

In a [fluent API](https://en.wikipedia.org/wiki/Fluent_interface) (another name for chaining) the expectation is that they can be chained in any order as they are operating on a common `this` value. If that's not the case and ordering matters I would find using chaining to be unintuitive.

matthewp:
Some questions I have on the `client` and `server` options are:

1. Should it be required that you provide at least one of these?
2. If not required then what is the default? That it be applied to both?

I think I lean towards making at least one of these be required rather than picking a default, because accidentally bringing in values into the client, even if "public" would be a footgun.

matthewp:
On the topic of `getSecret()` in the client, I think it should throw if you call that function. The only reason for it existing at all is because of TypeScript. There's no use-case for calling it on the client, so we should throw.

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,367 @@
...
+    }
+  }
+} satisfies AstroIntegration
+```
+
+## Public variables
+
+Public variables are checked at some point between `astro:config:setup` and `astro:server:setup`. Since we don't/won't support many data structures, custom validators are used for the validation. They will also be used for secret variables at runtime.
+
+Some names must be reserved, like `SSR` to avoid conflicts with `import.meta.env.SSR`.
```
matthewp, line 175: I don't think we should reserve any names. Someone might very well have that as an environment variable.

If it's due to `import.meta.env`, I'm not sure we should be using that in the implementation. IIRC, `import.meta.env` falls back to runtime on the server. And if we're saying that you must explicitly call `private()` to get a runtime variable, then we shouldn't use import.meta.env.

bluwy:
> In a [fluent API](https://en.wikipedia.org/wiki/Fluent_interface) (another name for chaining) the expectation is that they can be chained in any order as they are operating on a common `this` value. If that's not the case and ordering matters I would find using chaining to be unintuitive.

I don't think we should strictly follow this. Instead I think we should follow a zod-like API so it feels unified. Zod also has certain cases where certain chains can only be used after another, for example [`z.string().datetime()`](https://zod.dev/?id=datetimes). If we opt for an pure-options based config, then it doesn't feel as unified (to me).

Also, Zod supports [.optional()](https://zod.dev/?id=optional) and [.default()](https://zod.dev/?id=default) instead of `{ default/optional: boolean }`.

Princesseuh:
No comments from me, looks neat. I like the builder API for the public and stuff, looks cute

matthewp:
I think we should have the fluent vs. options API discussion externally. It will probably block this RFC from being fully accepted but shouldn't affect implementation very much. I'll start a thread elsewhere.

proposals/0046-astro-env.md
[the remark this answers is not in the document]
florian-lefebvre: This is not relevant anymore, will update the proposal

Tc-001:
Hmmm... one thing that is not quite clear to me is how you would define an env variable for both client *and* server. 

Something like `PUBLIC_SOME_THIRD_PARTY_API="https://example.com/api"` that gets imported into `custom_fetch.ts` that then gets used by both server and client

florian-lefebvre:
I agree it's not super clear but a client variable can be imported on both client and server

Tc-001:
Oh I see!
Then you probably need to change this: https://github.com/withastro/roadmap/pull/894/files#diff-3c58a592f2b3662c3cda9ddfdd25a71de2a989fff0a5fcebfcd805dee36be9baR68

Otherwise looks really nice!

florian-lefebvre:
I've updated the RFC to match the current implementation PR state and make things clearer

ematipico:
@florian-lefebvre thank you for this RFC. I think you mostly covered everything, however something important that misses is how an adapter gets to provide their "getter" function: `process.env` VS `Deno.env`.

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,573 @@
...
+							srcKey: 'NETLIFY',
+						},
+						{
+							destKey: 'PUBLIC_DEPLOY_URL',
+							srcKey: 'DEPLOY_URL',
+							default: url,
+						},
+					],
+				});
+			}
```
ematipico, line 397: I don't think an adapter developer should import `overrideProcessEnv` like this (from the runtime, using a long import). Instead, I think we should provide this function **via hook**, maybe in `astro:config:done` (or another hook before the build). As you explained in the RFC, this is required during the dev/build, and we want to have these environment variables available **before** generating and rendering an actual page.

This way, we can control it, and we will provide this function **only if the feature is enabled by the user and supported by the adapter**. Win-win.

We could call it `setOverrideProcessEnv`. 

Also, can you explain what `default`, `destKey`, and `srcKey` are? I looked at the source code and adapter PR, and I couldn't understand them.

ematipico: > * It can be called in quite a few hooks, from `astro:config:setup` to `astro:build:setup`

Maybe I worded my thoughts badly, sorry! I meant, let's choose only **one** hook. If it's required for dev too, maybe `astro:config:setup` makes more sense.

> * It does nothing that involves astro (eg. settings)

Why? It sets an internal runtime function that is used inside the rendering phase (components that can read the environment variables)



> * `destKey` is the key that should be set on `process.env`
> 
> * `srcKey` is the key passed to `getEnv` inside the function. For example on netlify, that's important because `NETLIFY` is exposed but it has to be `PUBLIC_NETLIFY` with Astro Env, so it helps reducing boilerplate
> 
> * `default` is meant to be used as a fallback value, for example for `DEPLOY_URL` in dev

Awesome! I am not very good at naming, so I am happy with the one chosen. It would be great if we could document them

florian-lefebvre: > Why? It sets an internal runtime function that is used inside the rendering phase (components that can read the environment variables)

It's not! It's just setting `process.env` so it doesn't call any internal astro stuff.

> Maybe I worded my thoughts badly, sorry! I meant, let's choose only one hook. If it's required for dev too, maybe astro:config:setup makes more sense.

Well it's not exactly true. Yes `astro:config:setup` makes sense for dev but I would also expose it another hook that runs in both dev and build (probably `astro:config:done`) to allow the netlify build env variable inlining

matthewp: If all this does is set a variable on `process.env` then why do we need an API for this at all? Why don't they just set the variable themselves.

```js
"astro:config:done": () => {
  process.env.PUBLIC_NETLIFY = process.env.NETLIFY;
}
```

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,573 @@
...
+
+- In dev mode, it uses `process.env`
+- While building (eg. static site or prerendered pages), it also uses `process.env`
+- In SSR, it throws unless an adapter provides its own implementation
+
+#### Adapters API
+
+##### `app.setGetEnv`
+
+Under the hood, `getSecret` uses `getEnv` from the runtime. It can be overrided by calling `app.setGetEnv`:
```
matthewp, line 277: Why make this part of the `App` class? The `App` class is for rendering. Setting environment variables is unrelated to rendering for my hosts.

In addition, doing it this way means that variables can only be accessed inside of a render function. If you try to access a variable in the top-level of a module like so:

```js
import { getSecret } from 'astro:env/server';

const API_URL = getSecret('API_URL');
```

This will result in the variable being `undefined` because `setGetEnv` is not called until the very end of the bundle.

florian-lefebvre: Initially I added `getEnv` to `app.render` options to have it per request but then moved it to a method. I thought calling `app.setGetEnv` right after the instantiation of `App` would be enough, where should that go then?

ematipico: I suppose we don't need the function `app.setGetEnv`, and just expose the function `setGetEnv` from `astro/env`.

Users can still call `setGetEnv` from anywhere inside their code, even during the rendering phase, and at the top level module.

florian-lefebvre: > I suppose we don't need the function app.setGetEnv, and just expose the function setGetEnv from astro/env.

Alright so we can't use `astro/env` as it's already taken by a dts (https://github.com/withastro/astro/blob/main/packages/astro/package.json#L34), what export should we use then?

> Users can still call setGetEnv from anywhere inside their code, even during the rendering phase, and at the top level module.

As for matthew's example, it will still fail for Cloudflare I believe but I guess it's just a Cloudflare limitation in general.

> Users can still call setGetEnv from anywhere inside their code, even during the rendering phase, and at the top level module.

Currently, I'm calling `setGetEnv` with an implementation that throws in the `App` constructor as I assumed it would run before anything else. Where should I put it instead?

matthewp: Yes, top-level usage is probably not possible in Cloudflare. That's just a limitation they've chosen, and shouldn't affect other runtimes.

As for where this lives, since adapters are built through Vite, is there any reason they can't just use a `astro:env` export? `astro:env/setup` or something.

proposals/0046-astro-env.md
matthewp: Given this is a large enough RFC already, I think adding this extra complexity (and maintenance) is not worth it at this time. I don't think this adds enough value to justify the maintenance in Astro. Integrations are advanced usage so they should be able to manage conditionally setting up environment variables in dev, if needed.

proposals/0046-astro-env.md
florian-lefebvre: Sounds good

florian-lefebvre: The only issue I see is typing it, I guess it should not be done through typegen?

matthewp: I'm pretty sure you can add types for this, we have them in `packages/astro/client.d.ts`

theoephraim:
Hello fellow Astronauts!

We over @dmno-dev have been very deep in this specific problem space for the last few months, building out a platform-agnostic configuration tool meant to be a unified system for your _entire stack_. You can read more at [dmno.dev](https://dmno.dev) and specifically check out our [Astro integration](https://dmno.dev/docs/integrations/astro/). While our solution may be overkill for some folks, we hope it becomes the go-to solution for anyone working in a monorepo or wanting some tools with a bit more than what is built-in.

Regardless, we are super happy to see the progress on this RFC and for Astro to have amazing **first-class** env var handling. 

Apologies for being a little late to the party, as I know some excellent implementation work by @florian-lefebvre is already underway, but I wanted to share a few thoughts here (that I’ve shared with him already) from what we've learned while building out DMNO.

Please understand this is all meant to be constructive, simply to provoke discussion, and get the best possible outcome for Astro users.

----
## Challenges of hybrid rendering
In any SSR app (output `server` or `hybrid`) it can be very difficult to reason about where/when certain code will be run and/or bundled. In fact it often changes because the parent page/component or the site settings rather than what is happening in the component itself.

This presents challenges on two fronts - preventing leaks, and reasoning about if config will be actually dynamic (vary at boot time) or not.

There just is no easy way to enforce that the script part of a component has access to secrets, but only when being run on the server... Or to ensure that a config item you want to vary dynamically at boot time was not used during a pre-render of a page, effectively making it static even though it was not "bundled" into the code.

When the line between client and server can be so blurry, I think it will lead to confusion to say an item is a “server” or “client” item, or vary behaviour depending on how it is accessed.

Instead I think we should rely on a clear config schema to express the user's intent, build the system to do the right thing whenever possible, and to protect the user from anything unexpected.

## Config Schema
When we think about how to classify our config items, to me it really boils down to 2 concepts:

- **Is this data "sensitive" (vs "public")** (alt terms "secret","private")
  _meaning do we need to handle it sensitively and make sure it is never leaked_
  This obviously implies you can never access the item from the client. But considering we are in a world of hybrid server and client rendering and items may be used in various code that could be run in both, it’s just not that simple, and it is often not obvious while authoring code whether it will be running in the client or server. The important thing is, we handle the data with care and never leak the value in ANYTHING sent over the wire.

- **Is the item "dynamic" (vs "static")** (alt related terms “build” vs “boot” time, "runtime", “bundled”)
  _meaning is the value is always loaded at boot time and never bundled into built or pre-rendered code_
  Obviously if you are building a static site, everything will be replaced at build time… But if not, you may have items that you want to replace, and some which you may want to dynamically change using env vars at boot time. I do not believe this is as tied to whether something is sensitive as the current proposal and most frameworks (ex: `PUBLIC_`/`NEXT_PUBLIC_`) assume, likely because our tooling mostly originated in a SSG, totally static era.

> The important part here is that these 2 concepts are fairly **easy to reason about** and **totally decoupled**

In DMNO this is accomplished by marking items with `sensitive: true/false` and `dynamic: true/false` along with project-level settings to define the default behaviour (ex: `default_static` vs `default_dynamic`). 

_It should be noted that most users probably won't care about this dynamic/static stuff, and the default behaviour can still be set to `public_static`, which matches what most are used to (sensitive items are dynamic and public items are static)._

### The Edge cases

- **sensitive+static** - it is less common, but I see no reason to not let a user have a sensitive config item affect a build and tree-shaking
  _Re: security - yes if someone has access to the source code, they could see a secret that has been "baked" at build time, but in the case of deploying code to a server you control, or publishing a docker container to a private registry, those secrets would not be compromised._

- **public+dynamic** - this presents some challenges as they would need to be loaded by the client at runtime, but it certainly possible
  _Re: performance - these can be fetched on demand and users can decide to use them sparingly, not on page load_

You can read more about DMNO's handling of this in our [dynamic config guide](https://dmno.dev/docs/guides/dynamic-config/).

## Access patterns
The current proposal has multiple methods of accessing certain config items which is supposed to help achieve safety. But this means you must think about which access pattern to use, and changing the schema of an item may necessitate changing how it is accessed throughout your code. 

By using a combination of proxy objects and build-time rewrites, the user gets to access everything exactly the same way, and it all just works as intended. In DMNO, we inject 2 globals (_which could of course be imported from virtual modules instead_)
- `DMNO_CONFIG` (could be `ALL_CONFIG`) - includes every config item
- `DMNO_PUBLIC_CONFIG` (could be `PUBLIC_CONFIG`) - includes only _non-sensitive_ items

_NOTE - while the extra PUBLIC object is not strictly necessary, I do think it's helpful when browsing what items can be safely used on the client, and it's a bit closer to what folks are used to with the `PUBLIC_` prefix._

The injected proxy objects mostly let us give the user more helpful error messages (ex: “that item doesn’t exist” vs “that item exists but it is sensitive”), and it handles loading of any dynamic-public items. Loading these dynamic config items in the client uses a _blocking_ http request so that we can maintain the same DX, rather than needing to do something like `await getConfigItem(‘DYNAMIC_THING’)`.

## Protection from footguns
Once we have a full schema to express the user's intent, we can effectively protect the user from doing the wrong thing, without having to think too much:
- we only replace "static" items at build time, both on the client and the server
- `ALL_CONFIG` is not available on the client, although we do inject a proxy object that gives helpful error messages
- any dynamic+public items accessed on the client load from a blocking http request, on demand via that proxy object
- We inject a middleware to detect any leaked secrets before it goes out over the wire
- We can inject a vite build hook to detect any leaked secrets in any built assets
- We can check if any dynamic config items were used during a pre-render and throw or warn the user


### `PUBLIC_` prefix
One more note on the current proposal. Once we've introduced a proper schema, I don't think it is helpful to enforce a naming convention. I'd recommend keeping one or the other as it just adds another source of truth and opportunity for confusion. Personally with the rest of the system keeping you safe, I'd recommend relying on the schema, as it means you only update the schema, not where the config is accessed.

_In DMNO, our type system allows extendable types to be marked as sensitive - for example any instance of a `StripeSecretTokenDataType` is always going to be sensitive - so we must use the schema_

----

Hopefully this is helpful, at the very least to kick off some discussion. And please if this sounds interesting, head over to our [Astro integration guide](https://dmno.dev/docs/integrations/astro/) and give it a try.  We're happy to get deep into the weeds with folks and to help in any way we can!

Also please don't hesitate to reach out by email or on the Astro Discord.

Cheers!

theoephraim:
In doing a larger review of other tools, it looks like sveltkit is thinking along very similar lines to what I have proposed above - see https://kit.svelte.dev/docs/modules#$env-dynamic-private
They let you import from 4 different virtual modules for the combos of **public/private** and **static/dynamic**

They use the prefix only as the marker of public/private, and the import method to help control static/dynamic behaviour.

florian-lefebvre:
Yep we looked at it during the stage 2 rfc but thanks for bringing it again!

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,594 @@
...
+# Non-Goals
+
+- Complex validation or type casting of env variables. We might want to enable this at some point, but there is likely a performance cost for this at runtime. We should punt on this if possible!
+- Future: allow adapters to customize build-time variable handling. Better runtime handling is the most important problem this proposal aims to solve, but if we find an API that enables this as well, that's great.
+
+# Detailed Design
+
+## Terminology
+
+- **Public variable**: variable replaced by its value at build time. Its key must be prefixed by `PUBLIC_`
```
matthewp, line 82: Do we need the `PUBLIC_` restriction here? The idea of forcing `PUBLIC_` in the past was, correct me if I'm wrong, to force you to think about the fact that it's possibly going to be exposed. But now, since you define the variables in the schema and have to use `access: 'public'`, I don't think the added prefix is adding any value here.

proposals/0046-astro-env.md
```diff
@@ -0,0 +1,594 @@
...
+
+// Any .ts loaded by Vite
+import { API_PORT } from "astro:env/server";
+```
+
+### Secret server variables
+
+#### User API
+
+Secret server variables are available through `getSecret` exported from `astro:env/server`. This function allows retrieving a variable by its key:
```
matthewp, line 230: Is there a reason not to allow defined secrets to be made available as exports of `astro:env/server`? I think having auto-import is still nice for these variables. They would still need to use `getSecret` under the hood of course.

proposals/0046-astro-env.md
florian-lefebvre: The opposite (ie. `PUBLIC_` to be automatically public) was discussed [here](https://github.com/withastro/astro/pull/10974#discussion_r1595071201) and didn't really make sense to me. I initially wanted to keep the `PUBLIC_` prefix to match the behavior of `import.meta.env` (since it was supposed to be used originally) but I think you're right, it's not needed anymore. I'll submit a PR today

proposals/0046-astro-env.md
florian-lefebvre: I agree with you, I guess proxies are the way go for this?

florian-lefebvre: I've tested this and I don't think it can work. We don't want importing a secret to call `getSecret` because it will fail on cloudflare, we want this to happen when calling a secret. The thing is, proxies only work on object AFAIK

florian-lefebvre: Actually it might work, I need to see

Tc-001: Maybe if your approach doesn't work, something like `import { secrets } from 'astro:env/server'` and then you could do a `const {FOO, BAR} = secrets`

florian-lefebvre: I have a poc working using an AST, pretty promising

theoephraim:
Maybe if this is something y'all have already thought through, and maybe I'm overestimating it's importance, but I think you want to make sure users can still benefit from build-time injection and the constant folding /  tree-shaking that it enables. 

I think anything that looks like `import { SOME_VAR } from 'astro:env'` is going to make this very difficult.

On the other hand, using anything like `import AstroEnv from 'astro:env'` and then using `AstroEnv.SOME_VAR` makes it fairly trivial to add extra entries in the vite's [define](https://vitejs.dev/config/shared-options.html#define) configuration. Of course you probably won't want to automatically enable this for all env items have some rules about which ones can be replaced, based on the schema, or (less optimally) on how it was imported/used.

florian-lefebvre:
Just to make sure I understand correctly, if I have this:
```ts
import { FOO } from 'astro:env/client'

if (FOO) {
	// do something
}
```
and FOO is false, you expect the whole thing to be removed on build.

I have to admit I'm not familiar enough with vite to know if it currently works, we're doing some pretty simple stuff for public variables so maybe it works already? I think @bluwy may know

proposals/0046-astro-env.md
florian-lefebvre: Alright so I've made some tests and I see a few possible solutions with our **current** approach. None is perfect, there are drawbacks for each of them.

> Note: for each of this proposal, `getSecret` would became a wrapper of `getEnv`, without any special handling.

### Limitations

Because we export identifiers that are not object, we can't just use proxies and call it a day. Also, we can't just call `getSecret` because it would not work on cloudflare. For example, this would be invalid:

```ts
// generated virtual module for astro:env/server

const _internalGetSecret = (key) => {
    return // ...
}

export const FOO = _internalGetSecret('FOO')
```

### 1. Export functions

The easy way would be to export functions, to delay the call of `_internalGetSecret`:

```ts
export const FOO = () => _internalGetSecret('FOO')
```

#### Pros

- Simple to implement
- We get auto-completion from the import

#### Cons

- Unusual approach, I think many people will use them thinking it's the value
- Mismatch between server/public and server/secrets


### 2. Export a simple proxy

We could export a simple `secrets` proxy and call `_internalGetSecret` when accessing a key:

```ts
export const secrets = new Proxy({}, {
    get(_, p) {
        return _internalGetSecret(p)
    }
})
```

#### Pros

- Simple to implement
- No hidden gotcha

#### Cons

- We loose auto-completion, which is Matthew's request. It may not be better than our current solution

### 3. Export proxies with lazy getSecret call

We could export a proxy that would account for most cases, as suggested by Luiz:

```ts
function makeEnvProxy(key) {
  // Env could be unset, so use null to differentiate it not being read yet.
  let value = null;

  const getValue = () => {
    if (value === null) {
      // Would be `getSecret`
      value = process.env[key];
    }
    return value;
  }

  return new Proxy({}, {
    defineProperty(_, ...args) {
      const val = getValue();
      return Reflect.defineProperty(val, ...args);
    },
    deleteProperty(_, p) {
      const val = getValue();
      delete val[p]
      return true;
    },
    get(_, p) {
      const val = getValue();
      const prop = val[p];
      return typeof prop === 'function' ? prop.bind(val) : prop;
    },
    getOwnPropertyDescriptor() {
      const val = getValue();
      return Reflect.getOwnPropertyDescriptor(val);
    },
    getPrototypeOf() {
      const val = getValue();
      return Reflect.getPrototypeOf(val);
    },
    has(_, p) {
      const val = getValue();
      return p in val;
    },
    isExtensible() {
      const val = getValue();
      return Reflect.isExtensible(val);
    },
    ownKeys() {
      const val = getValue();
      return Reflect.ownKeys(val);
    },
    preventExtensions() {
      const val = getValue();
      return Reflect.preventExtensions(val);
    },
    set(_, p, v) {
      const val = getValue();
      val[p] = v;
      return true;
    },
    setPrototypeOf(_, v) {
      const val = getValue();
      return Reflect.setPrototypeOf(val, v);
    },
  })
}

export const FOO = makeEnvProxy('FOO');
```

#### Pros

- Covers most cases
- Not too hard to implement 

#### Cons

- It doesn't work in cases where users will actually expect it. If you do `console.log(FOO)`, it will actually print `{}`

### 4. Export proxies with a special property

We could do something like vue with `.value` and a simple proxy:

```ts
function _createSecretProxy(key) {
	return new Proxy(
		{},
		{
			get(_, p) {
				if (p === 'value') {
					const rawVariable = getEnv(key);
					const variable = rawVariable === '' ? undefined : rawVariable;
					const options = schema[key];
					const result = validateEnvVariable(variable, options);
					if (result.ok) {
						return result.value;
					}
					throw createInvalidVariableError(key, result.type);
				}
			},
		}
	);
}

export const FOO = _createSecretProxy('FOO')
```

#### Pros

- Simple to implement
- No hidden gotcha

#### Cons

- A bit harder to deal with, by having to add `.value`
- Some people may be confused thining it's the value, although typed as `{ value: T }`

### 5. Use an AST

We could export functions just like in `1.` but transform the calls using an AST:

```ts
// typed as number
import { FOO } from 'astro:env/server'

const X = FOO
// becomes
const X = FOO()
```

#### Pros

- Simple usage

#### Cons

- Hard to implement, many cases to cover and error-prone

Tc-001: For 2, can't you add generated types to it with something like `export const secrets: SecretsConfigValues = new Proxy(...)`, or is there some gotcha I am not aware of?

florian-lefebvre: yes for 2 i'm just talking about auto-completion, you have to first import `secrets` then destructure it. Typing is not an issue

bluwy:
Using `import { FOO }`, `import * as AstroEnv`, or `import AstroEnv` (if we support this) will all work with Rollup's treeshaking, so I don't think we have to change the syntax to accommodate it.

proposals/0046-astro-env.md
matthewp: We shouldn't need anything this complicated. One thing to remember is that ESM exports are *live*, which means if you modify them then anything references them changes as well. I haven't looked at the code, but i'm imagining you can do this this way:

```js
export let FOO = undefined;

export let getEnv = ...

export function setGetEnv(_getEnv) {
  getEnv = _getEnv;
  
  FOO = getEnv('FOO');
}
```

Basically, when `getEnv` is set, you also set all of the exported secret variables. Since `setGetEnv` is called early, this should happen before anyone starts using these variables.

proposals/0046-astro-env.md
matthewp: Is `access` require? I think it should be. If not and public is the default, then this would allow someone to accidentally make something public and expose sensitive data.

theoephraim:
Ah - right of course because it's _not_ rewritten by vite/rollups `define`... However this does also mean it needs to be an actual constant and not some proxy or function call. Thanks for confirming!

narration-sd:
Very glad to hear about this, as I think it will allow me to unwind all the complexity made when Astro 4 summarily removed ability for import.met.env, for a package adding Sasenity live Presentation editing on Astro.

I have no idea why the discussion I raised a the time didn't inform me about this -- you 'd have gotten early feedback, and I would have had a great deal less trouble. 

The trouble came because of the ferocious, non-traceable errors which occur if you allow the client to one way or another access what looked innocent, the server-side variables when produced by the recommended method.  This confuses Vite, in a very bad way.

narration-sd:
I'd like to ask -- when do you foresee this becoming a more-than-experimental feature (months, etc.)??

florian-lefebvre:
This feature should go out of experimental in a few minors, hard to tell at this time

florian-lefebvre:
New related RFC https://github.com/withastro/roadmap/discussions/956

florian-lefebvre:
New related RFC #957

braden-w:
@florian-lefebvre Thank you for your work on this RFC! It's exciting to see Astro moving towards improved environment variable handling. I have a few questions that I hope can contribute to the discussion:

1. Cross-context Usage:

The current proposal seems to focus primarily on using environment variables within the Astro application context. However, in might many projects, we need to access these variables in non-Astro contexts as well (e.g., standalone scripts, database connections, etc.). Is there a way we could extend this proposal to allow for seamless use of environment variables across all contexts?

2. Comparison with T3 env:

The [T3 env package](https://env.t3.gg/docs/core) offers some nice features like runtime validation, type inference, and the ability to use environment variables outside of the main application. How does this proposal compare to T3 env in terms of features and ergonomics? After using the RFC, I found the ergonomics feel slightly more awkward, especially since I was unable to use it in my scripts due to point 1. If we're going to recommend Astro users to use Astro env, what benefits does it provide that aren't provided in using the alternatives?


This might instead be going into the direction of a standalone module that can be imported in both Astro and non-Astro contexts. It might look something like this:

   ```typescript
   import { createEnv } from 'astro/env';

   export const env = createEnv({
     clientPrefix: 'PUBLIC_',
     server: {
       DATABASE_URL: z.string().url(),
       API_KEY: z.string().min(1),
     },
     client: {
       PUBLIC_API_URL: z.string().url(),
     },
     runtimeEnv: process.env,
   });
   ```

Where `env` is imported as a typesafe object whenever we need to use it.

This approach would allow for consistent environment variable usage across all parts of a project, including scripts and other non-Astro files.

hxjo:
Hello there,

First of all, thank you very much for your work, I love Astro and the people behind it.

I have a question regarding environment variables, why are we currently only reading those passed with the CLI command ? 
It means all environment variables present on the machine are ignored ; this is very impractical when those are handled by, for example, kubernetes and injected in the pod ; which makes me rely on process.env for now.

While I understand exposing the whole process.env is not desirable, maybe something similar to Vite with a prefix might do the trick ?

Again, thanks a lot for your work, and I'm eager to try this new api.
Have a great day

hxjo:
Hello there,

First of all, thank you very much for your work, I love Astro and the people behind it.

I have a question regarding environment variables, why are we currently only reading those passed from .env.* and with the CLI command ? 
It means all environment variables present on the machine are ignored ; this is very impractical when those are handled by, for example, kubernetes and injected in the pod ; which makes me rely on process.env for now.

While I understand exposing the whole process.env is not desirable, maybe something similar to Vite with a prefix might do the trick ?

Again, thanks a lot for your work, and I'm eager to try this new api.
Have a great day

florian-lefebvre:
There are a few scenarios:
- In dev and build, we use `loadEnv` from vite and merge its result inside `process.env` so I don't see why those global k8s vars would not be available
- In SSR, it uses whatever thing the adapter specified, eg. `process.env` for node, `Deno.env.get()` for deno etc
- You can access this implementation directly by calling `getSecret('KEY')`. It's literally an interface for the point above

If you think that's a bug, feel free to open an issue on the main repo and ping me! Happy to look into this

hxjo:
> - In SSR, it uses whatever thing the adapter specified, eg. `process.env` for node, `Deno.env.get()` for deno etc
Ooh, that makes sense. I'm in SSR with node which explains the need to use process.env then.


> - You can access this implementation directly by calling `getSecret('KEY')`. It's literally an interface for the point above
I'll try that today then !

Does not seem like a bug at all, just a misunderstanding from me.
Thanks for the clarification!

florian-lefebvre:
This is a call for consensus, and we aim to close the RFC in the next three days. Let us know if you have any important concerns.

florian-lefebvre:
The final comment period has ended and this RFC is accepted. The feature will be unflagged in Astro 5.0.

Issue #896, opened 2024-04-16
Rerouting 


ematipico:
- Accepted Date: 2024-04-16
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/681
- Author: @lilnasy 
- Champion(s): @ematipico @lilnasy 
- Implementation PR: 

# Summary

Programmatic control over routing.

# Background & Motivation

As astro grows in popularity, it is being used for more and more use cases. One of these is multitenant websites where a tenant may be given a subdomain. This proposal introduces a powerful primitive that allows astro users to build websites with complex routing requirements. Rather than a limitation, file-based routing is now a sensible default!

# Goals

- Defuse.
- Make it unnecessary to rely on implicit routing rules
- Introduce framework-designed 404 and 500 pages 
- Make middleware run for requests that don't match a route (with framework-designed error pages)

# Non-Goals 

- Reroute to an external service.
- Support for `functionPerRoute`

florian-lefebvre:
Stage 3 RFC available in #901

PR #901, opened 2024-04-18
Rewriting

ematipico:
# Summary

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/feat/reroute/proposals/0047-rerouting.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/896)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/681)

### DIFF
--- a/proposals/0047-rerouting.md
+++ b/proposals/0047-rerouting.md
@@ -150,6 +150,21 @@ When a user triggers a rerouting to another URL/route:
 - Astro will be able to detect possible loops, in case the user tries to render the same route over and over again. I case a loop is detected, Astro will abort the rendering phase and return a [`508` response`]( https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/508).
 - Inside the middleware, when a user calls `ctx.reroute("/")`, Astro will **re-run** the middleware again. That's required because if the user has some middleware logic that runs in `/`, it's their expectation to have that logic to trigger when rendering the page.  
 
+## `next("/")` VS `ctx.reroute('/')`
+
+The two functions, **when used inside a middleware** will behave differently: 
+
+### `ctx.reroute('/')`
+- It will stop the normal execution of middleware functions, all middleware functions after the current one won't be called.
+- The middleware will re-run again with the new `Request`/`APIContext`
+
+
+### `next('/')`
+- It **won't** stop the normal execution of middleware functions.
+- The next function after `next('/')` will receive a new `Request`/`APIContext` that is **manipulated** and will contain the rerouted URL.
+- `APIContext.params` aren't recreated due to some architectural constraints, as we don't have a `RouteData` type, needed to construct the new `params`.
+
+
 ## Adapters
 
 I don't envision any major blocker for our current adapters. 
@@ -185,3 +200,5 @@ Another alternative that was evaluated was to make the feature available only fr
   // VS
   Astro.reroute(new Request());
   ```
+- Should we call it `reroute` or `rewrite`? SvelteKit uses reroute, but many other frameworks and web servers use the term rewrite.
+- Do we need `next('/')`? 
--- a/proposals/0047-rerouting.md
+++ b/proposals/0047-rerouting.md
@@ -1,0 +1,180 @@
+<!--
+  Note: You are probably looking for `stage-1--discussion-template.md`!
+  This template is reserved for anyone championing an already-approved proposal.
+
+  Community members who would like to propose an idea or feature should begin
+  by creating a GitHub Discussion. See the repo README.md for more info.
+
+  To use this template: create a new, empty file in the repo under `proposals/${ID}.md`.
+  Replace `${ID}` with the official accepted proposal ID, found in the GitHub Issue
+  of the accepted proposal.
+-->
+
+- Start Date: 2024-04-17
+- Reference Issues: https://github.com/withastro/roadmap/issues/896
+- Implementation PR: 
+
+# Summary
+
+Programmatic control over routing.
+
+# Background & Motivation
+
+
+As astro grows in popularity, it is being used for more and more use cases. One of these is multitenant websites where a tenant may be given a subdomain. This proposal introduces a powerful primitive that allows astro users to build websites with complex routing requirements. Rather than a limitation, file-based routing is now a sensible default!
+
+# Goals
+
+- Defuse.
+- Make it unnecessary to rely on implicit routing rules
+- Introduce framework-designed 404 and 500 pages
+- Make middleware run for requests that don't match a route (with framework-designed error pages)
+
+# Non-Goals
+
+- Reroute to an external service.
+- Support for `functionPerRoute`
+- fallback rerouting
+
+# Detailed Design
+
+## Proposed APIs
+
+The rerouting will be exposed to Astro pages, Astro endpoints and middleware. This API is a function called `reroute` that will contain the following signature:
+
+```ts
+type Reroute = string | URL | Request
+const reroute: (payload: Reroute) => Promise<Response>
+
+reroute("/fr/hello");
+reroute(new URL("https://example.com"));
+reroute(new Request("https://localhost:8080", options));
+```
+
+Internally, the `reroute` **must** create a new `Request` when attempting to render the rerouted route. The creation of this new `Request` will vary based on the signature used.
+
+- Accepting a `string` allows to quickly reroute to a URL without too much hassle. When using a `string`, Astro will create a new `Request` with the new URL, and it will inherit all the data from the previous request.
+  > Astro won't do any particular check on the string. For example, it won't check for trailing slashes. 
+- Accepting a `URL` allows to create a more stable URL using the standard `URL` object. When using a `URL`, Astro will create a new `Request` with the new URL, and it will inherit all the data from the previous request.
+  > Astro will check against domains that aren't allowed. Only reroutes to the current host are allowed.
+- Accepting a `Request` allows uses to manipulate the `Request` as much as they can. When using a `Request`, Astro **will not create** a new `Request` and it will use the one provided by the user.
+  > Astro will check against domains that aren't allowed. Only reroutes to the current host are allowed.
+
+### Astro pages usage
+
+The `Astro` global will expose a new method called `reroute`:
+
+```astro
+---
+if (Astro.url.startsWith("/fr/salut")) {
+    return Astro.reroute("/fr/hello")
+}
+---
+```
+
+### Astro endpoints usage
+
+The `context` - `APIContext` type - will expose a new method called `reroute`
+
+```js
+
+export const GET = (context) => {
+    return context.reroute("/fr/hello")
+}
+```
+### Middleware usage
+
+Other than exposing the `reroute` function via `context`, the signature of the `next` function will change to accept the same payload of the `reroute` function:
+
+```js
+export const onRequest = (context, next) => {
+    if (context.url.startsWith("/fr/salut")) {
+        return next("/fr/hello")
+    }
+    return next()
+}
+```
+
+The `next` function will have this new signature, and it will be backwards compatible with the existing signature:
+
+```ts
+type MiddlewareNext = (payload?: Reroute) => Promise<Response>;
+```
+
+
+### Results for static and prerendered pages
+
+In SSG, rerouting to a page will result in the compilation of a page that has the contents of a different page. 
+
+If we have two pages, `src/pages/about.astro` and `src/pages/contact.astro`, if the `contact.astro` will reroute to `/about` using the rerouting strategy, it's contents will be the same as the `about.astro` component.
+
+```html
+<!-- src/pages/about.astro-->
+<html>
+  <head>
+    <title>Astro</title>
+  </head>
+  <body>About</body>
+</html>
+```
+
+```astro
+---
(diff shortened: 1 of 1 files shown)


proposals/0047-rerouting.md
```diff
@@ -0,0 +1,186 @@
+<!--
```
matthewp, line 1: Can you remove the comment?

proposals/0047-rerouting.md
```diff
@@ -0,0 +1,186 @@
...
+- Implementation PR: 
+
+# Summary
+
+Programmatic control over routing.
+
+# Background & Motivation
+
+
+As astro grows in popularity, it is being used for more and more use cases. One of these is multitenant websites where a tenant may be given a subdomain. This proposal introduces a powerful primitive that allows astro users to build websites with complex routing requirements. Rather than a limitation, file-based routing is now a sensible default!
```
matthewp, line 24: I don't think I follow this explanation of motivation. What does subdomains have to do with this feature?

Reroute here is like a rewrite rule that has been popular with tradition HTTP servers for a long time, for example with nginx: https://www.nginx.com/blog/creating-nginx-rewrite-rules/

One common motivation is to support routes being accessible from multiple URLs. That's why you might use this feature in middleware.

Another use-case specific to Astro is wanting to render a different page with the same URL, for example to show a login page when the user is missing the login cookie.

proposals/0047-rerouting.md
```diff
@@ -0,0 +1,186 @@
...
+
+# Goals
+
+- Defuse.
```
matthewp, line 28: ?

proposals/0047-rerouting.md
```diff
@@ -0,0 +1,186 @@
...
+- Make it unnecessary to rely on implicit routing rules
+- Introduce framework-designed 404 and 500 pages
```
matthewp, line 30: How does this feature allow that?

proposals/0047-rerouting.md
```diff
@@ -0,0 +1,186 @@
...
+- Make middleware run for requests that don't match a route (with framework-designed error pages)
+
+# Non-Goals
+
+- Reroute to an external service.
+- Support for `functionPerRoute`
+- fallback rerouting
```
matthewp, line 37: I think all of these non-goals are good but could use a bit more explanation for those who might not understand the connections.

proposals/0047-rerouting.md
```diff
@@ -0,0 +1,186 @@
...
+
+reroute("/fr/hello");
+reroute(new URL("https://example.com"));
+reroute(new Request("https://localhost:8080", options));
+```
+
+Internally, the `reroute` **must** create a new `Request` when attempting to render the rerouted route. The creation of this new `Request` will vary based on the signature used.
+
+- Accepting a `string` allows to quickly reroute to a URL without too much hassle. When using a `string`, Astro will create a new `Request` with the new URL, and it will inherit all the data from the previous request.
+  > Astro won't do any particular check on the string. For example, it won't check for trailing slashes.
```
matthewp, line 57: Doesn't it call `matchRoute` internally? I'm pretty sure that code does check trailing slash rules.

proposals/0047-rerouting.md
```diff
@@ -0,0 +1,186 @@
...
+reroute("/fr/hello");
+reroute(new URL("https://example.com"));
+reroute(new Request("https://localhost:8080", options));
+```
+
+Internally, the `reroute` **must** create a new `Request` when attempting to render the rerouted route. The creation of this new `Request` will vary based on the signature used.
+
+- Accepting a `string` allows to quickly reroute to a URL without too much hassle. When using a `string`, Astro will create a new `Request` with the new URL, and it will inherit all the data from the previous request.
+  > Astro won't do any particular check on the string. For example, it won't check for trailing slashes. 
+- Accepting a `URL` allows to create a more stable URL using the standard `URL` object. When using a `URL`, Astro will create a new `Request` with the new URL, and it will inherit all the data from the previous request.
```
matthewp, line 58: If you're looking for a use-case, providing a URL is nice when you are rerouting from a nested path because you can use a relative like, for ex. if you are at `/foo/bar/baz` and want to go to `/foo/bar/qux` you can do `reroute(new URL('./qux', Astro.url))`

proposals/0047-rerouting.md
```diff
@@ -0,0 +1,186 @@
...
+  > Astro will check against domains that aren't allowed. Only reroutes to the current host are allowed.
+- Accepting a `Request` allows uses to manipulate the `Request` as much as they can. When using a `Request`, Astro **will not create** a new `Request` and it will use the one provided by the user.
```
matthewp, line 60: Might mention that this allows manipulating headers as that's a primary use-case.

proposals/0047-rerouting.md
```diff
@@ -0,0 +1,186 @@
...
+```html
+<html>
+  <head>
+    <title>Astro</title>
+  </head>
+  <body>About</body>
+</html>
+```
+
+### Results for on-demand pages
```
matthewp, line 138: What is an `on-demand` page, is that defined anywhere?

proposals/0047-rerouting.md
```diff
@@ -0,0 +1,186 @@
...
+- Integration testing
+- Manual testing
+
+# Drawbacks
+
+The current API is a blocker for `functionPerRoute`. Since the rerouting isn't statically analysable, it isn't possible to tell serverless computing to reroute a request to another function/lambda. 
+
+# Alternatives
+
+I considered a static approach like SvelteKit and Next.js. While this approach would allow to achieve `functionPerRoute` support, we risk to litter the configuration with many entries. Plus, a static approach doesn't allow enough flexibility.
```
matthewp, line 169: One alternative that we discussed was making this purely a middleware feature doing something like `next({ request: new Request('/other')... })`. I would add that and discuss the tradeoffs.

matthewp:
We should figure out what happens if there are cycles; for example if `/foo` reroutes to `/bar` which reroutes back to `/foo` on an endless loop. Should we do anything to try and prevent that?

proposals/0047-rerouting.md
ematipico: Internally, we use `route.pattern.test`. `pattern` is the path-to-regex. Does this regex check for trailing slashes? If so I can omit the paragraph

ematipico:
> We should figure out what happens if there are cycles; for example if `/foo` reroutes to `/bar` which reroutes back to `/foo` on an endless loop. Should we do anything to try and prevent that?

In my early prototype, I was able to detect loops. I added a paragraph for that here [`c8f69ac` (#901)](https://github.com/withastro/roadmap/pull/901/commits/c8f69ac5e7f72a6b60227cb2edeaba2316af8856)

matthewp:
Let's say you had a middleware stack like this:

```
i18n -> custom -> auth -> page
```

Given that `reroute()` returns a response, I think this means that if you call reroute in the `custom` middleware function, then `auth` will not be called. Is that right? Or it the function aware of where the current point is in the middleware chain and continues to call the proceeding middleware functions?

ematipico:
> Let's say you had a middleware stack like this:
> 
> ```
> i18n -> custom -> auth -> page
> ```
> 
> Given that `reroute()` returns a response, I think this means that if you call reroute in the `custom` middleware function, then `auth` will not be called. Is that right? Or it the function aware of where the current point is in the middleware chain and continues to call the proceeding middleware functions?

That's right, I updated the proposal to better explain the API and how `next("/new")` vs `ctx.reroute("/new")` differ

ematipico:
The commit [`f9ec599` (#901)](https://github.com/withastro/roadmap/pull/901/commits/f9ec5993b39bfc548b514581b467d5859f440703) changes "reroute" to "rewrite". We assessed that the majority of frameworks and services out there use the term "rewrite".

vchirikov:
Hi, I want to discuss about the limitations of the current implementation:
1. I think it will be useful to give an opportunity to middleware to change not only URL, but the properties of context. e.g. `locale` or `locals`. 

2. It will be useful to rewrite to `/404` / `/500`, e.g. I want to use a custom i18n middleware which extract the url part from pathname or use the default and rewrite (reroute) to the same page (no additional files for /en, /de, etc). For example `/de/about` -> about.astro, `/en/about` -> about.astro, `/about` -> about.astro. To do this I write a middleware to provide a locale, because I can't change `Astro.currentLocale` (point 1) I provide a locale via Astro.locals something like this:

```ts
import type { MiddlewareHandler } from 'astro';
import type { AstroError } from 'astro/errors';

export function isAstroError(value: unknown): value is AstroError {
  return !!value && typeof value === 'object' && 'type' in value && value.type === 'AstroError';
}

import { defaultLocale, extractLocale } from '@/lib/i18n';
import { removeTrailingSlash } from '@/lib/utils/url';

const frontendUrl = removeTrailingSlash(process.env.FRONTEND_URL);
const frontendUrlLen = frontendUrl.length;

export const onRequest: MiddlewareHandler = async (ctx, next) => {
  // can't set ctx.currentLocale, and can't rewrite with new currentLocale
  if (ctx.request.url.startsWith(frontendUrl)) {
    const urlWithoutBase = ctx.request.url.slice(frontendUrlLen);
    const { locale, url } = extractLocale(urlWithoutBase);
    ctx.locals.locale = locale ?? defaultLocale;
    try {
      await next(url);
      return;
    }
    catch (error: unknown) {
      // now in case of 404 or rendering error astro will throw RewriteEncounteredAnError
      if (isAstroError(error) && error.name === 'RewriteEncounteredAnError') {
        next('/404'); // can't do this, so I can't localize 404.astro properly
        // or next('/500'); in case or rendering error
      }
    }
  }
  await next();
};
```

But I can't rewrite to /404 for now.

ematipico:
> Hi, I want to discuss about the limitations of the current implementation:
> 1. I think it will be useful to give an opportunity to middleware to change not only URL, but the properties of context. e.g. `locale` or `locals`. 
> 

Can you provide more specifics here? As long your code runs in the middleware, you can always change `locals`, and I fail to see the connection with the rewriting, that's why I ask more specifics.

> 2. It will be useful to rewrite to `/404` / `/500`, e.g. I want to use a custom i18n middleware which extract the url part from pathname or use the default and rewrite (reroute) to the same page (no additional files for /en, /de, etc). For example `/de/about` -> about.astro, `/en/about` -> about.astro, `/about` -> about.astro. To do this I write a middleware to provide a locale, because I can't change `Astro.currentLocale` (point 1) I provide a locale via Astro.locals something like this:
> 
> ```ts
> import type { MiddlewareHandler } from 'astro';
> import type { AstroError } from 'astro/errors';
> 
> export function isAstroError(value: unknown): value is AstroError {
>   return !!value && typeof value === 'object' && 'type' in value && value.type === 'AstroError';
> }
> 
> import { defaultLocale, extractLocale } from '@/lib/i18n';
> import { removeTrailingSlash } from '@/lib/utils/url';
> 
> const frontendUrl = removeTrailingSlash(process.env.FRONTEND_URL);
> const frontendUrlLen = frontendUrl.length;
> 
> export const onRequest: MiddlewareHandler = async (ctx, next) => {
>   // can't set ctx.currentLocale, and can't rewrite with new currentLocale
>   if (ctx.request.url.startsWith(frontendUrl)) {
>     const urlWithoutBase = ctx.request.url.slice(frontendUrlLen);
>     const { locale, url } = extractLocale(urlWithoutBase);
>     ctx.locals.locale = locale ?? defaultLocale;
>     try {
>       await next(url);
>       return;
>     }
>     catch (error: unknown) {
>       // now in case of 404 or rendering error astro will throw RewriteEncounteredAnError
>       if (isAstroError(error) && error.name === 'RewriteEncounteredAnError') {
>         next('/404'); // can't do this, so I can't localize 404.astro properly
>         // or next('/500'); in case or rendering error
>       }
>     }
>   }
>   await next();
> };
> ```
> 
> But I can't rewrite to /404 for now. 

404 and 500 would require some attention. Instead of talking code, can you provide expectations base on use cases?

vchirikov:
1. about changing context:
I want to use same `.astro` files to show different localized pages. Translation of pages should be done with i18next (for example). The desired project structure and urls should be:

```
├───pages
│       404.astro        // same page should be usable with different locales (e.g. en,de,ru)
│       about.astro      // urls: /about, /en/about, /de/ebout, /ru/about
│       index.astro      // urls: /, /en, /de, /ru
```
To do this I can implement custom i18n middleware which allow to re-use same `.astro` pages with different locale (via rewriting from `/en/about` to `/about`), but I can't set the `ctx.currentLocale` (it has no setter) with current implementation of rewriting. The proof:

image

Because I can't change the context's `Astro.currentLocale` via rewriting, instead of Astro-integrated i18n features I have to use `Astro.locals` to provide locale to the pages.

My point is it seems useful to give a user an extension point to define/edit readonly context fields from the middleware. Especially redefine `Astro.currentLocale`.

p.s. thanks a lot for the rewriting feature :)
p.p.s I tried to be as clear as possible, but if something else remains unclear - feel free to ask :)

ematipico:
Thank you, @vchirikov; I will take your feedback into account. However, this is something **unrelated** to rewriting. These are features related to i18n.

ematipico:
This is a call for consensus, and we aim to close the RFC in the next three days. Let us know if there are any important concerns.

PR #904, opened 2024-04-19
Clarify RFC meetings

matthewp:
The process calls for RFC meetings, which we have not had formally in quite a while. This change brings those back and clarifies who can call for them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@
 
 **Location:** GitHub Issues [(see all accepted proposals).](https://github.com/withastro/roadmap/issues)
 
-**What to Expect:** A proposal reaches this stage (aka "is accepted") during an [#rfc-meetings](RFC Meeting) with Maintainers and TSC, following our existing [RFC voting](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#voting-rfc-proposals) process.
+**What to Expect:** A proposal reaches this stage (aka "is accepted") during an [#rfc-meetings](RFC Meeting) with Maintainers and TSC (>= L2), following our existing [RFC voting](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#voting-rfc-proposals) process.
 
 When a proposal is accepted, a TSC member will create a new GitHub Issue summarizing the original proposal using our official template. At this point, the proposal champion is free to move on to the next stage. If a champion doesn't exist yet, then an accepted proposal may remain open until a champion volunteers by posting in the GitHub Issue.
 
@@ -64,7 +64,7 @@ RFCs advance through the stages during RFC meetings with TSC and maintainers. Vo
 
 Meetings occur ad hoc rather than on a scheduled basis. They are called for when a proposal author or champion feels it is ready to advance to the next stage. The author or champion can ask for a meeting by contacting TSC to schedule a time.
 
-All maintainers are invited to the meeting. If consensus is reached the RFC advances to the next stage. See [RFC proposal](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#voting-rfc-proposals) documentation for full details.
+All maintainers are invited to the meeting. If consensus is reached, the RFC advances to the next stage. See [RFC proposal](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#voting-rfc-proposals) documentation for full details.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@
 
 **Location:** GitHub Issues [(see all accepted proposals).](https://github.com/withastro/roadmap/issues)
 
-**What to Expect:** A proposal reaches this stage (aka "is accepted") during a meeting with Maintainers and TSC, following our existing [RFC Proposal](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#voting-rfc-proposals) voting process.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,7 +30,7 @@
 
 **Location:** GitHub Issues [(see all accepted proposals).](https://github.com/withastro/roadmap/issues)
 
-**What to Expect:** A proposal reaches this stage (aka "is accepted") during a meeting with Maintainers and TSC, following our existing [RFC Proposal](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#voting-rfc-proposals) voting process.
+**What to Expect:** A proposal reaches this stage (aka "is accepted") during an [#rfc-meetings](RFC Meeting) with Maintainers and TSC, following our existing [RFC voting](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#voting-rfc-proposals) process.
```
sarah11918, line 33:
```suggestion
**What to Expect:** A proposal reaches this stage (aka "is accepted") during an [RFC Meeting](#rfc-meetings) with Maintainers and TSC, following our existing [RFC voting](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#voting-rfc-proposals) process.
```

README.md
```diff
@@ -58,6 +58,14 @@ An RFC is ready to be approved and finalized once it's Pull Request is ready for
 
 At this time, some member of the core team will motion for a final comment period (FCP). This follows our existing [RFC Proposal](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#voting-rfc-proposals) voting process. Once the final comment period has elapsed the RFC will be merged if there are no objections.
 
+## RFC Meetings
+
+RFCs advance through the stages during RFC meetings with TSC and maintainers. Voting follows the [RFC proposal](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#voting-rfc-proposals) voting process.
+
+Meetings occur ad hoc rather than on a scheduled basis. They are called for when a proposal author or champion feels it is ready to advance to the next stage. The author or champion can ask for a meeting by contacting TSC to schedule a time.
```
sarah11918, line 65:
```suggestion
Meetings occur ad hoc rather than on a scheduled basis. They are called for when a proposal author or champion feels the proposal is ready to advance to the next stage. The author or champion can ask for a meeting by contacting TSC to schedule a time.
```
"contacting TSC" is pretty vague. Unless we describe how to do that elsewhere on this page, I'd specify *how* they should contact TSC. Do they mention the role in the proposal? Email?

README.md
```diff
@@ -58,6 +58,14 @@ An RFC is ready to be approved and finalized once it's Pull Request is ready for
...
+
+All maintainers are invited to the meeting. If consensus is reached the RFC advances to the next stage. See [RFC proposal](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#voting-rfc-proposals) documentation for full details.
```
sarah11918, line 67:
```suggestion
All maintainers are invited to the meeting. If consensus is reached, the RFC advances to the next stage. See [RFC proposal](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#voting-rfc-proposals) documentation for full details.
```
Should we also say what happens if consensus is *not* reached? Is there a waiting limit before asking again? Is the proposal officially dead?

sarah11918 (commented):
Thanks @matthewp !  A couple of grammar nits, and some questions for clarity that we might also want to include in the text.

README.md
matthewp: The governance document talks about that some: https://github.com/withastro/.github/blob/main/GOVERNANCE.md#voting-rfc-proposals

Was trying not to repeat too much here that's already covered in the governance doc.

README.md
FredKSchott, line 65: Would doing these discussions during a Tuesday community call actually be a good idea? Would it increase participation/attendance (esp. from the rest of core), give @sarah11918 a built-in agenda item, and require less planning to organize and find a time?

FredKSchott: Especially reading the next paragraph, it feels a little "secret meeting" and the transparency wins of doing it on a public call could be worth it for that alone.

matthewp: We could, but some issues we had before when RFC meetings were open to all:

- Discussions tended to be very high-level, most of the audience didn't understand the technical details that are an important part.
- Discussions tended to veer towards bikeshedding often.
- It was always odd to me that we tried to reach consensus to the general community, which meant a non-maintainer could object and per the rules it then went to a TSC decision.

ematipico (commented):
The new changes often use the word "maintainers" (lower case), although it isn't clear who they are. L2 + L3? Only L2? A generic term to specify anyone who contributes to Astro? I think we should be more precise.

README.md
alexanderniebuhr: Going with what Matthew said yesterday in Community Call it refers to `>= L2`, but I agree we should be precise here.

matthewp: Updated, thanks!

README.md
matthewp: I think we should avoid being overly formal. If meetings are ad hoc then there is no formal process. Creating a process makes them now be formal again. So I think spelling it out too much defeats the goal here of making RFC meetings be a loose thing that get called when someone (champion for example) wants to advance an idea.

I don't think this is a real problem in practice, though. The type of people who create RFCs and put in the effort to advance them are part of the community and know how to reach out if they want to see something advance.

ematipico (approved):

PR #905, opened 2024-04-19
Clarify champion selection

matthewp:
2 points of clarification:

- An RFC is more likely to move through stages if a maintainer is one of its champions (there can be multiple).
- The proposal author should be given a chance to volunteer as the champion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,13 +26,15 @@
 
 **Goal:** Confirm proposal feasibility with Astro Maintainers and the [Technical Steering Committee (TSC)](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#technical-steering-committee-tsc).
 
-**Requirements:** An existing proposal (Stage 1). In addition, a proposal is more likely to be accepted if it is detailed and well thought-out, can demonstrate community interest, has at least one champion volunteer, and has buy-in/interest from Astro maintainer(s).
+**Requirements:** An existing proposal (Stage 1). In addition, a proposal is more likely to be accepted if it is detailed and well thought-out, can demonstrate community interest, has at least one champion volunteer, has a maintainer asking as a champion, and has buy-in/interest from Astro maintainer(s).
 
 **Location:** GitHub Issues [(see all accepted proposals).](https://github.com/withastro/roadmap/issues)
 
 **What to Expect:** A proposal reaches this stage (aka "is accepted") during a meeting with Maintainers and TSC, following our existing [RFC Proposal](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#voting-rfc-proposals) voting process.
 
-When a proposal is accepted, a TSC member will create a new GitHub Issue summarizing the original proposal using our official template. At this point, the proposal champion is free to move on to the next stage. If a champion doesn't exist yet, then an accepted proposal may remain open until a champion volunteers by posting in the GitHub Issue.
+When a proposal is accepted, a TSC member will create a new GitHub Issue summarizing the original proposal using our official template.
+
+The proposal's original author should be given a chance to volunteer to be its champion. If they refuse, someone else can offer for the position. If a champion doesn't exist yet, then an accepted proposal may remain open until a champion volunteers by posting in the GitHub Issue. Once a champion is selected, they are free to move on to the next stage.
 
 In some cases, a proposal may be explicitly rejected by TSC if it is known to be infeasible, or go against some existing goals/mission of the project. In the event of an explicit rejection, a TSC member will comment on to the proposal explaining the reasoning for rejection.
 
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@
 
 **Goal:** Confirm proposal feasibility with Astro Maintainers and the [Technical Steering Committee (TSC)](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#technical-steering-committee-tsc).
 
-**Requirements:** An existing proposal (Stage 1). In addition, a proposal is more likely to be accepted if it is detailed and well thought-out, can demonstrate community interest, has at least one champion volunteer, has a maintainer asking as a champion, and has buy-in/interest from Astro maintainer(s).
+**Requirements:** An existing proposal (Stage 1). In addition, a proposal is more likely to be accepted if it is detailed and well thought-out, can demonstrate community interest, has at least one champion volunteer, has a maintainer acting as a champion, and has buy-in/interest from Astro maintainer(s).
 
 **Location:** GitHub Issues [(see all accepted proposals).](https://github.com/withastro/roadmap/issues)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,13 +26,15 @@
 
 **Goal:** Confirm proposal feasibility with Astro Maintainers and the [Technical Steering Committee (TSC)](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#technical-steering-committee-tsc).
 
-**Requirements:** An existing proposal (Stage 1). In addition, a proposal is more likely to be accepted if it is detailed and well thought-out, can demonstrate community interest, has at least one champion volunteer, and has buy-in/interest from Astro maintainer(s).
+**Requirements:** An existing proposal (Stage 1). In addition, a proposal is more likely to be accepted if it is detailed and well thought-out, can demonstrate community interest, has at least one champion volunteer, has a maintainer asking as a champion, and has buy-in/interest from Astro maintainer(s).
```
sarah11918, line 29:
```suggestion
**Requirements:** An existing proposal (Stage 1). In addition, a proposal is more likely to be accepted if it is detailed and well thought-out, can demonstrate community interest, has at least one champion volunteer, has a maintainer acting as a champion, and has buy-in/interest from Astro maintainer(s).
```
I think typo here.

FredKSchott:
+1, LGTM! Since the section is about "things that help" it could even be worth calling out that a core maintainer as champion is an even stronger signal than a maintainer alone and no one from core involved.

README.md
```diff
@@ -26,13 +26,15 @@
...
 
 **What to Expect:** A proposal reaches this stage (aka "is accepted") during a meeting with Maintainers and TSC, following our existing [RFC Proposal](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#voting-rfc-proposals) voting process.
 
-When a proposal is accepted, a TSC member will create a new GitHub Issue summarizing the original proposal using our official template. At this point, the proposal champion is free to move on to the next stage. If a champion doesn't exist yet, then an accepted proposal may remain open until a champion volunteers by posting in the GitHub Issue.
+When a proposal is accepted, a TSC member will create a new GitHub Issue summarizing the original proposal using our official template.
+
+The proposal's original author should be given a chance to volunteer to be its champion. If they refuse, someone else can offer for the position. If a champion doesn't exist yet, then an accepted proposal may remain open until a champion volunteers by posting in the GitHub Issue. Once a champion is selected, they are free to move on to the next stage.
```
sarah11918, line 37:
```suggestion
The proposal's original author will be given a chance to volunteer to be its champion. If they refuse, someone else can volunteer for the position by posting in the GitHub Issue. An accepted proposal will remain an open issue until a champion is confirmed by TSC. Once confirmed, the champion is free to advance the proposal to the next stage.
```
Just adding some explicitness for consideration!

sarah11918 (commented):
Thanks @matthewp ! Typo, and a suggestion for laying things out a little more explicitly!

README.md
```diff
@@ -26,13 +26,15 @@
...
+**Requirements:** An existing proposal (Stage 1). In addition, a proposal is more likely to be accepted if it is detailed and well thought-out, can demonstrate community interest, has at least one champion volunteer, has a maintainer acting as a champion, and has buy-in/interest from Astro maintainer(s).
 
 **Location:** GitHub Issues [(see all accepted proposals).](https://github.com/withastro/roadmap/issues)
 
 **What to Expect:** A proposal reaches this stage (aka "is accepted") during a meeting with Maintainers and TSC, following our existing [RFC Proposal](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#voting-rfc-proposals) voting process.
 
-When a proposal is accepted, a TSC member will create a new GitHub Issue summarizing the original proposal using our official template. At this point, the proposal champion is free to move on to the next stage. If a champion doesn't exist yet, then an accepted proposal may remain open until a champion volunteers by posting in the GitHub Issue.
+When a proposal is accepted, a TSC member will create a new GitHub Issue summarizing the original proposal using our official template.
```
ematipico, line 35: I think we should change this paragraph, and say that a "TSC member **might** delegate this task to the champion of the proposal, if there's one". This will actually reflect what we have been doing recently.

README.md
```diff
@@ -26,13 +26,15 @@
...
+
+The proposal's original author will be given a chance to volunteer to be its champion. If they refuse, someone else can volunteer for the position by posting in the GitHub Issue. An accepted proposal will remain an open issue until a champion is confirmed by TSC. Once confirmed, the champion is free to advance the proposal to the next stage.
```
ematipico, line 37: I would be more precise and say that a champion "is confirmed" when the Stage 2 template is updated to reflect this.

README.md
```diff
@@ -26,13 +26,15 @@
 
 **Goal:** Confirm proposal feasibility with Astro Maintainers and the [Technical Steering Committee (TSC)](https://github.com/withastro/.github/blob/main/GOVERNANCE.md#technical-steering-committee-tsc).
 
-**Requirements:** An existing proposal (Stage 1). In addition, a proposal is more likely to be accepted if it is detailed and well thought-out, can demonstrate community interest, has at least one champion volunteer, and has buy-in/interest from Astro maintainer(s).
+**Requirements:** An existing proposal (Stage 1). In addition, a proposal is more likely to be accepted if it is detailed and well thought-out, can demonstrate community interest, has at least one champion volunteer, has a maintainer acting as a champion, and has buy-in/interest from Astro maintainer(s).
```
ematipico, line 29: It could be a nitpick, but here we use the term maintainer (lower case), but it's a generic term and a bit foggy. Who's an Astro maintainer? Did we want to use the L2 level from the governance?

README.md
matthewp: Currently a champion is not required to reach stage 2 and the text says this. Are you suggesting that we change it so that there is one? 

I could see wanting to do that, but the downside is some times a champion can't work on something right away, for example @natemoo-re proposal for the Container API. Maybe that was a past mistake we can avoid in the future though.

README.md
matthewp: Yes, I did mean L2 maintainer. Should I put that in here?

README.md
ematipico: Oh, I see now; I misunderstood things. However, this part is confusing: 

> Once confirmed, the champion is free to advance the proposal to the next stage.

Does that mean that Stage 2 proposals move forward only when they have a champion, meaning that the TSC doesn't need to be involved anymore?

README.md
ematipico: I think so, at least the first time we use the word

ematipico (approved):

README.md
matthewp: A proposal can't move to stage 3 (RFC PR) without a champion, yes. But it can move to stage 2. TSC is still involved approving moving to the stages beyond that.

ematipico:
I think it's safe to merge Matthew

Issue #898, opened 2024-04-17
Actions


bholmesdev:
### Body

# Summary

Let's make handling form submissions easy and type-safe both server-side and client-side.

# Goals

- Form actions can accept JSON or FormData payloads.
- **You no longer need boilerplate to safely parse the request body** based on the `Content-Type`.
- **You no longer need boilerplate to retrieve form data values** from the request body. The action should be able to enumerate expected input names, and the handler should be able to retrieve these values without a type cast. In other words, no more bangs `!` or `as string` casts as in the example `formData.get('expected')! as string`.
- **You can call an action using a standard HTML `form` element** with the `action` property.
- **You can use client JavaScript to call an action** using scripts or islands. When doing so, data returned by the action is type-safe without type casting. Note: This should consider form handling in popular component frameworks, like the [`useActionState()`](https://react.dev/reference/react-dom/hooks/useFormState) and [`useFormStatus()`](https://react.dev/reference/react-dom/hooks/useFormStatus) hooks in React 19.
- **You can declare actions as an endpoint** to call from prerendered / static pages using the `hybrid` output.

## Non-goals

- A solution to client-side validation. Validation is a major piece to forms with several community libraries to choose from (ex. react-hook-form). Astro may recommend standard HTML attributes for client validation including the `required` and `type` properties on an input.
- A hook to retrieve the return value of an action from Astro frontmatter. Action return values may be accessed from client JavaScript for optimistic updates. Otherwise, we expect actions to persist any state to a database to be retrieved manually from your Astro components.
- Declaring actions within `.astro` frontmatter. Frontmatter forms a function closure, which can lead to misuse of variables within an action handler. This challenge is shared by `getStaticPaths()` and it would be best to avoid repeating this pattern in future APIs.

# Background & Motivation

Form submissions are a core building block of the web that Astro has yet to form an opinion on (pun intended).

So far, Astro has been rewarded for waiting on the platform and the Astro community to mature before designing a primitive. By waiting on view transitions, we found a SPA-like routing solution grounded in native APIs. By waiting on libSQL, we found a data storage solution for content sites and apps alike. Now, we've waited on other major frameworks to forge new paths with form actions. This includes Remix actions, SvelteKit actions, React server actions, and more.

At the same time, Astro just launched its database primitive: Astro DB. This is propelling the Astro community from static content to more dynamic use cases:

- A like button that updates a database counter
- A comments section that inserts a new comment behind an auth gateway
- A newsletter form that pings a third party mailing service

[...]

PR #912, opened 2024-05-02
Actions

bholmesdev:
# Summary

Astro actions make it easy to define and call backend functions with type-safety.

```ts
// src/actions/index.ts
import { defineAction, z } from "astro:actions";

export const server = {
  like: defineAction({
    // accept json
    input: z.object({ postId: z.string() }),
    handler: async ({ postId }, context) => {
      // update likes in db

      return likes;
    },
  }),
};

// src/components/Like.tsx
import { actions } from "astro:actions";
import { useState } from "preact/hooks";

export function Like({ postId }: { postId: string }) {
  const [likes, setLikes] = useState(0);
  return (
    <button
      onClick={async () => {
        const newLikes = await actions.like({ postId });
        setLikes(newLikes);
      }}
    >
      {likes} likes
    </button>
  );
}
```

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/actions/proposals/0046-actions.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/898)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/871)

### DIFF
--- a/proposals/0046-actions.md
+++ b/proposals/0046-actions.md
@@ -5,7 +5,7 @@ Start Date: 2024-05-01
 
 # Summary
 
-Let's make backend functions easy to write and easy to call with type safety.
+Astro actions make it easy to define and call backend functions with type-safety.
 
 # Example
 
--- a/proposals/0046-actions.md
+++ b/proposals/0046-actions.md
@@ -53,7 +53,7 @@ export function Like({ postId }: { postId: string }) {
   return (
     <button
       onClick={async () => {
-        const newLikes = await actions.like({ postId });
+        const newLikes = await actions.like.orThrow({ postId });
         setLikes(newLikes);
       }}
     >
@@ -68,7 +68,7 @@ export function Comment({ postId }: { postId: string }) {
       onSubmit={async (e) => {
         e.preventDefault();
         const formData = new FormData(e.target);
-        const result = await actions.blog.comment(formData);
+        const result = await actions.blog.comment.orThrow(formData);
         // handle result
       }}
     >
@@ -141,7 +141,7 @@ Now, this action is callable from client components and server forms.
 
 ### Call actions from a client component
 
-To call an action from a client component, you can import the `actions` object from `astro:actions`. This will include `like()` as a function, which accepts type-safe arguments as a JSON object.
+To call an action from a client component, you can import the `actions` object from `astro:actions`. This will include `like()` as a function, which accepts type-safe arguments as a JSON object. The return value will be an object containing either `data` with the action result, or `error` with any validation error or custom exception.
 
 Implementation: When imported on the client, the `actions` object will _not_ expose server code. It will instead expose a proxy object that calls a given action using a `fetch()` call. The object path will mirror the request URL that is fetched. For example, a call to `actions.like()` will generate a fetch to the URL `/_actions/like`. For nested objects like `actions.blog.like()`, dot chaining is used: `/_actions/blog.like`. The request is routed to your action using an injected handler at `/_actions/[...path]`.
 
@@ -157,8 +157,10 @@ export function Like({ postId }: { postId: string }) {
   return (
     <button
       onClick={async () => {
-        const newLikes = await actions.like({ postId });
-        setLikes(newLikes);
+        const { data, error } = await actions.like({ postId });
+        if (!error) {
+          setLikes(data);
+        }
       }}
     >
       {likes} likes
@@ -166,6 +168,14 @@ export function Like({ postId }: { postId: string }) {
   );
 }
 ```
+
+To retrieve `data` directly and allow errors to throw, you can add `orThrow()` following the format `actions.NAME.orThrow(input)`:
+
+```ts
+const newLikes = await actions.like.orThrow({ postId });
+setLikes(newLikes);
+```
+
 
 ## Handle form requests
 
@@ -237,26 +247,28 @@ import { useActionState } from 'react';
 import { experimental_withState } from '@astrojs/react/actions';
 
 export function Like({ postId }: { postId: string }) {
-	const [state, action, pending] = useActionState(
+	const [{ data, error }, action, pending] = useActionState(
 		experimental_withState(actions.like),
-		0, // initial likes
+		{ data: 0 }, // initial likes
 	);
 
 	return (
 		<form action={action}>
 			<input type="hidden" name="postId" value={postId} />
-      <button disabled={pending}>{state} ❤️</button>
+      <button disabled={pending}>{data} ❤️</button>
 		</form>
 	);
 }
 ```
+
+Note: Chaining `orThrow()` on a function will _disable_ progressive enhancement. This is because uncaught exceptions must be handled from the client.
 
 Implementation: `withState()` updates an action to match the expected signature for `useActionState()`: `(state, formData) => any`. The `state` field is applied as a stringified `formData` property to retrieve in your action `handler` as-needed. `withState()` also ensures progressive enhancement metadata is applied.
 
 You can also access the state stored by `useActionState()` from your action `handler`. Call `experimental_getActionState()` with the API context, and optionally apply a type to the result:
 
 ```ts
-import { defineAction, z } from 'astro:actions';
+import { defineAction, type ActionError, z } from 'astro:actions';
 import { experimental_getActionState } from '@astrojs/react/actions';
 
 export const server = {
@@ -265,9 +277,9 @@ export const server = {
       postId: z.string(),
     }),
     handler: async ({ postId }, ctx) => {
-      const currentLikes = experimental_getActionState<number>(ctx);
+      const state = experimental_getActionState<{ data: number } | { error: ActionError}>(ctx);
       // write to database
-      return currentLikes + 1;
+      return state.data + 1;
     }
   })
 }
@@ -294,7 +306,7 @@ export function SignOut() {
 
 Actions work well when client JavaScript is enabled. The are also automatically enhanced for zero-JS scenarios using React 19 form actions. But in other contexts, you may see unexpected behavior when a form is submitted _before_ your component's JavaScript loads. This is common on slow internet connections or older devices. To offer a fallback experience, you can add a progressive fallback to your form.
 
-To add a fallback, add the `method="POST"` property to your `<form>` element. Then, call the `getActionProps()` function from `astro:actions` with the action you want to use (ex. `getActionProps(actions.comment)`). Spread the result onto an `input` within your form to apply metadata for Astro to handle the request:
+To add a fallback, add the `method="POST"` property to your `<form>` element. Then, apply the action function directly to the `action` property of the form. This will apply the function name as a query string to be handled by the server:
 
 ```tsx
 import { actions, getActionProps } from 'astro:actions';
@@ -302,20 +314,27 @@ import { actions, getActionProps } from 'astro:actions';
 // src/components/Comment.jsx
 export function Comment({ postId }: { postId: string }) {
 	return (
-		<form method="POST" onSubmit={...}>
-			<input {...getActionProps(actions.comment)} />
-			{/* result:
-			<input
-				type="hidden"
-				name="_astroAction"
-				value="/_actions/comment" />
-			*/}
+		<form method="POST" action={actions.comment} onSubmit={...}>
+			{/* output: action="?_astroAction=comment" */}
 		</form>
 	)
 }
 ```
 
-Implementation: The `getActionProps()` function returns hidden input attributes to tell Astro which action should handle the request. Using middleware, Astro will intercept incoming requests with a form data `Content-type`. The form data is parsed to check for the `_astroAction` field and call the appropriate action.
+Implementation: Astro injects middleware to check for the `_astroAction` param and call the appropriate action.
+
+You may also construct form action URLs using string concatenation, or by using the `URL()` constructor, with the an action's `.queryString` property:
+
+```astro
+---
+import { actions } from 'astro:actions';
+const confirmationUrl = new URL('/confirmation', Astro.url);
+confirmationUrl.search = actions.queryString;
+---
+<form method="POST" action={confirmationUrl.pathname}>
+  <button>Submit</button>
+</form>
+```
 
 ### Handle an action result on the server
 
@@ -335,34 +354,7 @@ const comment = Astro.getActionResult(actions.comment);
 )}
 ```
 
-## Handle errors
-
-Actions can raise exceptions for any number of reasons: an invalid input, an unauthorized request, a missing resource, etc. By default, these errors will `throw` wherever your action is called.
-
-To handle errors in your code without using a `try / catch`, you can chain the `.safe()` function onto your existing actions. This will change the return type to either a `data` object with the result, or an `error` object with the exception.
-
-You can call `.safe()` from a client component to handle `data` and `error`:
-
-```tsx
-// src/components/Like.tsx
-import { actions } from 'astro:actions';
-
-export function Like() {
-	return (
-		<button onClick={async () => {
-			const { data, error } = await actions.like.safe({ postId });
-			if (data) // set state
-			// otherwise, handle the error
-		}}>
-			{likes} likes
-		</button>
-	)
-}
-```
-
-You can also handle the `error` object returned by `Astro.getActionResult()`.
-
-### Input validation errors
+## Input validation errors
 
 Your `input` schema gives you type safety wherever you call your action. Still, you may have further refinements in your Zod schema that can raise a validation error at runtime.
 
@@ -382,7 +374,7 @@ export function Comment({ postId }: { postId: string }) {
       onSubmit={async (e) => {
         e.preventDefault();
         const formData = new FormData(e.target);
-        const { error } = await actions.blog.comment.safe(formData);
+        const { error } = await actions.blog.comment(formData);
         if (error && isInputError(error)) {
           setBodyError(error.fields.body);
         }
@@ -424,7 +416,7 @@ export const server = {
 };
 ```
 
-To handle this error, you can call your action using the `.safe()` extension and check whether an `error` is present. This property will be of type `ActionError`.
+To handle this error, you can call your action and check whether an `error` is present. This property will be of type `ActionError`.
 
 ## Access API Context
 
--- a/proposals/0046-actions.md
+++ b/proposals/0046-actions.md
@@ -255,7 +255,7 @@ Implementation: The `getNameProps()` function returns hidden input attributes to
 
 ### Handle an action result on the server
 
-When using a progressive fallback, you can get the result of an action from your Astro frontmatter. Call `Astro.getActionResult()` with the action you want (ex. `Astro.getActionResult(actions.comment)`). This will return type-safe data when a matching POST request is received and `undefined` otherwise.
+When using a progressive fallback, you can get the result of an action from your Astro frontmatter. Call `Astro.getActionResult()` with the action you want (ex. `Astro.getActionResult(actions.comment)`). This will return type-safe `data` or `error` objects when a matching POST request is received and `undefined` otherwise.
 
 ```astro
 ---
@@ -264,7 +264,7 @@ import { actions } from 'astro:actions';
 const comment = Astro.getActionResult(actions.comment);
 ---
 
-{comment && (
+{comment?.data && (
 	<article class="new-comment">
 		{/* ... */}
 	</article>
@@ -292,21 +292,11 @@ export function Like() {
 		}}>
 			{likes} likes
 		</button>
-	)487918
+	)
 }
 ```
 
-You can also use `.safe` when getting the action result from your Astro frontmatter. Add `.safe` to the end of your action when passing to `Astro.getActionResult()`:
-
-```astro
----
-import { actions } from 'astro:actions';
-
-const { data, error } = Astro.getActionResult(actions.like.safe);
-if (data) // handle result
-// otherwise, show an error message
----
-```
+You can also handle the `error` object returned by `Astro.getActionResult()`.
 
 ### Input validation errors
 
--- a/proposals/0046-actions.md
+++ b/proposals/0046-actions.md
@@ -224,11 +224,75 @@ export function Comment({ postId }: { postId: string }) {
 }
 ```
 
-### Add progressive enhancement with fallbacks
+### Call actions using React 19 form actions
+
+Astro actions are fully compatible with React 19 form actions. This includes automatic enhancement to submit your form in zero-JS scenarios. To use React 19 in your Astro project, [follow the React 19 beta installation guide.](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#installing)
+
+This example uses an Astro Action called `like()` that accepts a `postId` and returns the number of likes. With React 19, you can pass this action to the `useActionState()` hook to track the result. Apply the `experimental_withState()` function (to be renamed to `withState()` when stable) to apply progressive enhancement information:
+
+```tsx
+// src/components/Like.tsx
+import { actions } from 'astro:actions';
+import { useActionState } from 'react';
+import { experimental_withState } from '@astrojs/react/actions';
+
+export function Like({ postId }: { postId: string }) {
+	const [state, action, pending] = useActionState(
+		experimental_withState(actions.like),
+		0, // initial likes
+	);
+
+	return (
+		<form action={action}>
+			<input type="hidden" name="postId" value={postId} />
+      <button disabled={pending}>{state} ❤️</button>
+		</form>
+	);
+}
+```
+
+Implementation: `withState()` updates an action to match the expected signature for `useActionState()`: `(state, formData) => any`. The `state` field is applied as a stringified `formData` property to retrieve in your action `handler` as-needed. `withState()` also ensures progressive enhancement metadata is applied.
+
+You can also access the state stored by `useActionState()` from your action `handler`. Call `experimental_getActionState()` with the API context, and optionally apply a type to the result:
+
+```ts
+import { defineAction, z } from 'astro:actions';
+import { experimental_getActionState } from '@astrojs/react/actions';
+
+export const server = {
+  like: defineAction({
+    input: z.object({
+      postId: z.string(),
+    }),
+    handler: async ({ postId }, ctx) => {
+      const currentLikes = experimental_getActionState<number>(ctx);
+      // write to database
+      return currentLikes + 1;
+    }
+  })
+}
+```
+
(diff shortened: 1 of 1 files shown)


rishi-raj-jain:
Does this require View Transitions to be enabled like for form data processing?

Adammatthiesen:
I want every bit of this for StudioCMS....

bholmesdev:
@rishi-raj-jain Not at all! We are targeting client components to show how JS state management could work. But forms are callable from an `form` element with or without view transitions. Let me know if this section addresses that concern: https://github.com/withastro/roadmap/blob/actions/proposals/0046-actions.md#add-progressive-enhancement-with-fallbacks

bholmesdev:
@jdtjenkins Ah, good thing to note. Since actions allow you define backend handlers, you'll need a server adapter with `output: 'hybrid'`. You can keep all your existing routes static with that output

ascorbic:
My turn to bikeshed you. I think `getNameProps` is a bit of a confusing name to me. Would `getInputProps` work better?

bholmesdev:
@ascorbic That's fair! I agree that's a better name, happy to use it.

proposals/0046-actions.md
```diff
@@ -0,0 +1,637 @@
...
+        .get();
+      return likes;
+    },
+  }),
+};
+```
+
+Now, this action is callable from client components and server forms.
+
+> **@bholmesdev Note:** Users will call actions by importing an `actions` object from the `astro:actions` virtual module. To avoid misleading auto-import suggestions, we use the name `server` instead of `actions` for server code.
```
matthewp, line 140: Would using the `default` export have this problem? I think `default` would be preferred for API consistency reasons, the Astro config is a default export, renderer APIs use a default export, etc. I think that Content Collections is the exception here.

Unless there's a problem with default exports I'm not aware of.

proposals/0046-actions.md
```diff
@@ -0,0 +1,637 @@
...
+		{/* ... */}
+	</article>
+)}
+```
+
+## Handle errors
+
+Actions can raise exceptions for any number of reasons: an invalid input, an unauthorized request, a missing resource, etc. By default, these errors will `throw` wherever your action is called.
+
+To handle errors in your code without using a `try / catch`, you can chain the `.safe()` function onto your existing actions. This will change the return type to either a `data` object with the result, or an `error` object with the exception.
```
matthewp, line 278: Reading this section I can't help but think that I always want to use `safe()`. APIs can fail, especially when validation is involved. Is there a strong use-case for every using the non-safe version of this?

proposals/0046-actions.md
```diff
@@ -0,0 +1,637 @@
...
+			if (data) // set state
+			// otherwise, handle the error
+		}}>
+			{likes} likes
+		</button>
+	)487918
+}
+```
+
+You can also use `.safe` when getting the action result from your Astro frontmatter. Add `.safe` to the end of your action when passing to `Astro.getActionResult()`:
```
matthewp, line 299: Echoing my above comments on `safe()`, it feels even more necessary here. `getActionResult` is a way to prevent needing to check the method, etc. It allows you to keep your code flat and only operate if it returns a result. So having a throwing-form makes the whole thing way less nice to me.

proposals/0046-actions.md
bholmesdev: That's fair! Though with JSON APIs, I've found it's rare that I want to handle validation exceptions since I trust the type inference. I also might return `undefined` instead of raising my own `ActionError`. It seems a bit obtuse to require users to _always_ check that `data` exists for every call to an action.

proposals/0046-actions.md
matthewp: I think avoiding auto-import collisions is a good reason, so I'm fine with sticking with `server` or coming up with another name.

proposals/0046-actions.md
matthewp: I didn't think about `data` having a type of `returnType | undefined` which makes me think differently about this. If you trust your APIs then this becomes a bit of a burden. In which case I don't mind the `.safe()` as much.

proposals/0046-actions.md
matthewp: So given the above conservation, does `Astro.getActionResult()` currently throw? It doesn't say that it does.

bholmesdev: Yes, `getActionResult()` would throw the same way as the client. I agree `.safe` would be a more preferred pattern on the server, since you're more likely wanting to respect validation errors on the server. Still, wanted to keep the two APIs consistent. Do you have thoughts?

matthewp: I don't see them as being the same API, so I don't think consistency doesn't feel as important. In the case of the RPC you are:

1. Calling an action.
2. Getting the result of that action.

And `getActionResult` is only (2). It's even in the name `getActionResult`, *getting* something shouldn't throw, imo. The result is either data or an error, and you are just requesting that info.

bholmesdev: This is a good point. The real "smell" for me was the inconsistent throwing behavior. The action would not throw _unless_ you call `getActionResult()`. If throwing is really what we want, it should throw on the backend when it's unhandled as well.

I agree the name should imply it's safe only. I'm on board with flipping that behavior

zadeviggers:
Hey @bholmesdev, it's me back again to make sweeping criticisms about one of your big proposals again 😁 (I'm the one who got you to vendor Zod back when you were working on content collections). Also, sorry for being so late to submit on this - I only just found out about this proposal from the 4.8 release!

Reading through the proposal, I noticed almost every example uses the `onSubmit` event of the form and then cancels the actual submit event. The use of JavaScript is understandable for many use cases, but it lacks a nice API for submitting actions without JS. This functionality is already very possible using an API route and regular form submissions. It might drive some people to use JS when they don't really need to since there isn't an obvious API for doing form submissions without it. In that light, adding an elegant path to JS-free forms through the actions API would be nice, even if it isn't a main goal.

Here's [the single example](https://github.com/withastro/roadmap/blob/actions/proposals/0046-actions.md#add-progressive-enhancement-with-fallbacks) of how to implement actions without JS:
```ts
import { actions, getActionProps } from 'astro:actions';

export function Comment({ postId }: { postId: string }) {
	return (
		<form method="POST" onSubmit={...}>
			<input {...getActionProps(actions.comment)} />
			{/* result:
			<input
				type="hidden"
				name="_astroAction"
				value="/_actions/comment" />
			*/}
		</form>
	)
}
```

It seems like a bit of an afterthought, to be honest, which doesn't really fit with Astro's whole 'Ship less JS' thing, and I hope we can all agree that `<input {...getActionProps(actions.comment)} />` looks pretty ugly.

My understanding is that under the hood, Astro sees the `_astroAction` field and rewrites the request to go to an action handler. If that's the case, why not make this a first-class behaviour? Provide another function, say `getActionURL(actions.whatever)`, that just returns the path to the action handler. Then you can pass that string into the `action` attribute of a `<form>` element. This eliminates the jankiness of `<input {...getActionProps(actions.comment)} />` while still allowing full client-side JS control when wanted and helps guide people not to use JS when they don't need to.

E.g.
```ts
import { actions, getActionURL } from 'astro:actions';

export function Comment({ postId }: { postId: string }) {
	return (
		<form method="POST" action={getActionURL(actions.comment)} onSubmit={...}>
			{/* Regular form goes here */}
		</form>
	)
}
```

Some people wanted something similar to this that involved directly importing actions with import attributes, but it seems like that wasn't viable. This seems like a nice middle ground, where you still (kinda) pass the handler to the `action` prop, and it seems much more possible: it's just a simplification of `getActionProps(...)`. 

Also, the rest of the proposal looks great. I can't wait to have a play with some Actions!

matthewp:
@zadeviggers Form support was not an afterthought, it was the biggest part of the design and why the feature didn't ship sooner. An API like the one you're suggesting was considered but it's incomplete. What happens on the server-side after an action is executed? How do you redirect to a different page? How do you handle validation errors? @bholmesdev went through a few different designs to try and make all of those things possible but it came at the cost of making the action code much more complex.

The `getActionProps` design is nice because the form is submitted to whatever `action=""` you put in, and you are able to handle redirects / errors, etc all yourself in a normal Astro page. It gives you the freedom to do whatever you want, while keeping the Action API more a pure function-like interface.

That said, this RFC is still open so we can talk about other approaches.

bholmesdev:
Thanks for the overview @matthewp. Actually, I would not consider the `action` suggestion incomplete! An alternative may be to pass a `?queryParam` from his proposed `getActionUrl()`. This lets us use the same middleware strategy for retrieving the action result from frontmatter. We could also allow some sort of `prefix` parameter if you want to reroute the POST to another page.

This isn't _perfect_ though. The main reason for a hidden input over this was challenges with React 19. When using actions in React, they own the `action` property when you pass a function:

```ts
<form action={actions.like}>
```

But now, how can we add a fallback? React stubs `action` to an empty field now. I am working with the React core team to see if we can control the server-rendered `action` to avoid this somehow. To be honest, I prefer your API to `getActionProps()`, so I would like to see this.

matthewp:
@bholmesdev I'm not sure I understand. If the `action` attribute points to the Action then the action needs to be able to do redirects and other things. Where does that code live?

bholmesdev:
@matthewp Well I should clarify: **the action would NOT be a URL. It would be a query param to re-request the current page.**

```tsx
<form action={actions.like.url}>
<!--output: ?_actionName=like-->
```

This does the same work as passing a hidden `_actionName` input from my understanding

matthewp:
Oh ok, I understand now. I still like `getActionProps` better than this idea. When I see `<form action={actions.like.url}>` I expect that the request goes directly to the action. That it actually posts to the current page would be unexpected. Also, I might want to direct the request to another page, with `getActionProps` I can do that by setting `action="/other-page"`, but with this new idea it's not clear at all how to do that.

So I think this trades away flexibility for a subjective aesthetic gain only.

zadeviggers:
@matthewp To address your flexibility concern, we can add an optional parameter to the function to specify a different path for it to submit to. I don't think this will come up super often, though, since people will probably just do all their handling in the action, and then rewrite to the page they want to render. 

@bholmesdev I really like the `actions.like.url` idea over a whole extra function. To address @matthewp's concern, it could be a function instead: `<form action={actions.like.url(/* optional path */)`

mayank99:
I think I agree with both Zade and Matthew here. I also like the `actions.like(/* optional pathname */)` suggestion (works similar to `action={undefined}` submitting to the current page).

The big advantage I can see with an extra hidden `<input>` is that developers can and will leave it out if it works without it (on their machines with high-speed internet and JS available). Progressive enhancement should ideally work by default, rather than being "opt-in".

jberglund:
I've been toying around with actions the past week and I like it a lot, especially now that it works on the server side. ❤️ 

Speaking of progressive enhancement, would it be possible to return the values of FormData through the action, based on the schema? With a default form submission and no JS to prevent the reload, the values the user entered are lost after submission.

chiubaca:
Does/will Astro have an opinionated way to re-sync backend data back with the UI after an action has been run?

I'm thinking like how in Next.js Server Actions they provide [`revalidatePath`](https://nextjs.org/docs/app/api-reference/functions/revalidatePath) and [`revalidateTag`](https://nextjs.org/docs/app/api-reference/functions/revalidateTag) functions. Is there an Astro equivalent?

(BTW I love how Astro is shaping up, I'm having a lot of fun tinkering)

Southpaw1496:
I've been experimenting with this API and enjoying it, especially the JavaScript-less option that is much cleaner than my old manual implementation. However, at the moment I can't use it with `@astrojs/cloudflare`, since it uses the `node:async-hooks` module, which isn't supported by the Cloudflare runtime. The error I receive is:
```
[ERROR] [vite] x Build failed in 105ms
[commonjs--resolver] [plugin vite:resolve] Cannot bundle Node.js built-in "node:async_hooks" imported from "node_modules/astro/dist/actions/runtime/store.js". Consider disabling ssr.noExternal or remove the built-in dependency.
```

The Cloudflare runtime does support [some Node APIs](https://developers.cloudflare.com/workers/runtime-apis/nodejs), but `async-hooks` isn't one of them. Would it be possible to re-implement whatever functionality this provides without using Node.js, so that Cloudflare users don't miss out on the new API? Or is it something that the Cloudflare adapter can change to be compatible?

chiubaca:
@Southpaw1496 Its been a bit fiddly but i'm using actions with `@astrojs/cloudflare` in my test repo here : https://github.com/chiubaca/fullstack-astro-cloudflare

Sounds like you're missing a compatibility flag, check out my [wrangler.toml](https://github.com/chiubaca/fullstack-astro-cloudflare/blob/main/wrangler.toml). Hope it helps.

wesleycoder:
I've tried to access cookies inside the action handler as it seemed to be available on the TS side, but it has no effect over the request:
```ts
import { defineAction, z } from 'astro:actions'
import { THEME_CHOICES } from '~/constants'

export const server = {
  toggleDarkMode: defineAction({
    input: z.enum(THEME_CHOICES),
    handler: (input, ctx) => {
      console.log('actions toggleDarkMode', input)
      if (input === 'system') {
        ctx.cookies.delete('theme')
      } else {
        ctx.cookies.set('theme', input)
      }
    },
  }),
}
```

Can anyone confirm if it's suposed to work or am I misguided?

bholmesdev:
@wesleycoder that could be a bug! I'll investigate

ernestoalejo:
Love the feel of simplicity of use so far, but with complex objects the type safety is not completely correct yet.

For example if the handler returns a `Date` the types indicate it will be correctly received client side, but it is serialized as a string instead because it's JSON. I'm not familiar enough with tRPC source code, but it seems [this file](https://github.com/trpc/trpc/blob/next/packages/server/src/unstable-core-do-not-import/clientish/serialize.ts) is responsible for that mapping.

As part of that issue: would it be possible to configure a custom serializer like `superjson`? tRPC [recommends it](https://trpc.io/docs/server/data-transformers) to serialize complex objects like the previous Date example and receive the correct values, but it also increases the complexity of the types.

wesleycoder:
@bholmesdev
I was looking again at my code after looking into this [new post](https://discord.com/channels/830184174198718474/1250133609713631252) on Discord about this.

I noticed that passing the `path` option to the cookie makes this work as expected
Updating my previous code it becomes:

```ts
import { defineAction, z } from ['astro:actions'](astro:actions)
import { THEME_CHOICES } from '~/constants'

export const server = {
  toggleDarkMode: defineAction({
    input: z.enum(THEME_CHOICES),
    handler: (input, ctx) => {
      console.log('actions toggleDarkMode', input)
      if (input === 'system') {
        ctx.cookies.delete('theme', { path: '/' })
      } else {
        ctx.cookies.set('theme', input { path: '/' })
      }
    },
  }),
}
```

Notice the same occurs for the `cookies.delete` method.
I believe this could help narrow down the issue for you, it seems related to default options behavior.

firxworx:
I added some comments/feedback to the previous stage e.g. https://github.com/withastro/roadmap/issues/898#issuecomment-2189826220 and it looks like there are other more recent comments from others there as well after that one was closed.

I'll note here a few highlights 

**isInputError(..) issues**

Importantly `isInputError()` input types are too narrow and either it or another guard exported alongside it should accept input type `unknown` to narrow to `ActionInputError`. An example use-case and discussion explaining the importance of this is at the link.

While more minor I do feel fairly strongly about the misleading name of `isInputError` and that type guards should be explicitly named what they check vs. something arbitrary (i.e. it should be named `isActionInputError(..)`) this will benefit new developers learning the actions API and overall readability and maintainability of not just Astro but any apps written using it.

**Please stop with the z's**

Please stop exporting `z` from everywhere under the sun in Astro (zod). There should only be one source. It is no more "convenient" especially with modern IDE's its confusing. It led me to believe the `z` for actions might be special and I was informed it wasn't and that I wasn't the only one asking the question.

When devs use Astro and hit `z` with autocomplete IDE they're faced with a growing wall of them and its not clear at all what the importance is (or the fact that it isn't.. unless I'm mistaken.. but then that's confusing too).

I get if you want to export the Astro version to help ensure versions of zod are consistent in a codebase, but this should only be done in one place, arguably the _original_ place, and with the _original_ explanation in the docs to go with it.

firxworx:
Now I read the actual code for `isInputError`. 

```ts
export function isInputError<T extends ErrorInferenceObject>(
	error?: ActionError<T>
): error is ActionInputError<T> {
	return error instanceof ActionInputError;
}
```

That's it! Not the safest of types but I'll take it and it also means its trivial to make it accept a broader type as input.

I can confirm that client-side `useMutation()` onError callback (react-query) returns `true` for this check.

```ts
  const { isPending, mutateAsync } = useMutation(
    {
      mutationFn: actions.products.update,
      onSuccess: (_data) => {
        rqc.invalidateQueries({ queryKey: KEYS.products.all() })
        onSaveSuccess?.()
      },
      onError: (error) => {
        // the `error` is typed "Error" which is incompatible with `isInputError`
        // confirming that this check returns `true` from client-side code 
        console.log(`there was an error and it is an actioninputerror: ${error instanceof ActionInputError}`)
      },
   },
   queryClient,
  }
```

wesleycoder:
It seems the cookies issue is not restricted to the actions context as a [similar post](https://discord.com/channels/830184174198718474/1257220653627473920/1257220653627473920) has been made on Discord.
Should we open a new issue for that or is is already being handled somewhere I couldn't find?

bholmesdev:
Thanks for following up @wesleycoder! In hindsight, I should've suggested an issue instead of tracking myself. I knew I would be stepping away from actions for the following few weeks, sorry about that.

Since I said I'd take a look, my goal is to wake up tomorrow, get a cup of coffee, and try to PR a fix.

bholmesdev:
Haven't missed your comments @firxworx! All valid concerns here. There's some nuance to the API choices I could detail, but I'm not 100% sure on them. I'll take a second look to see what we can improve.

bholmesdev:
Missed your note on serializing complex objects @ernestoalejo. You're correct that we only allow JSON serializable inputs, though we don't document this in the RFC. We avoided superjson to keep the client-side of actions as lean as possible. Alternatives like [devalue](https://www.npmjs.com/package/devalue)  or [seroval](https://github.com/lxsmnsyc/seroval) may offer the same utility in a smaller footprint. I agree dates are fairly common to send over the wire. We can look into this.

firxworx:
`server.d.ts` export does not export the type `Handler`. 

In general I have found that TypeScript libraries that provide a smooth DX and avoid manufacturing headaches/issues generally export pretty much all of their types + interfaces even ones that seem internal or its hard to think of a use-case.

Especially with actions being new, to enable the community to work with them in ways that perhaps weren't foreseen and have low friction to building helpers + utilities without resorting to fake/wrong types with `as` casts or advanced typescript utilities + trickery to extract required types.

Example use-case: I thought perhaps I'd make a generic wrapper for handlers e.g. `withErrorHandling(...)` that maintains the type-safety and catches different errors including from my database client and maps them to the desired ActionError to eliminate try/catch repetition and clutter. There are many others I can think of e.g. sanitization, transformations, etc.

I'm not blocked per se but I was surprised to find some of the potentially relevant types aren't exported.

firxworx:
`ActionErrorCode` isn't exported either! Another real case of writing a simple helper function is foiled!

I won't keep posting since you got the general spiel already above however this time I have something to add:

Beyond exporting it, I suggest as a general philosophy for the majority of cases when you have a type like `type ActionErrorCode = 'BAD_REQUEST' | 'UNAUTHORIZED' | ...` that a const assertion (`as const`) be exported as well:

`export const ACTION_ERROR_CODES = ['BAD_REQUEST', 'UNAUTHORIZED', ...] as const`

And when you define the type: `export type ActionErrorCode = (typeof ACTION_ERROR_CODES )[number]`

This opens up worlds of magic for downstream devs who now can write helpers and use the validation libraries of their choice (e.g. zod's `enum()` accepts const assertions, valibot), etc, and they can make a copy (since its `as const`) to find/filter/etc. from. When writing a library or framework its a choice that enables rather than restricts.

bholmesdev:
Good point on utility exports @firxworx. There is a method to a madness: exposing internal utilities means an earlier maintenance burden. We could exposing all of our internals for library authors, but this subjects them to documentation, versioning, and caution with breaking changes. We prefer a more cautious route to only expose utilities when there is a strong user need, like what you've presented here.

Happyt o expose an `ASTRO_ERROR_CODES` const! I think that makes a lot of sense.

bholmesdev:
[PR is now up](https://github.com/withastro/astro/pull/11439) to expand `isInputError` to accept unknown error types. There _is_ a purpose for the `ErrorInferenceObject` that we needed to keep in-tact. I've [added an in-line comment](https://github.com/withastro/astro/pull/11439/files#diff-1811db2c2287c0b2d3c30ff42c6e7ab1286c8519858cf9a418989331c817fbc0R44) to explain the nuance if you're curious @firxworx

bholmesdev:
Big news! We've reached consensus on those API choices above.
- `action={actions.name}` is **the new way** to handle plain POST requests from forms. Shipped in the latest release. ✅ 
- `.safe()` is **now the default** for all action calls. To allow exceptions to throw, you can chain `orThrow()` to your function call. Shipped in the latest release ✅ 
- `z` will **now come from the `astro:schema` module.** This will become the single place to import `z` across Actions and Content Collections. `z` will stick around on the `astro:content` module marked deprecated for 5.0 in the coming months. Since Actions is experimental, we can remove from `astro:actions` today. PR inbound ⌛ 

We also introduced a change to **remove async local storage** from the Actions internals. This means Cloudflare and Stackblitz work out of the box, no build config needed.

Now, we're leaving 2 more weeks as experimental to get final thoughts on these changes. If all goes well, we expect Actions to be baselined as **stable** in the next minor (4.14.0) two weeks from now.

Before then, I'm opening a call for consensus on this RFC. If you have thoughts, please try the latest release and leave feedback here. If there is no contention, my goal is to close this RFC as accepted next Tuesday. 

cc to our active contributors: @jdtjenkins @wesleycoder @ernestoalejo @chiubaca @Southpaw1496 @mayank99 @zadeviggers

proposals/0046-actions.md
```diff
@@ -0,0 +1,707 @@
...
+
+## Call actions directly from server code
+
+You may need to call an action handler directly from server pages and endpoints. To do so, use the `Astro.callAction()` function. Pass the function you want to call as the first parameter, and any input arguments as the second parameter:
+
+```astro
+---
+import { actions } from 'astro:actions';
+
+const result = await Astro.callAction(actions.searchPosts, {
```
mayank99, line 458: It would feel more natural to call `action.searchPosts()` directly on the server too, similar to how it's done on the client.

bholmesdev: Ah, this was removed recently due to technical limitations of injecting `APIContext`. This _did_ work in an earlier version, but it was removed due to environment issues with async local storage. I agree it's a compromise, but we unfortunately don't have alternatives today.

mayank99: Ah that's a bit unfortunate. Maybe one day!

This is a very minor issue anyway, since the big problem/use-case that actions solve is calling backend code from frontend.

bholmesdev: @mayank99 Yeah, that was my thinking. I agree that "just call it" is the dream! There's an avenue using compiler hacks that I dreamt up, but the maintenance burden was hard to justify. Open to any JS heroes that can find a path.

firxworx:
In the current release 4.13.1 the types are broken for using `orThrow` with actions that do not define an `input` in their `defineAction`, i.e. actions that essentially play the role of a "GET" request.

I discovered the issue when upgrading from 4.11.1 and revising code to accommodate `.safe()` now being the default behaviour and `.orThrow()` now being the option.

### Representative Example

Consider the following example with `@tanstack/react-query` that used to work fine in 4.11.1:

```ts
const x = useQuery({
    queryKey: QK.widgets.all(),
    queryFn: actions.widgets.getAll,
  })
```

If I revise the query function line to be `queryFn: actions.widgets.getAll.orThrow` then TypeScript will complain: _Property 'orThrow' does not exist on type '((input?: any) => Promise<SafeResult<never..._.

However... the types lie! 

If I revise to the following and force it to compile...

```ts
    queryFn: actions.patients.getAll.orThrow,
```

...the code works: the browser sends the expected request body and I get the expected response back from the server.

### Workarounds

Adding `@ts-expect-error` to the line with `.orThrow` is no good because this forces the query result type to become `unknown` and will break the point of using typed Astro actions.

A hacky workaround is adding `input` to the action definition using `z.object({})` as the schema and then calling with `.orThrow({})`. This is not desirable (and thus this is an issue that should be fixed) because:

- `{}` is a misleading type and is considered a bad practice (the following is valid TypeScript: `const x: {} = 'I am a teapot'`)
- the intent of server-side actions code is misleading because this action should have no input
- by using `.orThrow({})` on the client-side it actually sends that `{}`

### Interesting Observations

Funnily enough if add the `input: z.object({})` to the action definition and I keep the line `actions.patients.getAll.orThrow` then TypeScript will agree that "orThrow" exists however this time the result type inferred by React Query is bungled and it thinks the result is wrapped in a Promise (!!).

### Use Cases

The use-case where `.orThrow()` is required is to support React Query: it needs a rejected promise to handle errors correctly.

The use-case for having a "GET"-like action without an `input` is the desire/experiment to define all CRUD operations of server-side code together and for there to be a consistent RPC-like API on the front-end. The resulting developer experience is similar to ts-rest, hono RPC, etc, etc.

---

Admittedly without reviewing the current code it makes me wonder if the underlying code is not type safe due to type cheats / escape hatches (`as`, `any`, etc.) or else I would expect the type system to have flagged this issue and prevented this issue from being shipped. It might be upstream (even react-query or zod) but off-hand this smells like type cheats to me.

bholmesdev:
Alright, decided to keep us in experiment for _one more release._ We snuck in a request change: automatic redirects to avoid the "confirm post resubmission" dialog. You can try it out in the latest patch!

As part of this, I also overhauled the instructions for using Actions with plain HTML forms. We better explain how redirects are handled, and include some advice for success and error banners. You can find this on the RFC or the latest documentation draft:

https://deploy-preview-8980--astro-docs-2.netlify.app/en/guides/actions/#call-actions-from-an-html-form-action

Share your feedback and bug reports! If we're happy with this, we should be good for stable.

bholmesdev:
Alright y'all, it's happening. Actions is scheduled to ship _stable_ in 4.15 🥳 

There are a few pieces we didn't manage to finish, but they're still on the list. Namely:
- discriminated union support for form submissions
- encrypting action data when using `<form action={action}>`
- flashing form inputs to preserve state (`transition:persist` is a documented workaround though)

With that, I'm merging this RFC as **accepted.** Keep sharing bugs and feedback as always! Thanks everyone.

Issue #533, opened 2023-03-23
Container API: render components in isolation


Princesseuh:
### Body

- Accepted Date: 23/03/23
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/462
- Author: @natemoo-re 
- Implementation PR: No PR yet. See the [`feat/container`](https://github.com/withastro/astro/tree/feat/container) branch for an exploration of implementation.

# Summary

Astro components are tightly coupled to `astro` (the metaframework). This proposal introduces a possible server-side API for rendering `.astro` files in isolation, outside of the full `astro` build pipeline.

# Background & Motivation

Some of our own proposals, as well as many third-party tool integrations, are blocked until we expose a proper server-side rendering API for `.astro` components. Other frameworks have tools like `preact-render-to-string` or `react-dom/server` that make rendering components on the server straight-forward.

We've avoided doing this because... it's very hard! Part of the appeal of `.astro` components is that they have immediate access to a number of powerful APIs, context about your site, and a rich set of framework renderers. In the full `astro` build process, we are able to wire up all this context invisibly. Wiring up the context manually as a third-party is prohibitively complex.

# Goals

- Provide an API for rendering `.astro` components in isolation
- Expose a familiar, user-friendly API
- Surface enough control for full `astro` parity, but abstract away internal APIs
- Enable third-party tools to consume `.astro` files on the server
- Enable unit testing of `.astro` component output
- Possibly unblock `.mdx` compiledContent/html output?
- Support Astro framework renderers (`@astrojs/*`) if possible

# Example

```js
import { AstroContainer } from 'astro/container';
import Component from './components/Component.astro';

const astro = new AstroContainer();
console.log(await astro.renderToString(Component, { props, slots }))
```

The container can be optionally constructed with some settings that are typically reserved for Astro configuration.
```js
const astro = new AstroContainer({
  mode?: 'development' | 'production';
  site?: string;
  renderers?: Renderer[]
});
```

The second argument to `renderToString` optionally provides render-specific data that may be exposed through the Astro global, like `props`, `slots`, `request`, and/or `params`.
```js
await astro.renderToString(Component, { props, slots, request, params })
```

natemoo-re:
Major point to be resolved: how do we surface `styles` and `scripts`? Ideally this API would work similarly to Svelte's `Component.render()`, but Astro's compiler strips out scripts and styles to be handled by Vite. I have yet to figure out what we should do here!

JReinhold:
[...]

PR #916, opened 2024-05-06
Container APIs

ematipico:
# Summary

An API for rendering components in isolation:

```js
// Card.test.js
import Card from "../src/components/Card.astro"
import astroConfig from "../src/astro.config.mjs";

const container = await AstroContainer.create()
const response = await container.renderToString(Card);
// assertions
```

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/rfc/container-api/proposals/0048-container-api.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/533)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/462)

### DIFF
--- a/proposals/0048-container-api.md
+++ b/proposals/0048-container-api.md
@@ -69,7 +69,7 @@ That's why the first iteration of the APIs won't provide a built-in way to compi
 The API took a lot of time to land because it required some internal refactors. Thanks to these refactors, we will be able
 to land this API by using the **very same engine** that Astro uses under the hood.
 
-Eventually, we landed to a common and abstract concept called **Pipeline**. A pipeline is simply an abstract
+Eventually, we landed to a common and abstract concept called **Pipeline**. A pipeline is an abstract
 class that is responsible to render any kind of Astro route: page, endpoint, redirect, virtual.
 
 Each pipeline inside the Astro codebase (dev, build, SSR and now test) is responsible to collect the important information in different way,
--- a/proposals/0048-container-api.md
+++ b/proposals/0048-container-api.md
@@ -19,7 +19,6 @@ console.log(await container.renderToString(Component, { props, slots }))
 The container can be optionally constructed with some settings that are typically reserved for Astro configuration.
 ```js
 const container = await AstroContainer.create({
-    mode: "production",
     streaming: true,
     astroConfig: {
         site: "https://example.com",
@@ -43,16 +42,14 @@ We've avoided doing this because... it's very hard! Part of the appeal of `.astr
 # Goals
 
 - Provide a **low-level** API for rendering `.astro` components in isolation
-- Expose a familiar, user-friendly API
 - Surface enough control for full `astro` parity, but abstract away internal APIs
 - Enable third-party tools to consume `.astro` files on the server
 - Enable unit testing of `.astro` component output
-- Possibly unblock `.mdx` compiledContent/html output?
-- Support Astro framework renderers (`@astrojs/*`) if possible
+- Support Astro framework renderers (`@astrojs/*`)
 
 # Non-Goals
 
-- Provide a way to **import** `.astro` components. Users will be responsible to provide a compiled component.
+- Provide a way to **import** `.astro` components in a non-vite environment. Users will be responsible to provide a compiled component.
 - Provide a way to **import** `astro.config.(mjs|mts)` out of the box.
 
 # Detailed Design
@@ -60,7 +57,8 @@ We've avoided doing this because... it's very hard! Part of the appeal of `.astr
 ## Preface
 
 **This** RFC will have a smaller scoped compared to what users have envisioned. The reason why the scope shrunk is that 
-I want to ship a smaller feature on the surface, which will allow us, later to enhance it based on user's feedback and use cases.
+
+We want to ship a smaller feature on the surface, which will allow us, later to enhance it based on user's feedback and use cases.
 
 That's why the first iteration of the APIs won't provide a built-in way to compile Astro components.
 
@@ -146,7 +144,7 @@ import { onRequest } from "../src/middleware.js"
 
 const response = await container.renderToString(Card, {
     slots: {
-        default: await container.renderToString(CardItem).text() 
+        default: await container.renderToString(CardItem) 
     },
     request: new Request("https://example.com/blog/blog-slug", {
         headers: {}
--- a/proposals/0048-container-api.md
+++ b/proposals/0048-container-api.md
@@ -82,6 +82,7 @@ but eventually each route must be rendered using the same data, which are:
 - `AstroContainer::create`: creates a new instance of the container.
 - `AstroContainer.renderToString`: renders a component and return a string.
 - `AstroContainer.renderToResponse`: renders a component and returns the `Response` emitted by the rendering phase.
+- `AstroContainer.addServerRenderer`: to programmatically store a renderer inside the container. 
 
 ### `create` function
 
@@ -126,7 +127,7 @@ It will be possible to tweak the container APIs with options and such. The `.cre
 ```ts
 type AstroContainerOptions = {
     streaming: boolean;
-    renderers: AstroRenderer[];
+    renderers: AddServerRenderer[];
     astroConfig: AstroUserConfig;
 }
 ```
@@ -134,6 +135,39 @@ type AstroContainerOptions = {
 The `astroConfig` object is literally the same object exposed by the `defineConfig`, inside the `astro.config.mjs` file. This very configuration 
 will go under the same schema validation that Astro uses internally. This means that an invalid schema will result in an error of the `create` function.
 
+#### Add a renderer
+
+Adding a renderer can be done manually or automatically. 
+
+The automatic way is meant for static applications, or cases where the container isn't called at runtime. The renderers maintained by the Astro org will expose
+a method called `getContainerRenderer` that will return the correct information that will tell the container **how to import the renderer**. The importing of the renderer is done via a function called `loadRenderers`, exported by a new virtual module called `astro:container`: 
+
+```js
+import { getContainerRenderer as reactRenderer } from "@astrojs/react";
+import { getContainerRenderer as vueRenderer } from "@astrojs/vue";
+import { laodRenderers } from "astro:container";
+import { AstroContainer } from "astro/container";
+
+const renderers = loadRenderers([
+    reactRenderer(),
+    vueRenderer()
+]);
+const container = await AstroContainer.create({ renderers })
+```
+
+The manual way is meant for on-demand applications, or cases where the container is called at runtime or inside other "shells" (PHP, Ruby, Java, etc.). 
+The developer is in charge of **importing** the server renderer and store it inside the container via `AstroContainer.addServerRenderer`:
+
+The renderers maintained by the Astro org will ship proper types for importing server renderers.
+
+```js
+import reactRenderer from "@astrojs/react/server.js";
+import vueRenderer from "@astrojs/vue/server.js";
+
+const container = await AstroContainer.create();
+container.addServerRenderer({renderer: reactRenderer});
+container.addServerRenderer({renderer: vueRenderer});
+```
 
 ### `renderToString` and `renderToResponse` options
 
--- a/proposals/0048-container-api.md
+++ b/proposals/0048-container-api.md
@@ -4,7 +4,7 @@
 
 # Summary
 
-Astro components are tightly coupled to `Astro` (the metaframework). This proposal introduces a possible server-side API for rendering `.astro` files in isolation, outside the full `astro` build pipeline.
+Astro components are tightly coupled to `Astro` (the metaframework). This proposal introduces a server-side API for rendering `.astro` files in isolation, outside the full `astro` build pipeline.
 
 # Example
 
@@ -80,8 +80,8 @@ but eventually each route must be rendered using the same data, which are:
 ## Public APIs
 
 - `AstroContainer::create`: creates a new instance of the container.
-- `AstroContainer::renderToString`: renders a component and return a string.
-- `AstroContainer::renderToResponse`: renders a component and returns the `Response` emitted by the rendering phase.
+- `AstroContainer.renderToString`: renders a component and return a string.
+- `AstroContainer.renderToResponse`: renders a component and returns the `Response` emitted by the rendering phase.
 
 ### `create` function
 
--- a/proposals/0048-container-api.md
+++ b/proposals/0048-container-api.md
@@ -79,7 +79,13 @@ but eventually each route must be rendered using the same data, which are:
 - **RouteData**: a type that identifies a generic Astro route. It contains information such as the type of route, the associated component, etc.
 - **ComponentInstance**: The instance of an Astro component. This is the **compiled** component. Internally, an Astro component undergoes various transformations via Vite, and it eventually gets consumed by our rendering engine (if it's a page).
 
-## Create a container
+## Public APIs
+
+- `AstroContainer::create`: creates a new instance of the container.
+- `AstroContainer::renderToString`: renders a component and return a string.
+- `AstroContainer::renderToResponse`: renders a component and returns the `Response` emitted by the rendering phase.
+
+### `create` function
 
 A container is class exposed via the `astro/container` specifier. Users **shouldn't** create a container using the `new` instance, but they should use he static function `create`:
 
@@ -115,55 +121,23 @@ const container = await AstroContainer.create({
 const content = await container.renderToString(AstroComponent);
 ```
 
-## Options
+#### Options
 
 It will be possible to tweak the container APIs with options and such. The `.create` function will accept the following interface:
 
 ```ts
 type AstroContainerOptions = {
-    mode: "development" | "production";
     streaming: boolean;
-    renderers: SSRLoadedRenderer[];
+    renderers: AstroRenderer[];
     astroConfig: AstroUserConfig;
-    middleware: MiddlewareHandler
 }
 ```
 
 The `astroConfig` object is literally the same object exposed by the `defineConfig`, inside the `astro.config.mjs` file. This very configuration 
 will go under the same schema validation that Astro uses internally. This means that an invalid schema will result in an error of the `create` function.
 
-The `middleware` is the `onRequest` instance of the middleware, in case you're rendering a page that needs to trigger the middleware.
 
-
-## `vite` configuration
-
-Astro exposes already its internal `vite` configuration, so users that already use `vite` can take compile Astro components using the same their pipeline.
-
-For example, if users use `vitest`, they can add Astro's `vite` configuration by importing it from `astro/config` specifier:
-
-```js
-// vitest.config.js
-import { getViteConfig } from "astro/config";
-
-export default getViteConfig({
-    test: {}
-})
-```
-
-```js
-// Card.test.js
-import { Card } from "../src/components/Card.astro"
-import astroConfig from "../src/astro.config.mjs";
-
-const container = await AstroContainer.create({
-    astroConfig
-})
-
-const response = await container.renderToString(Card);
-// assertions
-```
-
-## `renderToStringOptions`
+### `renderToString` and `renderToResponse` options
 
 This function can accept an optional object as a second parameter:
 
@@ -171,12 +145,15 @@ This function can accept an optional object as a second parameter:
 import { onRequest } from "../src/middleware.js"
 
 const response = await container.renderToString(Card, {
-    slots: [
-        (await container.renderToString(CardItem).text()),
-        "Footer of the card"
-    ],
-    request: new Request("https://example.com/blog/blog-slug"),
-    params: ["slug"], // in case your route is `pages/blog/[slug].astro`
+    slots: {
+        default: await container.renderToString(CardItem).text() 
+    },
+    request: new Request("https://example.com/blog/blog-slug", {
+        headers: {}
+    }),
+    params: {
+        "slug": "blog-slug"
+    }, // in case your route is `pages/blog/[slug].astro`
     locals: {
         someFn() {
             
@@ -188,9 +165,9 @@ const response = await container.renderToString(Card, {
 });
 ```
 
-- `slots`: required in case your component is designed to render some slots inside of it.
+- `slots`: required in case your component is designed to render some slots inside of it. It supposed named slots too.
 - `request`: required in case your component/page access to some information such as `Astro.url` or `Astro.request`.
-- `params`
+- `params`: the `Astro.params` to provide to the components 
 - `locals`: initial value of the `Astro.locals`.
 - `status`: useful in case you're rendering an error page.
 
@@ -212,10 +189,8 @@ I have considered the idea of exposing the compiler itself, with a thin layer of
 # Adoption strategy
 
 Considering the fact that this API doesn't involve any configuration or virtual module, the API
-will be released with a `unstable_` prefix, e.g. `unstable_AstroContainer`. 
+will be released with a `experimental_` prefix, e.g. `experimental_AstroContainer`. 
 
 This implies that the public APIs of the class will be deemed unstable, and they can change anytime, in `patch` releases too. 
 
-Once the API is deemed stable, the `unstable_` prefix will be removed.
-
-# Unresolved Questions
+Once the API is deemed stable, the `experimental_` prefix will be removed.
--- a/proposals/0048-container-api.md
+++ b/proposals/0048-container-api.md
@@ -159,7 +159,6 @@ const response = await container.renderToString(Card, {
         someString: "string",
         someNumber: 100
     },
-    status: 400
 });
 ```
 
@@ -167,7 +166,7 @@ const response = await container.renderToString(Card, {
 - `request`: required in case your component/page access to some information such as `Astro.url` or `Astro.request`.
 - `params`: the `Astro.params` to provide to the components 
 - `locals`: initial value of the `Astro.locals`.
-- `status`: useful in case you're rendering an error page.
+- `routeType`: useful to test endpoints.
 
 # Testing Strategy
 
--- a/proposals/0048-container-api.md
+++ b/proposals/0048-container-api.md
@@ -4,7 +4,7 @@
 
 # Summary
 
-Astro components are tightly coupled to `astro` (the metaframework). This proposal introduces a possible server-side API for rendering `.astro` files in isolation, outside the full `astro` build pipeline.
+Astro components are tightly coupled to `Astro` (the metaframework). This proposal introduces a possible server-side API for rendering `.astro` files in isolation, outside the full `astro` build pipeline.
 
 # Example
 
--- a/proposals/0048-container-api.md
+++ b/proposals/0048-container-api.md
@@ -145,7 +145,7 @@ a method called `getContainerRenderer` that will return the correct information 
 ```js
 import { getContainerRenderer as reactRenderer } from "@astrojs/react";
 import { getContainerRenderer as vueRenderer } from "@astrojs/vue";
-import { laodRenderers } from "astro:container";
+import { loadRenderers } from "astro:container";
 import { AstroContainer } from "astro/container";
 
 const renderers = loadRenderers([
--- a/proposals/0048-container-api.md
+++ b/proposals/0048-container-api.md
@@ -1,0 +1,221 @@
+- Start Date: 2024-04-26
+- Reference Issues: https://github.com/withastro/roadmap/issues/533
+- Implementation PR: 
+
+# Summary
+
+Astro components are tightly coupled to `astro` (the metaframework). This proposal introduces a possible server-side API for rendering `.astro` files in isolation, outside the full `astro` build pipeline.
+
+# Example
+
+```js
+import { AstroContainer } from 'astro/container';
+import Component from './components/Component.astro';
+
+const container = await AstroContainer.create();
+console.log(await container.renderToString(Component, { props, slots }))
+```
+
+The container can be optionally constructed with some settings that are typically reserved for Astro configuration.
+```js
+const container = await AstroContainer.create({
+    mode: "production",
+    streaming: true,
+    astroConfig: {
+        site: "https://example.com",
+        trailingSlash: false
+    }
+})
+```
+
+The second argument to `renderToString` optionally provides render-specific data that may be exposed through the Astro global, like `props`, `slots`, `request`, and/or `params`.
+```js
+await astro.renderToString(Component, { props, slots, request, params })
+```
+
+# Background & Motivation
+
+Some of our own proposals, as well as many third-party tool integrations, are blocked until we expose a proper server-side rendering API for `.astro` components. Other frameworks have tools like `preact-render-to-string` or `react-dom/server` that make rendering components on the server straight-forward.
+
+We've avoided doing this because... it's very hard! Part of the appeal of `.astro` components is that they have immediate access to a number of powerful APIs, context about your site, and a rich set of framework renderers. In the full `astro` build process, we are able to wire up all this context invisibly. Wiring up the context manually as a third-party is prohibitively complex.
+
+
+# Goals
+
+- Provide a **low-level** API for rendering `.astro` components in isolation
+- Expose a familiar, user-friendly API
+- Surface enough control for full `astro` parity, but abstract away internal APIs
+- Enable third-party tools to consume `.astro` files on the server
+- Enable unit testing of `.astro` component output
+- Possibly unblock `.mdx` compiledContent/html output?
+- Support Astro framework renderers (`@astrojs/*`) if possible
+
+# Non-Goals
+
+- Provide a way to **import** `.astro` components. Users will be responsible to provide a compiled component.
+- Provide a way to **import** `astro.config.(mjs|mts)` out of the box.
+
+# Detailed Design
+
+## Preface
+
+**This** RFC will have a smaller scoped compared to what users have envisioned. The reason why the scope shrunk is that 
+I want to ship a smaller feature on the surface, which will allow us, later to enhance it based on user's feedback and use cases.
+
+That's why the first iteration of the APIs won't provide a built-in way to compile Astro components.
+
+## Internal concepts
+
+The API took a lot of time to land because it required some internal refactors. Thanks to these refactors, we will be able
+to land this API by using the **very same engine** that Astro uses under the hood.
+
+Eventually, we landed to a common and abstract concept called **Pipeline**. A pipeline is simply an abstract
+class that is responsible to render any kind of Astro route: page, endpoint, redirect, virtual.
+
+Each pipeline inside the Astro codebase (dev, build, SSR and now test) is responsible to collect the important information in different way,
+but eventually each route must be rendered using the same data, which are:
+- **Manifest**: this is a serializable/deserializable version of our "configuration", however it also contains more information, such as renderers, client directives, styles.
+- **Request**: a standard https://developer.mozilla.org/en-US/docs/Web/API/Request object.
+- **RouteData**: a type that identifies a generic Astro route. It contains information such as the type of route, the associated component, etc.
+- **ComponentInstance**: The instance of an Astro component. This is the **compiled** component. Internally, an Astro component undergoes various transformations via Vite, and it eventually gets consumed by our rendering engine (if it's a page).
+
+## Create a container
+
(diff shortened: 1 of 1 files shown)


proposals/0048-container-api.md
```diff
@@ -0,0 +1,221 @@
...
+})
+```
+
+The function `create` will throw an error if there's a validation error.
+
+> [!NOTE]
+> If you use a TypeScript file for your configuration, you are responsible for loading and transforming it to JavaScript.
+
+
+The `container` binding will expose a `renderToString` that accepts a Astro component and will return a `Response` object:
```
Princesseuh, line 106: Should it be called `renderToString` if it doesn't.. return a string? Maybe it'd be better to have a `renderToResponse` or something, and a `renderToString` wrapper that does the response.body dance for you.

ematipico: My bad. It returns a `string` :P

matthewp: +1, agreed, `renderToString` is a common need for testing, so makes sense to have a helper for this case.

proposals/0048-container-api.md
```diff
@@ -0,0 +1,221 @@
+- Start Date: 2024-04-26
+- Reference Issues: https://github.com/withastro/roadmap/issues/533
+- Implementation PR: 
+
+# Summary
+
+Astro components are tightly coupled to `astro` (the metaframework). This proposal introduces a possible server-side API for rendering `.astro` files in isolation, outside the full `astro` build pipeline.
```
matthewp, line 7:
```suggestion
Astro components are tightly coupled to `Astro` (the metaframework). This proposal introduces a possible server-side API for rendering `.astro` files in isolation, outside the full `astro` build pipeline.
```

proposals/0048-container-api.md
```diff
@@ -0,0 +1,221 @@
...
+const response = await container.renderToString(AstroComponent);
+```
+
+## Options
+
+It will be possible to tweak the container APIs with options and such. The `.create` function will accept the following interface:
+
+```ts
+type AstroContainerOptions = {
+    mode: "development" | "production";
```
matthewp, line 124: What do these two modes do? Anything that would cause the user of the API to change them?

ematipico: Yeah I don't think it's needed

proposals/0048-container-api.md
```diff
@@ -0,0 +1,196 @@
...
+
+
+# Goals
+
+- Provide a **low-level** API for rendering `.astro` components in isolation
+- Expose a familiar, user-friendly API
+- Surface enough control for full `astro` parity, but abstract away internal APIs
+- Enable third-party tools to consume `.astro` files on the server
+- Enable unit testing of `.astro` component output
+- Possibly unblock `.mdx` compiledContent/html output?
```
Princesseuh, line 50: In my opinion something cannot be a "possibly" a goal in a stage 3 RFC, same thing for the one below. What this possibly unlock can be discussed in other sections imo.

proposals/0048-container-api.md
```diff
@@ -0,0 +1,196 @@
...
+- Support Astro framework renderers (`@astrojs/*`) if possible
+
+# Non-Goals
+
+- Provide a way to **import** `.astro` components. Users will be responsible to provide a compiled component.
```
Princesseuh, line 55: Is this outdated?

proposals/0048-container-api.md
```diff
@@ -0,0 +1,196 @@
...
+- Provide a way to **import** `astro.config.(mjs|mts)` out of the box.
+
+# Detailed Design
+
+## Preface
+
+**This** RFC will have a smaller scoped compared to what users have envisioned. The reason why the scope shrunk is that 
+I want to ship a smaller feature on the surface, which will allow us, later to enhance it based on user's feedback and use cases.
```
Princesseuh, line 63: There's a mix of I / We in the RFC, even this sentence include both "**I** want" and "allow **us**". I personally believe it's best to always use `we` when we're writing RFCs from the perspective of the core team, but either works as long as it's consistent!

proposals/0048-container-api.md
```diff
@@ -0,0 +1,196 @@
...
+
+That's why the first iteration of the APIs won't provide a built-in way to compile Astro components.
+
+## Internal concepts
+
+The API took a lot of time to land because it required some internal refactors. Thanks to these refactors, we will be able
+to land this API by using the **very same engine** that Astro uses under the hood.
+
+Eventually, we landed to a common and abstract concept called **Pipeline**. A pipeline is simply an abstract
```
Princesseuh, line 72:
```suggestion
Eventually, we landed to a common and abstract concept called **Pipeline**. A pipeline is an abstract
```

If something is simple, you wouldn't be explaining what it is 😄

proposals/0048-container-api.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Public APIs
+
+- `AstroContainer::create`: creates a new instance of the container.
+- `AstroContainer::renderToString`: renders a component and return a string.
+- `AstroContainer::renderToResponse`: renders a component and returns the `Response` emitted by the rendering phase.
+
+### `create` function
+
+A container is class exposed via the `astro/container` specifier. Users **shouldn't** create a container using the `new` instance, but they should use he static function `create`:
```
Princesseuh, line 90:
```suggestion
A container is class exposed via the `astro/container` specifier. Users **shouldn't** create a container using the `new` instance, but they should use the static function `create`:
```

proposals/0048-container-api.md
```diff
@@ -0,0 +1,196 @@
+- Start Date: 2024-04-26
+- Reference Issues: https://github.com/withastro/roadmap/issues/533
+- Implementation PR: 
+
+# Summary
+
+Astro components are tightly coupled to `Astro` (the metaframework). This proposal introduces a possible server-side API for rendering `.astro` files in isolation, outside the full `astro` build pipeline.
```
matthewp, line 7:
```suggestion
Astro components are tightly coupled to `Astro` (the metaframework). This proposal introduces a server-side API for rendering `.astro` files in isolation, outside the full `astro` build pipeline.
```

proposals/0048-container-api.md
```diff
@@ -0,0 +1,196 @@
...
+import Component from './components/Component.astro';
+
+const container = await AstroContainer.create();
+console.log(await container.renderToString(Component, { props, slots }))
+```
+
+The container can be optionally constructed with some settings that are typically reserved for Astro configuration.
+```js
+const container = await AstroContainer.create({
+    mode: "production",
```
matthewp, line 22: Should this be removed? what does this do?

proposals/0048-container-api.md
ematipico: I removed the `mdx` thing. Not in scope for now

proposals/0048-container-api.md
ematipico: I updated it, by adding `non-vite environement`

proposals/0048-container-api.md
```diff
@@ -0,0 +1,194 @@
...
+
+```ts
+type AstroContainerOptions = {
+    streaming: boolean;
+    renderers: AstroRenderer[];
+    astroConfig: AstroUserConfig;
+}
+```
+
+The `astroConfig` object is literally the same object exposed by the `defineConfig`, inside the `astro.config.mjs` file. This very configuration
```
matthewp, line 134: Have you found use-case for passing the astroConfig yet? I'm worried about getting bug reports when people try to pass through `vite` config and other non-supported stuff.

What do you think about instead raising the relevant config values up to the `AstroContainerOptions` level. That would be stuff like `trailingSlash`.

proposals/0048-container-api.md
```diff
@@ -0,0 +1,196 @@
...
+While a low-level API requires more maintenance and work to provide more work, the main idea is to provide a bare-metal API so users can tweak it as they see fit. 
+
+# Alternatives
+
+I have considered the idea of exposing the compiler itself, with a thin layer of APIs, however users won't be able to compile the majority of formats like Markdown, MDX, etc.
+
+# Adoption strategy
+
+Considering the fact that this API doesn't involve any configuration or virtual module, the API
+will be released with a `experimental_` prefix, e.g. `experimental_AstroContainer`.
```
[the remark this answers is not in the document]
matthewp: Discussed this sync and we decided to keep the prefix.

proposals/0048-container-api.md
```diff
@@ -0,0 +1,194 @@
...
+    },
+    status: 400
+});
+```
+
+- `slots`: required in case your component is designed to render some slots inside of it. It supposed named slots too.
+- `request`: required in case your component/page access to some information such as `Astro.url` or `Astro.request`.
+- `params`: the `Astro.params` to provide to the components 
+- `locals`: initial value of the `Astro.locals`.
+- `status`: useful in case you're rendering an error page.
```
matthewp, line 170: can you explain this more? What is this option? Is it a string or what?

proposals/0048-container-api.md
ematipico: > What do you think about instead raising the relevant config values up to the AstroContainerOptions level. That would be stuff like trailingSlash.

I don't think that would work easily. These options, internally, are used to create a manifest. If we pass these options when rendering a component, it would mean generating a new manifest every time we attempt to render a component, and then discard that manifest. We have to be careful not to override the existing manifest, because the manifest is tight to the lifetime of the instance of the container.

proposals/0048-container-api.md
ematipico: That was a leftover. I removed it, and it seems it's not needed for now.

proposals/0048-container-api.md
matthewp: Not when rendering, `AstroContainerOptions` is the name of the options passed to `AstroContainer.create()`.

ematipico: Oh, I misunderstood what you meant. Yeah we should be able to provide the individual options.

proposals/0048-container-api.md
```diff
@@ -0,0 +1,193 @@
...
+        someNumber: 100
+    },
+});
+```
+
+- `slots`: required in case your component is designed to render some slots inside of it. It supposed named slots too.
+- `request`: required in case your component/page access to some information such as `Astro.url` or `Astro.request`.
+- `params`: the `Astro.params` to provide to the components 
+- `locals`: initial value of the `Astro.locals`.
+- `routeType`: useful to test endpoints.
```
abiriadev, line 169: Seems it lacks `props`. Can I know why they are not here?

ematipico: I update this document every time things change. When I released the container API, the props weren't there. 

The latest patch allows you to do that now. I will update the document soon

abiriadev: Great! :+1:

I'm looking forward to the update.

Princesseuh: It was released this morning!

tordans:
Hey! Can I use the new container API to render `const { Content } = await page.render()` into a string, pass this string to a React Component and render it there with `dangerouslySetInnerHTML`? I am looking for a way to render all posts in a blog post index kind of situation ([Example](https://github.com/withastro/docs/pull/8364/files#diff-a6fd5674a02c61d6110497ce6010118e4e9c6cfb50e3b5a9168faee52f77348dR484)) and pass them over to React without loosing all the markdown rendering that Astro does in `<Content/>`. It sounds like this API might help, but `Content` is not a component but returned by `await page.render()` so how does that work?

I tried passing `Content` but that fails:
```js
const rawPages = await getCollection('posts')
const sortedPages = rawPages.sort((a, b) => a.data.order - b.data.order)
const pages: RadnetzPage[] = []
for (const page of sortedPages) {
  const { Content } = await page.render()
  const container = await AstroContainer.create()
  const result = await container.renderToString(Content)
  pages.push({
    slug: page.slug,
    menu: page.data.menu,
    order: page.data.order,
    title: page.data.title,
    links: page.data.links,
    Content: Content,
    contentHtml: result,
  })
}
```
Results in…
```
Unhandled rejection
Astro detected an unhandled rejection. Here's the stack trace:
NoMatchingRenderer: Unable to render Content.

No valid renderer was found for this file extension.
    at renderFrameworkComponent (/…/node_modules/astro/dist/runtime/server/render/component.js:190:15)
    at async Module.renderComponent (…/node_modules/astro/dist/runtime/server/render/component.js:396:10)
```

(I was directed here from [the changelog](https://github.com/withastro/astro/blob/main/packages/astro/CHANGELOG.md#minor-changes).)

ematipico:
@tordans yes you can, however you'll have for this PR to be merged and released: https://github.com/withastro/astro/pull/11141

tordans:
> @tordans yes you can, however you'll have to wait for this PR to be merged and released: [withastro/astro#11141](https://github.com/withastro/astro/pull/11141)

Perfect, thanks @ematipico. 
I noticed that https://github.com/withastro/astro/pull/11141 does not mention any Docs changes (yet) and  https://github.com/withastro/docs/pulls?q=is%3Apr+is%3Aopen+container-reference doesn't either (yet). Given that I noticed a lack of docs on the topic of rendering lists – see https://github.com/withastro/docs/pull/8364 – that might be something to show in the docs for this API as well.

universse:
Can I use this in a Cloudflare project? It currently throws an error during build.

```
[commonjs--resolver] [plugin vite:resolve] Cannot bundle Node.js built-in "node:fs/promises" imported from "..\..\node_modules\.pnpm\@astrojs+mdx@3.1.3_astro@4.12.2_@types+node@20.14.12_typescript@5.5.4_\node_modules\@astrojs\mdx\dist\index.js". Consider disabling ssr.noExternal or remove the built-in dependency.
```

I am using the MDX renderer specifically to render `<Content />` from content collection.

universse:
```js
const { Content } = await entry.render()
const descriptionHtml = await astroContainer.renderToString(Content)
```

I'm using Astro container to render `<Content />` to string and pass it to React component. I'm using `@astrojs/cloudflare` integration and run into the above build error. When I swap to `@astrojs/node`, the build successfully completes. So I think there's some bundling issue for Cloudflare worker.

ematipico:
@alidcast  If you run the container inside a Vite environment, then yes, all your `vite` settings should be taken into consideration during the rendering phase of the container.

dwighthouse:
Hi. I've been using the `experimental_AstroContainer` feature as it exists today in Astro 4.15.11. I've noticed only one potential issue so far:

While script tags and scoped style tags appear to work correctly, the use of imported css modules will not be rendered via renderToString(), regardless if the existence of a head tag.

This could be sometimes useful. For example when rendering html for injection into the content:encoded tag of RSS or copying strings into JS variables that only needs the rendered HTML.

However, if rendering for the purposes of being placing on a page, the loss of styles is a showstopper. I didn't see any setting that controlled whether or not to inject module style tags in the container docs.

ematipico:
@dwighthouse 

This hasn't been implemented because we failed to find a valid use case. You just mentioned RSS, JS variables and such. Since we are evaluating a real use case, it would be amazing if you could go in detail and explain to me what you're looking for, and if you'd explain me very well you use case.

dwighthouse:
@ematipico
It would be my pleasure.

## Scenario 1: Creating a page at `/404/` or `/500/`

Because AstroJS has made [a page rendering exception](https://docs.astro.build/en/basics/astro-pages/#custom-404-error-page) for "404" and "500" named pages, it is currently extremely difficult to statically render a page to "/404/index.html". As mentioned in a Discord thread, I have attempted all of the following:

* `pages/404.astro` - outputs "404.html"
* `pages/404.mdx` - outputs "404.html"
* `pages/404/index.astro` - outputs "404.html"
* `pages/[404].astro` and set getStaticPaths to output "404" - outputs "/404.html".
* `pages/[...404].astro` and set getStaticPaths to output "404" - outputs "/404.html".
* `pages/[...404].astro` and set getStaticPaths to output "404/index.html" - outputs "/404/index.html/index.html".
* `content/pages/404.mdx` and then use a collection renderer for "pages" (via [...pages].astro) which renders most of the rest of my pages, getStaticPaths set to "404" - outputs "/404.html"

So as you can see, AstroJS seems to have an algorithm that forces all output of the page rendering system that would normally go to `/404/index.html` or `/500/index.html` to a different location.

> Aside: In my opinion, this should be a setting you can turn off. In my case, I want `/404/` to be my canonical 404 error page location. But there are other legitimate scenarios where you would want this feature. For example, if I made a website with the address "https://pokedex.net", and I wanted to list Pokemon by their number. In AstroJS, https://pokedex.net/404/ would never work as expected. Instead of taking you to the page for [Luxio](https://db.pokemongohub.net/pokemon/404), it would redirect you to https://pokedex.net/404.html, the error page. Depending on how you implemented the site, the page would either contain error information or Luxio information. Either way, it doesn't work as expected.

After much experimentation, I was able to find that it IS possible to output to the location of `/404/index.html`, but you have to go around the world to do it. Here's how:

1. Create a file at `pages/[...404].js`.
2. In the file, set the getStaticPaths function to export data containing:
    ```
    {
        params: {
            404: '404/index.html'
        }
    }
    ```
3. Use `experimental_AstroContainer` to render a different `.astro` file that knows how to get the correct page markdown (or whatever) directly to a string. Use the new option (as of [4.16.6](https://github.com/withastro/astro/releases/tag/astro%404.16.6)) of `{ partial: false }` so we get the doctype in the output.
4. Output that string as the file's content.

If you do all this, the output will be a 404 error page at the path `/404/index.html`. However, while inline scripts and scoped css stylesheet tags will be included, imported css modules do not get inserted into the `<head>` tag as they would in a normal page rendering scenario.

Since I use css modules, my 404 error page has no styles, even with all these workarounds for the fact I can't treat the word "404" like any other page name.

I have suggestions for how this could be handled in scenario 3 below.

## Scenario 2: RSS Feed Encoded Content

As described in the docs for [RSS Feeds](https://docs.astro.build/en/guides/rss/#including-full-post-content), you can render full post content to the feed. You can use `(new MarkdownIt()).render()` or `compiledContent()` to render content to the the RSS. However, these are hugely limited.

* `(new MarkdownIt()).render()` - Can't render components or JSX expressions in MDX files.
* `compiledContent()` - Can't render MDX files at all.

This makes these solutions non-starters for me. All my content is in MDX format, and I make extensive use of components. Components are used for every image, every embedded video, and numerous other things I have planned. In some cases, the content of components can represent more than 50% of the content of a post.

Because of this and some minor validation related things with how the RSS plugin renders feeds, I elected to roll my own by exporting my own constructed string as my feed.

Using `experimental_AstroContainer`, you can export just the partial of the page content, which will include the rendered components down to raw HTML, exactly as you'd want. Generally, RSS feeds don't include styles as they are ignored, by many RSS Feed Readers. So the failure to include imported styles is no problem here. This just represents a good use for `experimental_AstroContainer` to get the actual rendered MDX content into raw HTML.

## Scenario 3: Read More Post Splitting

Various CMS systems, with the most common example being [in Wordpress](https://wordpress.org/documentation/article/more-block/), feature a way of specifying that a post should be split at a certain point. Under preview conditions like a search page or a blog homepage, the content at the top of the post will be shown as full HTML rendered data. After the preview section, there will be a "Read More..." or "Continue Reading..." link that refers to the full post's location. When going to the full post, the entire content of the post is shown, with nothing indicating the split, except perhaps an invisible anchor tag so the links can jump to the continuation point in the page.

In Wordpress, you activate this mode by inserting an HTML comment or some block to indicate the intention to break the page there. It's similar on other systems.

Such a system appears to be currently impossible, as described, in AstroJS. I was able to implement the equivalent feature by creating a component that looks like this:
```
Preview content.

<ReadMore readMore={props.readMore}>
    After the jump content.
</ReadMore>
```

As you can see, I have to play a few special tricks to make sure appropriate props get generated on the way down the rendering path in order to make this work. This works OK, but could be better. Issues include:

* I have to do weird stuff with the props.
* I have to store after-the-jump content inside a component, when really I just want to split the content at a point.
* On pages that only show the preview content, there may be components (and therefore imported styles and scripts) that only apply to after-the-jump content that will be included in the output even though they are never used.

What would solve these things, which may be beyond the scope of `experimental_AstroContainer`, would be the ability to generate an arbitrary string of Astro content (or MDX content) where we could pass as options the values that would normally be determined by an actual file's location and other parts of the import/rendering process. With the raw string of imports and MDX/Astro content, `experimental_AstroContainer` would import the necessary components and render them to a raw HTML string.

Using such a system, splitting the post content could be as simple as `contentText.split('<!-- Read More')[0]`, followed by an optional optimization of manually removing any component or style imports we knew we wouldn't need just for the preview. Thus, only the preview content and its dependencies would be output on the final page; as optimal as one could hope for. (In the current rendering system, if a component or CSS is imported but never used, any of its CSS and JS dependencies will be imported anyway, which I would like to avoid in these cases were we are arbitrarily splitting the content.)

However, to use anything like this, `experimental_AstroContainer` would need to operate as a real non-partial. If components or the content itself imported a CSS module, that module would need to be inserted into the normal rendering path for this page.

The main page renderer doesn't know that we're using another renderer to generate HTML, so how could it know to insert some CSS or JS? This implies to me that to truly provide this level of control, there would need to be some way of interfacing with the existing rendering system. For example, the `experimental_AstroContainer` could output not just a raw HTML string, but also the CSS and JS imports it detected, and then pipe that data into the existing page renderer to wind up on the final page, if a given scenario calls for it. Doing so would yield additional control in the form of filtering CSS and JS imports mid-render.

## Scenario 4: Post HTML Excerpt Feature

Once again inspired by Wordpress, there is a feature to basically split the HTML content so you only get some subset at the top of the post. The content is rendered to HTML, and after filtering out certain problematic tags like tables, some amount of valid HTML is returned that constitutes the first part of the post, probably limited by some number of characters if the HTML content had its `innerText` property checked. I believe it is done by the [the_excerpt function](https://developer.wordpress.org/reference/functions/the_excerpt/).

Such things are relatively easy to do with a virtual DOM library like [Cheerio](https://github.com/cheeriojs/cheerio), but only if you can get the data as raw HTML strings, hence `experimental_AstroContainer`. Similar to Scenario 3 above, this feature would benefit from being able to filter the content and its imports prior to rendering, as well as the ability to tell the page renderer to take on extra imported CSS and JS content when outputting its own raw HTML to files.

## Scenario 5: Post Text Excerpt Feature

Almost identical to Scenario 4 above, this feature would do the same thing, except the final output is retrieved via the `innerText` property, resulting in a pure-text representation of the excerpt of the post after the components have been rendered down to raw HTML. This can be useful for auto-generating meta descriptions, search results, social media excepts, and the like.

---

I hope this gives you an idea where I'm coming from. The ability to arbitrarily render content within a content rendering system is indeed powerful, but unless these other cases are addressed, I fear its inability to retain styles and split content logically will seriously limit its application to just its use when creating HTML to be connected to client-side JS libraries.

Thanks for coming to my TED Talk. ;)

taoeffect:
Yes @dwighthouse's post is excellent. You can see how we are forced to use `Astro.glob` instead of `getCollection` because of this in this repo: https://github.com/okTurtles/wordpress-to-astro/issues/3

JonathonRP:
I wanted to add my use case for accessing styles, I have a portfolio site with resume/CV component I've made and wish to display it on page differently than download versions only need component with styles while page has layout and supplies styles to show more pretty.

jarrp001:
I am working on porting my company's SDUI framework from mobile only to web. Astro looks like a great fit, as our SDUI framework is component based as well. SDUI controls the component tree (and properties). Each component can be mapped to an `.astro` file. Then use the `renderString` to build the component, and finally combine all of them into the final tree for the page. The page maybe delivered as SSR or cached on content changes and pushed to a CDN for delivery.

Unfortunately, without style support, it means we cannot style the components as they need to be. Each component, in isolation. The same is the case for scripts, as our SDUI framework has something called "handlers". These would be scoped to the component, in most cases. And so we would need to provide the scoped JS with the component when rendering.

I am able to use the Container API to build the components from SDUI into Astro. But, without styles and scripts, the page isn't usable :-(. I would love to hear about the options to include styles and scripts.

proposals/0048-container-api.md
```diff
@@ -0,0 +1,231 @@
...
+
+Adding a renderer can be done manually or automatically. 
+
+The automatic way is meant for static applications, or cases where the container isn't called at runtime. The renderers maintained by the Astro org will expose
+a method called `getContainerRenderer` that will return the correct information that will tell the container **how to import the renderer**. The importing of the renderer is done via a function called `loadRenderers`, exported by a new virtual module called `astro:container`: 
+
+```js
+import { getContainerRenderer as reactRenderer } from "@astrojs/react";
+import { getContainerRenderer as vueRenderer } from "@astrojs/vue";
+import { laodRenderers } from "astro:container";
```
Scot3004, line 148: I think this import should be `{ loadRenderers }` instead of `{ laodRenderers }`

zadeviggers:
It would be great if containers could be used in browser environments to enable things like live previews in CMSes.

proposals/0048-container-api.md
```diff
@@ -0,0 +1,231 @@
...
+import { AstroContainer } from "astro/container";
```
coder0107git, line 149: It's confusing having to import from both `astro:container` and `astro/container`. It might be nicer to re-export `astro/container` in the virtual `astro:container` module.

proposals/0048-container-api.md
ematipico: Add an inline suggestion! :) I will apply it straight away

Scot3004: https://github.com/withastro/roadmap/pull/1109

JonathonRP:
```suggestion
import { loadRenderers } from "astro/container"
```

reesericci:
Running into this as well because I'm trying to render Markdown from my articles and grab the content to feed into an open graph description with cheerio.

```astro
---
import { getCollection } from "astro:content";
import Layout from "../../layouts/Layout.astro";
import og from "../../components/og";
import * as cheerio from 'cheerio';

  const writings = await getCollection("writings");
  return writings.map((entry) => ({
    params: { slug: entry.data.date.toString() },
    props: { entry },
  }));
}

const { entry } = Astro.props;
const { Content, remarkPluginFrontmatter } = await entry.render();

[...]

const content = cheerio.load(Content())
const description = content('p').toArray().map((el) => content(el).text()).join("\n")
---
<Layout title={entry.data.title}>
  <meta property="og:description" content={description}>
[...]
</Layout>
```

Naryalin:
I want to share another possible use case:

I'm currently working with email sending directly inside Astro API Endpoints, and I find it very useful to use Astro components as HTML templates. The `experimental_AstroContainer` with the `renderToString` method works pretty well for this.

However, I noticed that `<style>` tags defined inside `.astro` files are not included in the output. Since emails require internal or inline styles (external CSS isn't supported in most clients), this becomes a blocker for using Astro components as standalone email templates.

It would be great if either:

* `<style>` blocks were automatically included in the rendered HTML, or
* there was an option to explicitly include styles when calling `renderToString()`.

This would make Astro a very strong option for email templating use cases and take advantage of predefined variables and styles in layouts to maintain brand consistency across all email communications, without needing to duplicate markup or styling logic in separate tools.

Thanks!

mnchaser:
Thank you all for the work on this!  

In addition to the comment above... how would (could) a plugin like Tailwind be handled by the renderToString()?  
I believe right now it does nothing with those classes because I still see them in the html? 

_Although, "classes" are getting stripped completely from my  elements...but I suspect that is from sanitizeHtml._

Like many others, I am trying to add full post content to an RSS feed.  I feel like it's 90% there (at least for my need). The last 10% is around styling.

In the meantime, I'm exploring if it's possible to take my rendered HTML and generate css for Tailwind programmatically in javascript and merge that for my RSS.

For reference...
`      
let postHtml = await container.renderToString(Content);

//manual step removing elements that use client-side script that won't render correctly in RSS or email.
postHtml = cleanPostsForRSS(postHtml); 

const sanitizedHtml = sanitizeHtml(postHtml, {
     allowedTags: sanitizeHtml.defaults.allowedTags.concat(["img"]),
});

//TODO: generate CSS for Tailwind
`

louiss0:
What is going on? I don't like using experimental features! This one is important to me!

kocv59:
Why does container API renderToString have to depend on an Astro project? `astro(astroConfigFile).renderToString("Template.astro", data)` should produce a string or stream so it can be used in any NodeJS file.

rktyt:
Currently, `renderToString` depends on `renderToResponse`.
Because of this, even if something is valid as a component, it cannot be passed to `renderToString`, which makes component unit testing difficult.

For example, a component that returns `null` in its frontmatter:

```astro
---
return null
---
```

This is currently valid as a component (not a page), but it cannot be handled by `renderToString`.

If I change it to:

```astro
---
return new Response(null)
---
```

then it can be handled by `renderToString`, but in that case it can no longer be used as a component within a page.

This is a significant issue for me, so I would really appreciate it if `renderToString` could support non-page components.

If this has already been discussed, please feel free to disregard.

ematipico:
@rktyt are you testing/rendering a page instead of a component?

rktyt:
> @rktyt are you testing/rendering a page instead of a component?

I'm currently testing both.
For that reason, I avoid returning at the front matter root in the component.
However, due to this issue, one drawback is that the amount of code in the component ends up increasing.

ematipico:
Thank you for the feedback. It's indeed an annoying quirk but at least it gives me good insights on how to properly shape the APIs once it's stable

jmrona:
## Headless fragment rendering outside Vite — real-world usage report

We have been using `experimental_AstroContainer` to render individual island
components to standalone HTML fragments in a Node.js environment where Vite
is absent at runtime. The use case is fragment-only headless SSR under
`@astrojs/node` in standalone mode, with no full-page Astro rendering.

The API works for this use case. We are reporting the quirks we hit
as concrete feedback toward stabilization.

**Astro `6.4.5`, `@astrojs/node` `10.1.4`, `@astrojs/preact` `5.1.5`, Node `v26.0.0`**

---

### Quirk 1 — Built component chunk exports `$` not `default`

After `astro build`, the server chunk exports the component as the minified
alias `$`, not `default`:

```js
export { $$IslandOne as $ };
```

We work around this with `mod.default ?? mod.$` but this is an internal
Vite/Rollup detail with no stability guarantee.

**Question:** Is there a stable, documented way to import a pre-built
component for the Container API outside Vite?

---

### Quirk 2 — No stable chunk path — build hash changes every build

The built component chunk is emitted as `IslandOne_dYIra1Kf.mjs`.
The hash changes on every build and `dist/client/.vite/manifest.json`
is not emitted for our configuration, so we resort to a glob scan of
`dist/server/chunks/` at runtime.

**Question:** Is there a supported way to get a stable build-time map of
component name → chunk path for Container API use at runtime?

---

### Quirk 3 — `@astrojs/node` standalone autostarts on chunk import

Importing the built component chunk triggers the standalone Node server
to autostart. Workaround:

```js
process.env.ASTRO_NODE_AUTOSTART = 'disabled';
```

This env var is not documented as a supported Container API escape hatch.

**Question:** Is `ASTRO_NODE_AUTOSTART=disabled` stable? If not, is there
a recommended alternative?

---

### Quirk 4 — Dummy build entrypoint required for separate chunk emission

Without an explicit page-level import, Astro/Vite inlines the component
into the page chunk as a non-importable closure-local. A dummy endpoint
is required to force its own chunk:

```ts
// src/pages/island-one-component.ts
export { default } from '../components/IslandOne.astro';
```

**Question:** Is there a build config option to force individual chunk
emission without dummy entrypoints?

---

### Quirk 5 — `@astrojs/preact/server.js` requires a `registerHooks` shim

Using Preact components inside `.astro` wrappers via the Container API
outside Vite requires a `node:module` `registerHooks` shim for the virtual
`astro:preact:opts` module. Not documented for no-Vite runtimes.

**Question:** Is there a supported path for `@astrojs/preact` + Container
API without this shim?

---

### What worked well

- `renderToString` produces valid fragments with correct `astro-island`
  markup, `component-url`, and `renderer-url` attributes
- Preact `client:load` inside `.astro` wrappers correctly emits client
  bundles with automatic deduplication via Vite code-splitting
- The fragment-rendering contract is solid for production use

We are happy to answer questions about this use case if it helps
the stabilization effort.

wiesson:
A use case and a question, in case it helps with prioritising.

We keep a catalog of about 80 page sections as `.astro` components and render client sites from data: a section id plus a props object per section, stored in a backend. An external editor (a separate app, not Astro) needs live previews of single sections while the props change.

Today we do this with an iframe pointing at a small SSR endpoint in the catalog project (`/render/<id>?props=…`, Astro 7.2, Cloudflare adapter, so it runs in workerd). That works well and gives us the complete output, including scoped styles and the sections' own scripts.

We looked at the Container API as the "proper" way to do this, and held back for the reasons already listed in this thread: styles and scripts are not part of the `renderToString` output, and using it outside Vite or inside workerd is undocumented.

Two questions:

1. Is "render one component with props, including its styles and scripts, at runtime in workerd" something the Container API is meant to cover once it is stable? Or is an SSR endpoint plus iframe the intended pattern for previews in external tools?
2. Is there a rough idea of what is still missing before this RFC can move forward? We would be happy to test pre-releases against a real catalog.

Thanks for the work on this.

ematipico:
> Two questions:
> 
>     1. Is "render one component with props, including its styles and scripts, at runtime in workerd" something the Container API is meant to cover once it is stable? Or is an SSR endpoint plus iframe the intended pattern for previews in external tools?
> 

I am looking into it, and I believe it's doable, with some limitations e.g. the component renders only it's own styles i.e. `<style>`, not the styles imported.

>     2. Is there a rough idea of what is still missing before this RFC can move forward? We would be happy to test pre-releases against a real catalog.

Not yet, but I might be able to dedicate some time in Q4 and stabilise with proper use cases

tleunen:
> styles and scripts are not part of the renderToString output

Just to add a bit on this, they are part of the output, but the urls go nowhere instead of being linked to the assets directory. I believe they are simply "unprocessed" because the container has no knowledge about the production build process.

In any case @ematipico, let us know when you get something working. I'm also very interested in the container api to allow rendering a full astro component to string. 
Not all scripts can always be loaded upfront, and so being able to have scripts within astro components is a must have for me, especially as Astro documents using [web components to enhance static components](https://docs.astro.build/en/guides/client-side-scripts/#web-components-with-custom-elements). 

In some cases, [page partials](https://docs.astro.build/en/basics/astro-pages/#page-partials) can be used as a work around because they do run within the context of the full astro build, and even if the docs says not to use them with scripts, the scripts work fine.

ematipico:
Hey folks, I just shipped a function that allows rendering a component with script and styles, with some known limitations: https://docs.astro.build/en/reference/container-reference/#rendercomponent

wiesson:
Thank you, that was quick! We'll try `renderComponent` against our catalog (~90 section components, rendered at runtime in workerd via the Cloudflare adapter) and report back here with concrete results: which components render fully, which need our global CSS loaded separately, and how scripts with imports behave.

tleunen:
@ematipico I'm confused as to why you had to make a different function to render this instead of fixing the existing `renderToString`? Because the existing `renderToString` still output a broken html. The html still contains a script tag with a broken script url.

In fact, I now have tried `renderToComponent` and the output is exactly the same as `renderToString` for a very basic test case.

Maybe I'm missing something, but the test case is the following. Very basic component with a script:
```
<div id="astro-container-probe">
  Processed script: <output id="astro-container-probe-status">not executed</output>
</div>

<script>
  const status = document.getElementById('astro-container-probe-status')
  if (status) status.textContent = 'executed'
</script>
```

And it gets the following output both in the dev server, and then built files:
```
<div id="astro-container-probe">
    Processed script: <output id="astro-container-probe-status">not executed</output>
</div>
<script type="module" src="/Users/[..........]/component.astro?astro&type=script&index=0&lang.ts"></script>
```

ematipico:
@tleunen  That's not a test case, but only the output from something I can't debug. Did you use the `?container` query string?

I decided to have a different function for two reason:
- the new function doesn't accept `routeType`, which is an information usually reserved to routes, not components. `renderToString` accepts that option
- we're still evaluating if it's best to have multiple functions, or a function that accepts what to render. i.e. rendering a route VS rendering a component. Rendering in `vite` VS rendering outside `vite`

tleunen:
Yup, the component was rendered using the new function and the import url. I thought that was obvious but sorry for that confusion 😅 Thanks for the explanation though!

```
import Component from './component.astro?container'

export const GET: APIRoute = async () => {
  const container = await AstroContainer.create()
  return new Response(await container.renderComponent(Component), {
    headers: { 'Content-Type': 'text/html; charset=utf-8' },
  })
}
```

I'm still debugging with my agent as well at the same time, hoping I can help in any way

Issue #945, opened 2024-06-07
Server islands


matthewp:
- Accepted Date: 2024-06-07
- Reference Issues/Discussions:
  - https://github.com/withastro/roadmap/discussions/907
  - https://github.com/withastro/roadmap/discussions/116
- Author: @Tc-001 @ascorbic
- Champion(s): @matthewp 
- Implementation PR: 

# Summary

Allow islands within a prerendered page to be server-rendered at runtime.

# Background & Motivation

Often a page may be mostly static, but with only parts that need to be rendered on-demand. For example, a product page might need to show stock levels or personalised recommendations. Currently the only option for these is to either render the whole page on demand, or to render the dynamic parts on the client. This proposal introduces the concept of deferred islands, which are not prerendered, but rather server rendered on demand at runtime.

Next.js is working on a solution called [partial pre-rendering](https://nextjs.org/docs/app/api-reference/next-config-js/partial-prerendering), which allows most of the page to be prerendered, with individual postponed parts rendered using SSR on-demand. The implementation is quite different from what I propose for Astro, but the concept is similar.

# Goals

## Goals

- Allow the shell to be fully rendered as valid, cacheable, static HTML, with optional placeholders for postponed islands.
- Replace the deferred islands at runtime, using server-rendered HTML
- Require no JavaScript except a very small script to load the data
- Load the deferred islands individually and asynchronously, allowing each to be rendered as soon as it is available
- Allow access to on-demand rendering features in deferred components, such as cookies and the response object
- Allow components to know if they are currently being rendered in deferred mode

## Possible goals

- Allow deferred components to be cacheable (see [GET vs POST](#GET-vs-POST))

## Non-goals

- Automatic pre-rendering of deferred components. Only fallback content would be pre-rendered
- Batching of requests for deferred content. At this stage there would be one request per component, with batching as a possible later optimization
- Allowing components to default to deferred rendering. Deferring would be specified at the page level.

## Example

A component would be deferred by setting the `server:defer` directive. 

```astro
<Like server:defer />
```
The `"fallback"` slot can be used to specify a placeholder that is pre-rendered and displayed while the component is loading.

```html
<Avatar server:defer>
    <div slot="fallback">Guest</div>
</Avatar>
```

The page can pass props to the component like normal, and these are available when rendering the component:

```astro
---
import Like from "../components/Like";

export const prerender = true;

const post = await getPost(Astro.params.slug)
---
<Like server:defer post={post.id} />
```

[...]

PR #963, opened 2024-06-25
Server Islands

matthewp:
# Summary

Allow islands of server-rendered content that renders after the page load, allowing more cacheable pages.

```astro
<Avatar server:defer>
  <div slot="fallback">Guest</div>
</Avatar>
```

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/server-islands/proposals/server-islands.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/945)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/907)

### DIFF
--- a/proposals/server-islands.md
+++ b/proposals/server-islands.md
@@ -1,0 +1,147 @@
+- Start Date: 2024-06-25
+- Reference Issues: https://github.com/withastro/roadmap/issues/945
+- Implementation PR: https://github.com/withastro/astro/pull/11305
+
+# Summary
+
+Allow islands of server-rendered content that renders after the page load, allowing more cacheable pages.
+
+# Example
+
+A component, Astro or framework, can be deferred using the `server:defer` directive:
+
+```astro
+<Avatar server:defer>
+  <div slot="fallback">Guest</div>
+</Avatar>
+```
+
+The page can also pass props like normal. These props are included in the request to fetch the server island:
+
+```astro
+---
+import Like from "../components/Like";
+
+export const prerender = true;
+
+const post = await getPost(Astro.params.slug)
+---
+<Like server:defer post={post.id} />
+```
+
+# Background & Motivation
+
+Personalized and dynamic content reduce the ability to cache pages. Forgoing that content in the initial page request allows more effective caching strategies. This allows a CDN to deliver an initial page, either from static content or server-rendered and cached content, closer to the user immediately in most cases. Personal and dynamic content can still be delivered after the initial HTML request.
+
+## Definitions
+
+- **Personalized content**: Content on a web page that is distinct for a user, usually one who is logged in. Examples include a logged in user's avatar and menu items.
+- **Dynamic content**: Content delivered on a page that changes frequently. An example would be a carousel of *related products* on an ecommerce site.
+
+# Goals
+
+- Allow deferred content to be rendered asynchronously with the page request.
+- Explicit opt-in to server islands; no magic discovery based on a heuristic.
+- Host agnostic and simplicity are preferred. Ideally when prerendering the static pages can be deployed to any host.
+- Individual server islands per usage, no global fetch of all islands, to allow parallel and async loading.
+- Allow access to on-demand rendering features in deferred components, such as cookies and the response object.
+
+# Non-Goals
+
+- Prerendering of deferred components, only specified fallback content will be rendered.
+- Static content inside of a server island; like with client islands once you are inside of a server island all components rendered within are also server rendered.
+- Zero JS. For portability and simplicity, using a small client script to fetch the island is a better approach.
+
+# Detailed Design
+
+Server islands are declared with the `server:defer` directive. The compiler will:
+
+- Scan components looking for this directive.
+- When it finds one, traces the component to its associated import statement.
+- Creates metadata, like with client islands, returned from compilation that gives a list of each island used in each component.
+
+In Astro a route is created, `/_server-islands/[name]` that serves discovered islands. During the "server" phase of the build the islands are discovered and collected into a map.
+
+## Naming algorithm
+
+After the server build there is a secondary build for the discovered islands. Each island is given a distinct name using this algorithm:
+
+- A component is by default named its usage. If `src/components/Avatar.astro` is imported as `Avatar` and used as a server island it is by default named `Avatar`.
+- If the same component is used somewhere else, but renamed to another name, the first discovered usage serves as the name.
+- If another component has already claimed the name `Avatar` then the name is appended a number, `Avatar1`. The name is recursively checked with the number incremented until it finds a distinct name.
+
+## Rendering
+
+### Island rendering
+
+Server islands are rendered with the same rules as a `partial`; no `<doctype>` is appended to them, nor are scripts and styles included in the response. Since the islands are used within pages their scripts and styles are already collected, bundled, and injected as part of the page's own build process.
+
+When a request for `/_server-islands/Avatar` comes through the runtime:
+
+- Looks in the `serverComponents` field in the `SSRManifest`. This field is a `Map<string, () => Promise<ComponentInstance>>` where the key is the component's distinct name and the value is a function that will return a promise for the component. This is similar to the data structure used to lazy load pages.
+- The server island calls the value of this map to retrieve the `ComponentInstance` which is then rendered inside of the endpoint.
+
+### Page rendering
+
+When the page renders, either at build time (`output: 'hybrid'`) or at runtime (`output: 'server'`), components with the `server:defer` directive are not rendered. Instead a script is injected (explained in next section).
+
+Additionally the `slot="fallback"` is rendered and returned before the hydration script. The hydration script is injected along with stringified:
+
+- Component `name` as described in the naming algorithm.
+- `props` passed to the component. An island can be rendered multiple times; the props are representative of a particular usage.
+- `slots` that are passed to the island component.
+
+## Hydration
+
+The hydration script performs the following steps:
+
+- Creates an HTTP request to `/_server-islands/[name]`
+- Consumes the body of the request into a string.
+- Turns the string of HTML into a document fragment.
+- Removes the fallback content, if there is any.
+- Injects the new fragment.
+- Removes the script.
+
+# Testing Strategy
+
+This feature spans multiple parts of Astro so it will be tested in layers:
+
+## Compiler
+
+- The compiler piece of this mostly deals with the metadata that is returned. So integration/wasm tests will be added to verify the right output.
+
+## Rendering
+
+- Fixture tests for server generated content, dev and build, to ensure the script is emitted for islands.
+
+## E2E
+
+- Playwright tests to verify the island hydrates, requests the server contents, and renders it properly on the client. 
+
+# Drawbacks
+
+- There is overlap between this feature and client directives, particularly `client:only` which can include fallback content. It is hard to explain why you would use server islands over this feature. One reason is that client directives that fetch from an API cause a waterfall that is not included with server directives which only have a small inlined script.
+
+# Alternatives
+
+The major alternative implementation idea is to not fetch the islands via a script but rather to do so inside of an Edge function and then stitch together the HTML as the islands stream in. Such an approach would have these advantages:
+
+- Prevention of a waterfall caused by the island only being fetched once the page loads.
+- No fallback content needed. Fallback requires design considerations, using an Edge function would be more akin to SSR.
+
+However this approach has some downsides:
+
+- Eliminates the caching advantage gained by the script approach. Since the edge function injects personalized content the page cannot be cached globally.
+- Only works will with Edge functions, so limited choice of hosts. Loss of portability.
+- The page's main content will often be delayed from being visible to the user as it is blocked by server islands being fetched further up in the page.
+
+# Adoption strategy
+
+- Experimental release while the stage 3 RFC goes through revisions.
+- This is an opt-in feature that does not include any breaking changes to existing features. Only users who want to use it will.
+- This feature requires compiler integration so there are no similar features in the Astro ecosystem.
+
+# Unresolved Questions
+
+- During stage 2 there was some discussion about `props` which get serialized to the island. It could make sense to encrypt them to prevent mistakening leaking secrets. How/if this can be done hasn't been determined yet.
+
--- a/proposals/server-islands.md
+++ b/proposals/server-islands.md
@@ -91,6 +91,14 @@ Additionally the `slot="fallback"` is rendered and returned before the hydration
 - `props` passed to the component. An island can be rendered multiple times; the props are representative of a particular usage.
 - `slots` that are passed to the island component.
 
+### Props serialization
+
+Since the island is replaced with a script and fallback content at build time, the props must be serializable. This is done using `JSON.stringify`.
+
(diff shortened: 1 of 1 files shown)


proposals/server-islands.md
```diff
@@ -0,0 +1,147 @@
...
+
+The major alternative implementation idea is to not fetch the islands via a script but rather to do so inside of an Edge function and then stitch together the HTML as the islands stream in. Such an approach would have these advantages:
+
+- Prevention of a waterfall caused by the island only being fetched once the page loads.
+- No fallback content needed. Fallback requires design considerations, using an Edge function would be more akin to SSR.
+
+However this approach has some downsides:
+
+- Eliminates the caching advantage gained by the script approach. Since the edge function injects personalized content the page cannot be cached globally.
+- Only works will with Edge functions, so limited choice of hosts. Loss of portability.
```
kandros, line 135: > Only works will with Edge functions,
typo

doeixd:
Can server islands be nested?

proposals/server-islands.md
```diff
@@ -0,0 +1,147 @@
...
+
+# Adoption strategy
+
+- Experimental release while the stage 3 RFC goes through revisions.
+- This is an opt-in feature that does not include any breaking changes to existing features. Only users who want to use it will.
+- This feature requires compiler integration so there are no similar features in the Astro ecosystem.
+
+# Unresolved Questions
+
+- During stage 2 there was some discussion about `props` which get serialized to the island. It could make sense to encrypt them to prevent mistakening leaking secrets. How/if this can be done hasn't been determined yet.
```
ascorbic, line 146: An alternative is signing the props, which would address the potential issue of untrusted props - though not the one of leaking secrets.

jamesli2021:
What will happen if we decide to include localization? We would need to change the term 'Guest'.

I think the fallback could be made easier by residing in the components itself with:

```
...
<Fragment slot="fallback">
<div>Guest</div>
<Fragment>
```

Or it is going to result in more breaking changes across projects in the future especially for theme developers.

proposals/server-islands.md
Fryuni: I think it would be best to use a built-time-generated symmetric key use with some algorithm available under the Web Crypto standard. That way it is available on basically every JS server runtime.

thomasledoux1:
Not sure if this is the place to comment on this, but I tried this out and it works great in dev mode, but once deployed to Vercel, it does not work (yet).
I have the following dependency installed for the adaptor
"@astrojs/vercel": "^0.0.0-server-islands-20240709194928"
On the deployed version I see an API call going out for the server island, but that API calls gets back a 404.
Example can be found here:
https://website-thomas-astr-git-8c2233-thomas-ledouxs-projects-74351fc8.vercel.app/blog/search-static-astro-website

Branch: https://github.com/thomasledoux1/website-thomas-astro/tree/feature/server-islands
Relevant file: https://github.com/thomasledoux1/website-thomas-astro/blob/feature/server-islands/src/layouts/BlogLayout.astro

matthewp:
@thomasledoux1 thank you, that is definitely odd. Are you using `'hybrid'` or `'server'` `output`?

thomasledoux1:
> @thomasledoux1 thank you, that is definitely odd. Are you using `'hybrid'` or `'server'` `output`?

I’m using hybrid mode

PeterDraex:
Is it possible that the code running in the browser will be outdated - that the server will be running a newer version than the browser? Maybe this could happen due to browser cache? Or CDN cache? 

In such cases, the request to server islands could fail - for example, when the server island has been renamed in the new version. Should this be addressed?

jamesli2021:
In one scenario, if the page is served from a CDN is an outdated content, what would happen if the script on client-side attempted to render components that are no longer valid or at the same time you have deploy updated Astro? Would the problem be similar to that of a Single-Page Application (SPA)?

ascorbic:
@jamesli2021 yes, it would be similar to an SPA, but without the problems that SPAs have of staying open in a browser. As you identify, it's only an issue for the CDN. The solution there is to invalidate pages when you do a new build. Lots of hosts will do that automatically (Netlify, Vercel, Cloudflare Pages for example all do it by default), but if you're running your own server with a separate CDN then you'd need to set it up to do the invalidation.

Tc-001:
Maybe an integration could define a custom fetcher so platforms like Vercel could take advantage of skew protection?

For something like cloudflare the fallback could instead be the deploy-specific URL. Not *ideal* as cookies would break, but the island will be at least rendered.

Fryuni:
This is something that can already be done on the adapters. It only needs to modify fetch requests to indicate the build version the frontend comes from and direct to the appropriate version on the backend, even though it is the URL for the "latest" release.

Vercel, for example, has this as ["coming soon" for Astro](https://vercel.com/docs/deployments/skew-protection#supported-frameworks). It could be added relatively quickly to the adapter.

I don't know about an equivalent feature on other platforms tho

Tc-001:
> I don't know about an equivalent feature on other platforms tho

I don't know if there are any other ones, although maybe an adapter could add a middleware with the same function, this would also keep the cookies.

Something like the below could be automatically added (maybe behind a toggle) 
```js
app.get("/_astro/skew/:version/*", () => {
 return fetch(`https://${version}.${name}.pages.dev/${req.path}`)  
})
```

But that also needs safeguards for SSRF stuff

matthewp:
@Fryuni I'm pretty sure we already implemented this for Vercel, cc @ematipico

jamesli2021:
Thanks for enlightening me!

ematipico:
Yeah we already did :) not sure why they didn't update it

jamesli2021:
There is one scenario on our site which has <Nav/> on the layouts, if we have some case we can't easily modify the components, we can defer using `<Fragment>`? I have tested out experimental Server Islands and it does not work for my use case.

Layouts.astro
```
<Fragment server:defer>
    {Astro.url.pathname !== '/' && <Nav image={'myAvatar'} />}
</Fragment>
```

When the page is pre-render, this mean I would need to move all the logic into my components.

jamesli2021:
If the client-side JavaScript is included in the components, it will execute before the components are loaded.

matthewp:
@jamesli2021 you can't `server:defer` a fragment. It needs to be a component you import.

davidgauch:
I would like to throw in my 2 cents about GET vs POST discussion.
My understanding is that server-islands works best with static-site generation.
Because of that we have the opportunity to raise an error at SSG time.
That would mean that we could potentially give the user 3 options (and possibly the ability to set it globally)

1. GET. Triggers s SSG failure if request is too large, (fallback with error log in other envs).
2. GET with Fallback. Try to use GET, silently fallback to POST when too large. (I think this should be mentioned prominently in the build message.)
3. POST. I foresee some people wanting this, potentially as a quick and dirty `no-cache` option.

I know that for my projects I would use Option 1 as the default behavior in my apps to help me ensure I'm maximally leverage CDN caching.

PeterDraex:
> I know that for my projects I would use Option 1 as the default behavior in my apps to help me ensure I'm maximally leverage CDN caching.

What kind of CDN caching are you referring to? @davidgauch 

I think typically, `server:defer` would be only used for resources, which cannot be part of the static build, because they are user-specific or change frequently. To cache such resources, you would need to invalidate the cache for a specific URL when data change. If you can do that, you are probably not using SSG in the first place, but rather run astro in SSR mode and cache the output ([incremental static regeneration](https://vercel.com/docs/frameworks/astro#incremental-static-regeneration)).

I think caching `server:defer` requests makes sense only in niche cases - when you have an Astro component, which is big (many kB of HTML) and it's not critical for First Contentful Paint. Deferring it's loading until later would give you a better FCP time. Is that your case?

PeterDraex:
I'm in favor of using `GET` requests by default, together with HTTP header preloading. It will make the pages noticeably faster.

davidgauch:
@PeterDraex 

> What kind of CDN caching are you referring to? @davidgauch 

I was re-iterating the caching mentioned in the 4th bullet in the "GET vs POST" section of [this (now closed) ticket](https://github.com/withastro/roadmap/issues/945#GET-vs-POST). No CDN I have used has supporting caching based on request bodies.


> you are probably not using SSG in the first place, but rather run astro in SSR mode and cache the output

Sorry about being unclear. I should have said 'hybrid' page rendering which, when I use it, statically generates most of my performance-sensitive pages.

> Is that your case?
Close enough. The thing I most want it for is not too large, it is just somewhat slow to generate and can be stale. Ideally I would slap on `stale-while-revalidate` in a header and reduce the number of requests that make their way to my backend.

jamesli2021:
Once the cookie is set in the middleware, it doesn't persist the cookies in server mode with pre-rendered pages or hybrid mode. It seems to be designed primarily for static pages.

matthewp:
Server Islands are not just for SSG. The demo app shows an e-commerce site, which would most likely be SSR in order to add new products without redeploying.

As far as the GET vs. POST, I very much favor being able to do GET as well. That's why I want to avoid adding extra features some have asked for like providing the origin URL, as those things would bloat the URL. 

I do think in most cases the props will be small enough that GET is possible. I would probably favor approach (2) from @davidgauch's proposal here as it's the most straightfoward to implement. 

Also before rushing into config we need to confirm that GET is practical, so we need some data on that. We can implement that as part of the experimental flag and see how well it works.

Will probably do prop encryption first since that will affect the size, then follow up with this idea.

jamesli2021:
Just to get my understanding if e-commerce demo in SSR with no pre-render can still leverage on Server Islands?

matthewp:
Right, prerendered not required. Can use regular SSR. You want to set cache headers in that case though, as having cached main page content is the primary point.

matthewp:
Here's what I'm tracking needs to be added to the RFC before it's complete:

1.  Encrypting props.
1. Use `GET` if possible with `POST` fallback, including adding preload links to speed up the island loading.
1. Handle non-200 responses from islands.

----

There's a lot of discussion already about (1) and (2) already in this thread, I'm curious to hear what people think we should do about (3)?

My initial instinct is that if the response is 4xx or 5xx we should emit an event and not render the response to the DOM. I do need to check what libraries like htmx do here for reference.

ascorbic:
> I'm curious to hear what people think we should do about (3)

I wonder if there's value in an `error` slot, so there can be a separate fallback state that's not "loading"

matthewp:
That's a pretty good idea. The one downside to a slot is that you need to define it on every island you use. I imagine a lot of people won't do so, so we'd still probably want to have an event. Maybe this makes sense then:

- `slot="error"` this will render statically, inside of a `<template>`.
  - If there's a 200 it gets wiped out along with the fallback content. 
  - If there's a non-200 it's used as the replacement for the fallback.
- If not error slot is provided an `astro:server-island-error` event is fired.

Fryuni:
If we are going to add events it would be nice to have an `astro:server-island-content` or similar so client code could react to a server island receiving the updated content.

matthewp:
@Fryuni what sort of thing would you do in that event?

Fryuni:
Tracking and performance monitoring from the user's perspective. Server might be replying fast but we could measure how long is it taking to actually get the information on the page.
That way we can account for network, queueing and processing time to see what is the real experience.

matthewp:
Have a draft up for encrypted props: https://github.com/withastro/astro/pull/11535 Still a few things to do. Would appreciate anyone who can review this to make sure I did things generally right.

tedevra:
Heyho,

i have an issue trying to make the server-islands demo work with my catchall page route and cloudflare adapter, maybe you guys know whats going on.

as soon as i switch 
`/src/pages/index.astro`
to 
`/src/pages/[...index].astro`

running "`npm run dev`" (`astro dev`)
=> everything still works....

running "`npm run preview`" (`astro build && wrangler pages dev ./dist`)
=> does not work anymore and ends up in a loop (rerendering another page instance again and again, breaking the server-island -> default fallback)

code as zip
[astro-server-islands-poc.zip](https://github.com/user-attachments/files/16400981/astro-server-islands-poc.zip)

working as expected or is there a misinvocation of the catchall page route during the `/_server-islands/*` request?

any help is highly appreciated, thanks a lot and have a nice day
Tobi

screenshot:
left side: run dev
right side: run preview
image
image

jamesli2021:
How can we cache individual components if we have content that will likely remain unchanged over time, or revalidate as set in the header's cache control, such as the navigation menu and footer?

ascorbic:
@jamesli2021 I think it will depend on implementing GET request support, but you should be able to just set response headers as usual inside the island, using `Astro.response`

jamesli2021:
@ascorbic I think the naming convention for server islands could pose problems when we deploy changes, as previous components may still be cached (e.g., Element1, Element2, etc.).

My idea is to use Unix time as part of the naming `/server-islands/element2?1818181818` or `/server-islands/element1818181818` to properly invalidate the old cache, which could solve these issues.

ascorbic:
@jamesli2021 that's the same with any page though. As with any other page it's your responsibility to handle the cache and purge it if needed. Hosts like Netlify and Vercel will handle it automatically, or you do it yourself.

danthegoodman1:
Is it possible to handle the case of rendering a page where you are dynamically listing something, and for example you want to:
1. Server side render direct page loads with search (incl empty search) from a search param
2. Switch to server islands when doing subsequent searches on the page

In Remix I'd do this with a `form.Fetcher` and just recall the loader, but I'm not sure if in Astro I can dynamically switch between a list that was SSR, then re-render that with a `server:defer` component (basically dynamically adding that).

That would still allow caching for the no search result and no search pagination, but if someone did text search then we'd hit the server always and modify the response to not cache, if that makes sense.

matthewp:
I've implemented the `GET` behavior as discussed in this thread, including the preload link. It's available in a preview release: `astro@experimental--si-get`

I deployed it to server-islands.com and ran the tests a few times. Here are the relevant waterfalls:

- [Using GET](https://www.webpagetest.org/result/240815_AiDc6J_B2N/1/details/#waterfall_view_step1)
- [Using POST](https://www.webpagetest.org/result/240716_AiDc6D_D1N/1/details/#waterfall_view_step1)

## GET

Screen Shot 2024-08-15 at 2 49 38 PM

## POST

Screen Shot 2024-08-15 at 2 49 45 PM

## Results

Using GET with the preload is causing these requests to start between 180 - 400ms faster. Looking at the waterfalls I think the biggest factor here is that the preloads are being kicked off before images. Whereas with POST the requests are competing with the images. This might be a case where this particular app has a larger effect than others will.

matthewp:
We're hoping to ship this in 5.0, so I am doing a **call for consensus** on this RFC. This allows us to take the feature out of experimental and mark it as stable. The CFC period lasts a minimum of 3 days, if there are concerns about the API please raise them now. Bugs should be filed in the core repo, thanks!

danfascia:
> That's a pretty good idea. The one downside to a slot is that you need to define it on every island you use. I imagine a lot of people won't do so, so we'd still probably want to have an event. Maybe this makes sense then:
> 
> * `slot="error"` this will render statically, inside of a `<template>`.
>   
>   * If there's a 200 it gets wiped out along with the fallback content.
>   * If there's a non-200 it's used as the replacement for the fallback.
> * If not error slot is provided an `astro:server-island-error` event is fired.

An error slot would be an amazing idea...

If you could have slots for error and loading state I think so much work of the typical suspense process that you repeat over and over would be dealt with. It would very much emulate the behaviour of a try... catch... finally... sequence

Issue #946, opened 2024-06-07
Content Layer


matthewp:
- Accepted Date: 2024-06-07
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/935
- Author: @FredKSchott 
- Champion(s): 
- Implementation PR: 

# Summary

- Explore a new and improved content layer for Astro.
- Improve the current experience of loading/defining data into content collections
- Improve the current experience of querying data from content collections

# Background & Motivation

Content Collections are a key primitive that brings people to Astro. Content Collections make it easy to work with local content (MD, MDX, Markdoc, etc) inside of your Astro project. They give you structure (`src/content/[collection-name]/*`), schema validation for frontmatter, and querying APIs.

# Goals

- Explore a new and improved content layer for Astro.
- Improve the current experience of loading/defining data into content collections
- Improve the current experience of querying data from content collections

# Example

```js
// A folder full of Markdown (MDX) files
defineCollection({
    name: 'blog',
    data: glob('./content/blog/*.mdx'),
  });
// A single file containing an array of objects
defineCollection({
    name: 'authors',
    data: file('./content/authors.json'),
});
// Remote data, loaded with a custom npm package
defineCollection({
    name: 'articles',
    data: storyblokLoader({startsWith: 'articles/posts'}),
});
// Custom data, loaded from anywhere you'd like
defineCollection({
    name: 'my-custom-collection',
    data: () => { /* ... */ },
});
```

florian-lefebvre:
I think it would be nice to support singletons, as suggested in #449 and #806. Many CMSs support those (eg. Keystatic, the content layer seems like it would fit well)

lloydjatkinson:
Just wondering how this will work (if at all) for components, like in MDX? So the use case I'm thinking is when a CMS is used, they typically have a WYSIWYG rich text editor where custom components can be inserted as shown here: https://www.storyblok.com/tp/create-custom-components-in-storyblok-and-astro

Will this new API support this concept?

xavdid:
This is cool! 

I do a sort-of version of this for my review site ([david.reviews](https://david.reviews/)), where all the data is stored in Airtable.

I [load objects from the Airtable API](https://xavd.id/blog/post/static-review-site-with-airtable/) based on a schema and cache responses in JSON locally (since loading and paging takes a while and is too slow for local development). It sort of feels like what an API-backed custom content collection could look like.

[...]

PR #982, opened 2024-07-23
Content Layer

ascorbic:
# Summary

Creates a successor to content collections with expanded use cases, including remote data sources and improved performance.

```ts
// src/content/config.ts
import { defineCollection, z } from "astro:content";
import { glob, file } from "astro/loaders";
import { feedLoader } from "@ascorbic/feed-loader";

// Loaders can be defined inline 
const countries = defineCollection({
  loader: async () => {
    const response = await fetch("https://restcountries.com/v3.1/all");
    const data = await response.json();
    // Must return an array of entries with an id property, or an object with IDs as keys and entries as values
    return data.map((country) => ({
      id: country.cca3,
      ...country,
    }));
  },
});

// Loaders can also be distributed as packages
const podcasts = defineCollection({
  loader: feedLoader({
    url: "https://feeds.99percentinvisible.org/99percentinvisible",
  }),
});

// The `glob()` loader loads multiple files, with one entry per file
const spacecraft = defineCollection({
  loader: glob({ pattern: "*.md", base: "src/data/spacecraft" }),
  // A schema is optional, but provides validation and type safety for data.
  // It can also be used to transform data before it is stored.
  schema: ({ image }) =>
    z.object({
      title: z.string(),
      description: z.string(),
      heroImage: image().optional(),
    }),
});

// The `file()` loader loads multiple entries from one file
const dogs = defineCollection({
  loader: file("src/data/dogs.json"),
  schema: z.object({
    id: z.string(),
    breed: z.string(),
    temperament: z.array(z.string()),
  }),
});

export const collections = { spacecraft, dogs, countries, podcasts };
```

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/content-layer/proposals/0050-content-layer.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/946)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/935)


proposals/content-layer.md
```diff
@@ -0,0 +1,345 @@
...
+- Make the API scalable to tens of thousands of entries, with good performance and low memory usage.
+- Allow loaders to define their own schemas, including dynamic schemas that can be introspected from the data source.
+
+# Non-Goals
+
+- Non-goal: allowing loaders to define multiple collections automatically. e.g. separate collections would need to be manually defined for posts and categories in a blog.
+- Non-goal: dependency tracing for entries.
+- Out of scope: hot-reloading remote data.
+- Out of scope: rendering markdown from remote data. A loader could store rendered HTML, but it would be up to the loader to handle this.
+- Out of scope: custom `Content` components.
```
ematipico, line 90:
```suggestion
- Allowing loaders to define multiple collections automatically. e.g. separate collections would need to be manually defined for posts and categories in a blog.
- Dependency tracing for entries.
- Hot-reloading remote data.
- Rendering markdown from remote data. A loader could store rendered HTML, but it would be up to the loader to handle this.
- Custom `Content` components.
```

proposals/content-layer.md
```diff
@@ -0,0 +1,345 @@
...
+- Future: support for Markdoc and MDX rendering.
+- Future: SQLite-based backend for collections.
+- Future: support for queries more complex than get by ID.
```
ematipico, line 93: We could create a `## Stretch Goals` header. Or not mention them at all. I will remove the SQLite part for now.

proposals/content-layer.md
```diff
@@ -0,0 +1,345 @@
...
+
+# Detailed Design
+
+## Collection Definition
+
+Collections are defined in a similar way to current content collections, using `defineCollection()` in `src/content/config.ts`. There are two new collection types: `experimental_content` and `experimental_data`. The only difference between them is that when a collection is defined as `experimental_content`, entries will have a `render()` method that generates a component for rendering HTML content. This requires the loader to have set the `rendered.html` property.
```
ematipico, line 99: We should write the RFC as if it were production-ready. We will mention `experimental_ content` and `experimental_data` as part of the adoption strategy.

If we don't know yet how users will opt-in the content layer, I believe it's OK to keep it as `TBD` for now, since this is a draft PR. 

Another suggestion: start the design with a glossary, or some preface, where we lay down new terminologies and concepts. For example, we should explain first what's a "Loader" and what's its role in the content layer proposal.

proposals/content-layer.md
```diff
@@ -0,0 +1,345 @@
...
+
+The `reference()` helper can be used in the same way as content collections, to reference other collections.
+
+## Built-in loaders
+
+There are two built-in loaders: `file()` and `glob()`, which load data from the local filesystem. The `glob()` loader covers the current use case of directories full of markdown or JSON content. The `glob()` helper is more flexible than in current content collections, as it can load data from anywhere on the filesystem. The `file()` loader loads multiple entries from a single file. Both loaders can process markdown in the same way as content collections. They can also extract images in the same way as content collections.
```
ematipico, line 105: My suggestion: I would change the narrative of the RFC like this:
1. Start by showing how to define content and render it, in the *most minimal* way. This is an excuse to show how **simple** is the new API, because a user can use **a simple JS function** to define some data.  
2. From there, let's talk about the `loader` API and explain it. Talk about its design (astro/vite-like), then let's get technical and explain the other APIs, e.g. `watcher`, `store`, etc. We will talk about the fact that the `store` is scoped to the current collection. Here we can be as technical as we want. We can also talk about "behind the scenes" business logic, if it is relevant to us developers, to provide context when reviewing the PRs.
3. Eventually, we can **mention** our built-in loaders, but I believe there's no need to go into detail too much since they use the APIs that were explained before. However, I believe it's worth mentioning how `glob` **could be used** for an almost built-in replacement of the current content collections.

proposals/content-layer.md
```diff
@@ -0,0 +1,345 @@
...
+
+- Create a successor to content collections that can be used with local and remote data.
+- Decouple content from Vite, so that data is no longer implemented as Rollup chunks.
+- Provide a simple API for defining collections with a migration path from content collections.
+- Support local files in user-defined locations with built-in file and glob loaders.
+- Initially support Markdown rendering and JSON data for local files, with support for other formats in the future.
+- Provide an API for defining custom loaders that is powerful enough to handle a wide range of use cases, including CMS integrations.
+- Provide a higher-level API for simple inline loaders.
+- Make the API scalable to tens of thousands of entries, with good performance and low memory usage.
+- Allow loaders to define their own schemas, including dynamic schemas that can be introspected from the data source.
```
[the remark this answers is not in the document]
ascorbic: I included it because the current implementation _isn't_ very scalable, and wanted it to be a specific goal to change that. I've changed the wording to make it clear that it's the scalability that's the goal

proposals/content-layer.md
```diff
@@ -0,0 +1,345 @@
...
+# Goals
+
+- Create a successor to content collections that can be used with local and remote data.
+- Decouple content from Vite, so that data is no longer implemented as Rollup chunks.
+- Provide a simple API for defining collections with a migration path from content collections.
+- Support local files in user-defined locations with built-in file and glob loaders.
+- Initially support Markdown rendering and JSON data for local files, with support for other formats in the future.
+- Provide an API for defining custom loaders that is powerful enough to handle a wide range of use cases, including CMS integrations.
+- Provide a higher-level API for simple inline loaders.
+- Make the API scalable to tens of thousands of entries, with good performance and low memory usage.
```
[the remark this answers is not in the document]
ascorbic: I included it because the current implementation isn't very scalable, and wanted it to be a specific goal to change that. I've changed the wording to make it clear that it's the scalability that's the goal

proposals/0047-content-layer.md
```diff
@@ -0,0 +1,399 @@
...
+---
+
+<h1>{craft.data.title}</h1>
+<p>{craft.data.description}</p>
+
+```
+
+### Rendered content
+
+Some entry types may have HTML content that can be rendered as a component. While this can be accessed like any other property, a loader can also store the rendered HTML in the `rendered.html` property. This allows users to use the `<Content />` component to render the HTML. The `rendered` property can also include metadata such as frontmatter or headings, which can be accessed as properties on the `rendered.metadata` object.
```
florian-lefebvre, line 307: I think it would be useful to add an example loader for this

proposals/0047-content-layer.md
```diff
@@ -304,7 +302,26 @@ const { craft } = Astro.props;
...
+```ts
+// src/content/config.ts
+store.set({
+  id,
+  data,
+  rendered: {
+    // A raw HTML string
+    html: data.description ?? "",
+    metadata: {
+      // Optionally, metadata such as headings can be stored here
```
florian-lefebvre, line 316: Can there be anything else then headings?

krishna-santosh:
@ascorbic can we define schema when fetching data from a remote URL?

krishna-santosh:
@ascorbic can we define a schema when fetching data from a remote URL?

ascorbic:
@krishna-santosh yes. You can either define it at the top level (inside the `defineCollection` object) in the same way as now, or a loader can define it. In that case the loader can generate it dynamically, e.g. by introspecting the API.

proposals/0047-content-layer.md
ascorbic: Yes, it's arbitrary, schemaless metadata. Markdown uses it for image paths and raw frontmatter too.

simonswiss:
Multiple entries per file is great — but how about multiple files per entry?

There are many cases where an entry has a `heroIntro` paragraph which wants to be a rich text field, as well as the main entry body.

The ability to store multiple files part of the same entry would be great!

Alternatively, storing other Markdown (Markdoc/MDX) fields in the frontmatter (we do this with [Keystatic's `markdoc.inline` field](https://keystatic.com/docs/fields/markdoc#inline)) and providing a similar `.render()` method on those would be awesome.

Apologies if this is not the right place for this feedback 🤗

joelvarty:
Let me know if this is the wrong place for this, but I wanted to provide feedback on what I feel would be important for the success of this layer from a CMS perspective:

### Dependancy Tracking
- Being able to track dependancies automatically is VERY useful if I have content that is being used in multiple places and I invalidate it.
- Being able to incrementally update the data layer should work hand in hand with that dependancy checking.

### Function Support 
- As I understand it, this layer will exist as files somewhere - how will that work in runtimes that don't allow file system support.

ascorbic:
@joelvarty yes, we'd love to do dependency tracking at some point. We are in a good place to do that, but it's not in scope for this version.

The data is read-only at runtime, and even though it's implemented as files on disk it's loaded as a virtual module, so it's all handled by rollup and works on serverless runtimes. I haven't tested it with Cloudflare yet though, so we'll need to ensure that works.

joelvarty:
@ascorbic It sounds like we would need to do a `build` to get new data into the project.  Am I correct in that?

ascorbic:
@joelvarty Yes, there's no way to update a deployed site without a build. Locally you can sync by running `astro build`, `astro sync` or when running `astro dev` you can use the `s + enter` hotkey

Suven:
@ascorbic there is no chance of exporting globalContentLayer right? I am generally doing SSG, but am currently using an instance of `npm run dev` for live-previews of the CMS. If I had access to the global instance, I could do some trickery like resyncing every n pageviews or all X seconds.

ascorbic:
@Suven No (and please don't try: you will break things), but I will be exposing a refresh method in `astro:server:start` for integrations to use

ascorbic:
I have a proposed addition to the RFC, which I'd welcome feedback on.

## Integration support for content layer

The is a proposal to add support for syncing the content layer to integrations. It would allow integrations to trigger a sync during dev, optionally of just certain loaders. It would also allow them to pass metadata such as a webhook body to the loader

### Use cases

- Adding a sync button to the Astro toolbar
- Opening a web socket to a CMS that listens for updates
- Creates a webhook URL that can be tunnelled to a public address for CMSs to trigger
- Allow e.g. code sandboxes or hosted dev serversa to trigger reloads

### API

Adds a `syncContent` function to the `astro:server:setup` hook options, with the following signature:

```ts
async syncContent(options: { 
   loaders?: Array<string>,
   context?: Record<string, any>
})
```

`loaders` is an optional array of loader names. If set, only those loaders will be synced. This allows integrations to selectively sync their own content.

`context` is an optional object with arbitrary data that is passed to the loader's `load` function as `syncContext`.

### Usage

This shows an integration that creates a refresh webhook endpoint:

```ts
export default function() {
    return {
        name: '@astrojs/my-integration',
        hooks: {
            'astro:server:setup': async ({ server, refreshContent }) => {
                server.middlewares.use('/_refresh', async (req, res) => {
                  	if(req.method !== 'POST') {
                      res.statusCode = 405
											res.end('Method Not Allowed');
                      return
                    }
                    let body = '';
                    req.on('data', chunk => {
                        body += chunk.toString();
                    });
                    req.on('end', async () => {
                        try {
                          	const webhookBody = JSON.parse(body);
                            await refreshContent({
                              // Include the parsed request body. `webhookBody` is an arbitrary name
                              context: { webhookBody },
                              // Only refresh a particular loader
                              loaders: ['my-loader']
                            });
                            res.writeHead(200, { 'Content-Type': 'application/json' });
                            res.end(JSON.stringify({ message: 'Content refreshed successfully' }));
                        } catch (error) {
                            res.writeHead(500, { 'Content-Type': 'application/json' });
                            res.end(JSON.stringify({ error: 'Failed to refresh content' }));
                        }
                    });
                });
            }
        }
    }
}
```

Inside the loader:

```ts
import { type Loader } from "astro/loaders"
export function myLoader(): Loader {
  return {
    name: "my-loader",
    load: async ({ store, logger, syncContext, meta }) => {
      if(syncContext?.webhookBody?.action) {
        logger.info("Received incoming webhook")
        // do something with the webhook body
      }
      // this is a normal sync...
    }
  }
}
```

### Questions

How should we handle the case where a sync is already in progress? Should it be queued? Should it be skipped?

matthewp:
Outside of you using `refreshContent` in one of the examples (which I kind of like better...) I think the idea sounds reasonable and `astro:server:setup` is the place to do it.

ascorbic:
Yeah, maybe `refreshContent` is better!

modestaspruckus:
Hello, nice feature
Regarding this `The store is persisted to disk between builds, so loaders can handle incremental updates`.

Can we control the path and what is the location of store in disk? We'd like to mount it as PVC on k8.

florian-lefebvre:
It's currently stored to the [cacheDir](https://docs.astro.build/en/reference/configuration-reference/#cachedir) (configurable)

NotWoods:
Is it possible to set up deferred rendering by passing a function/promise to `rendered`? The current `deferredRender` property seems like it's designed around local files, but loaders that pull from a third-party API could get a performance boost by just fetching metadata then fetching the HTML only when needed.

ematipico:
> Is it possible to set up deferred rendering by passing a function/promise to `rendered`? The current `deferredRender` property seems like it's designed around local files, but loaders that pull from a third-party API could get a performance boost by just fetching metadata then fetching the HTML only when needed.

Did you have anything in particular in mind? The deferred renderers are designed around virtual modules, so you can pull your content from them?

Suven:
That proposal sounds great! Regarding your question: I guess triggering a sync while syncing is very likely in CMS-Preview-contexts as the user is continuously making changes. In those I guess the user is only interested in the latest version of his change and this, cancelling the current sync and starting a new one would make sense.

NotWoods:
> > Is it possible to set up deferred rendering by passing a function/promise to `rendered`? The current `deferredRender` property seems like it's designed around local files, but loaders that pull from a third-party API could get a performance boost by just fetching metadata then fetching the HTML only when needed.
> 
> Did you have anything in particular in mind? The deferred renderers are designed around virtual modules, so you can pull your content from them? 

The virtual module API is a lot of overhead if you're not working with files on a file system. I'm writing a [loader for the Notion REST API](https://github.com/NotWoods/travel/tree/main/packages/notion-astro-loader) and it feels weird to make a virtual module per API call. After chatting with folks on the Discord I get the impression virtual modules isn't design for this use case.

From an ease-of-loader-implementation perspective a function that can use dynamic imports feels much simpler. I don't know if that makes it harder to optimize the file loader use case.

werfred:
Hi everyone,

I recently started using Astro and wanted to achieve SSG with caching so that pages wouldn't rebuild if nothing had changed. I have an API where all the data comes from, and before the ContentLayer feature, I noticed that Astro has a content collection cache. So, I generated a bunch of JSON files for every blog post into a collection. However, unfortunately, it still rebuilds every page during each build and says: 'generating static routes.'

I then converted my collection to try using ContentLayer, but it seems to work the same way. Am I missing something? I was expecting it to NOT rebuild existing pages if the corresponding JSON file hadn't changed.

Sorry if this isn't the right place to ask about it.

yeehaa123:
Maybe I'm doing something wrong, but references don't seem to work with this API. Is this correct? Is this something that will be implemented down the line? Is there any way I can help out?

ematipico:
@werfred this isn't the right place, please come to discord and open a new help thread

twodft:
I'm wondering if the render also works for remote MDX, if not, how can I implement my own loader to correctly render our remote MDX content from APIs

ascorbic:
> Maybe I'm doing something wrong, but references don't seem to work with this API. Is this correct? Is this something that will be implemented down the line? Is there any way I can help out?

They should work in the same way as existing collections. If you can't get the working, can you ask in Discord

matthewp:
@ascorbic what happens if there's a conflict between the schema the loader is providing and the schema the user provides in `defineCollection`?

ascorbic:
I've added [a section on integration support](https://github.com/withastro/roadmap/blob/content-layer/proposals/0047-content-layer.md#integration-support) to the RFC, and have [a PR](https://github.com/withastro/astro/pull/11878) with an implementation.

HiDeoo:
Tiny feedback on the Content Layer loader API, not sure if it's on purpose or not in this context: if you throw an `AstroError` with an hint in a loader, the hint is never displayed to the user. I think it would be nice to display it, e.g. to help user with obvious configuration mistakes, etc.

ascorbic:
I've made a few small changes to the RFC, with the only API change being that the type for the data store is now `DataStore`. We're now ready for a **call for consensus**. If you have any final comments please make them now. Otherwise I'll merge this in three days.

There will be follow-up RFCs for future features, particularly the libSQL backend.

proposals/0050-content-layer.md
```diff
@@ -0,0 +1,494 @@
...
+- Opening a web socket to a CMS that listens for updates during dev
+- Creates a webhook URL that can be tunnelled to a public address for CMSs to trigger
+- Allow services such as code sandboxes or hosted dev servers to automatically trigger reloads
+
+### API
+
+Adds a `refreshContent` function to the `astro:server:setup` hook options, with the following signature:
+
+```ts
+async syncContent(options: {
```
Princesseuh, line 402:
```suggestion
async refreshContent(options: {
```
?

jcayzac:
Some user with a lot of content recently asked about it on Discord, if I recall correctly: how about making the loaders generator functions that yield entries, rather than async functions that have to return entire collections?

hfournier:
The comment on the built-in file loader example is:
```
// The file loader loads a single file which contains multiple entries. The path is relative to the project root, or an absolute path.
// The data must be an array of objects, each with a unique `id` property, or an object with IDs as keys and entries as values.
```
So, a data structure that may already have some other unique identifier, must also have an `id` property, which seems redundant. It results in data that looks like this:
```
[
  {
    id: 'abc',
    data: {
      id: 'abc',
      myUniqueId: 'abc',
      name: 'Abc'
    },
    filePath: 'src/myFolder/myData.json',
    collection: 'myData'
  },
  ...
```
with 3 properties with the same value.
Would it be possible to add an optional 2nd param to the `file()` loader that indicates which property to use as the `id`?
This would eliminate one redundant property and not require existing json files to be altered with an additional `id` for each entry.

ascorbic:
@hfournier that's an interesting idea. I'm not sure about having it as a second argument, but it is something I'll look at

ascorbic:
This is now included in Astro 5 beta

jurajkapsz:
Not sure if this is still the right place to give feedback - I've followed a link from docs - as Astro 5 beta recently came out, but I've tried out this API and my page styles were afterwards rendered somehow broken. I have the latest Astro v4 release `v4.15.6` and I use `mdx` in content collections.

Does this have something to do with `@astrojs/mdx`, which is mentioned in [Astro 5 beta docs](https://5-0-0-beta.docs.astro.build/en/guides/upgrade-to/v5/#astrojsmdx) to be of `v4`, while with Astro v4 its `v2.3.1`?

Looking forward for this API.

ascorbic:
@jurajkapsz can you open an issue on the main astro repo please

jurajkapsz:
@ascorbic OK, will do, atm I am doing some tests on my side to better understand what happened and to eventually write a more precise bug report.

I've noticed that what visually broke pages after switching to the new Content Layer API render was a different final order of processed CSS styles, which changed style cascade, making eg CSS BEM modifiers unusable.

I happen to have certain style definitions in one component and modifying style definition of given component in another component. I'd say it is correct to have it that way, but I give it a second thought; anyhow it worked before, and the final styles where somehow in correct order.

Issue #1037, opened 2024-10-07
Fonts


florian-lefebvre:
- Accepted Date: 2024-10-07
- Reference Issues/Discussions: #1033
- Author: @florian-lefebvre
- Champion(s): @florian-lefebvre 
- Implementation PR: N/A

# Summary

Have first-party support for fonts in Astro.

# Background & Motivation

Fonts is one of those basic things when making a website, but also an annoying one to deal with. Should I just use a link to a remote font? Or download it locally? How should I handle preloads then?

The goal is to improve the DX around using fonts in Astro.

> Why not using fontsource?

Fontsource is great! But it's not intuitive to preload, and more importantly, doesn't have all fonts. The goal is to have a more generic API for fonts (eg. you want to use a paid provider like adobe).

# Goals

- Specify what font to use
- Cache fonts
- Specify what provider to use
- Load/preload font on a font basis
- Generate fallbacks automatically
- Performant defaults
- Runtime agnostic

# Non-Goals 

- Runtime API (SSR is supported tho)
- Automatic subsetting
- Font detection

florian-lefebvre:
I think we should have a config that can be easily used, but customized as needed. For example, here is an example of minimal usage:

```js
// astro config
export default defineConfig({
	fonts: {
		families: ["Roboto", "Lato"]
	}
})
```

```astro
---
// layouts/Layout.astro
import { Font } from "astro:fonts"
---
<head>
	<Font family="Inter" preload />
	<Font family="Lato" />
	<style>
		h1 {
			font-family: var(--astro-font-inter);
		}
		p {
			font-family: var(--astro-font-lato);
		}
	</style>
</head>
```

That would try pulling fonts from Google fonts by default.

---

Here is an example of a more involved fonts config that I think would be nice to support

```js
import { fontProviders } from "astro/config"

// astro config
export default defineConfig({
	fonts: {
		providers: [fontProviders.adobe({ apiKey: "ADOBE_API_KEY" })],
		local: {
			directories: ['./my-custom-directory']
		},
		defaults: {
			provider: "local",
			weights: [800,900]
		},
		families: [
			"Roboto", // From local provider
			{
				// From local provider
				name: "Lato",
				weights: [400,600]
			},
			{
				name: "Inter",
				provider: "google",
				subsets: ["latin"]
			},
			{
				name: "..."
				provider: "adobe"
			}
		]
	}
})
```

ematipico:
Some comments regarding the proposal.

[...]

PR #1039, opened 2024-10-15
[WIP] Fonts

florian-lefebvre:
# Summary

Have first-party support for fonts in Astro:

```js
// astro config
export default defineConfig({
	fonts: {
		families: ["Roboto", "Lato"]
	}
})
```

```astro
---
// layouts/Layout.astro
import { Font } from "astro:fonts"
---
<head>
	<Font family="Inter" preload />
	<Font family="Lato" />
	<style>
		h1 {
			font-family: var(--astro-font-inter);
		}
		p {
			font-family: var(--astro-font-lato);
		}
	</style>
</head>
```

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/rfc/fonts/proposals/0052-fonts.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/1037)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/1033)


lilnasy:
How does preloading pick the `src` when the font provides multiple files for multiple unicode ranges? See [IBM Plex Mono on font source](https://github.com/fontsource/font-files/blob/643d5496631adb16c66ab3fc2f9e128d01817307/fonts/google/ibm-plex-mono/400.css#L18) for example.

What's the reasoning behind considering subsetting a non-goal? It helps a lot when the font participates in LCP and is trivial to do when using google fonts directly. Moving to astro fonts in that case would be a downgrade.

proposals/0052-fonts.md
```diff
@@ -0,0 +1,368 @@
...
+- Specify what provider to use
+- Load/preload font on a font basis
+- Generate fallbacks automatically
+- Performant defaults
+- Runtime agnostic
+
+# Non-Goals
+
+- Runtime API (SSR is supported tho)
+- Automatic subsetting
```
florian-lefebvre, line 66:
```suggestion
- Automatic subsetting (eg. analyzing static content)
```

proposals/0052-fonts.md
```diff
@@ -0,0 +1,368 @@
...
+
+# Goals
+
+- Specify what font to use
+- Cache fonts
+- Specify what provider to use
+- Load/preload font on a font basis
+- Generate fallbacks automatically
+- Performant defaults
+- Runtime agnostic
```
florian-lefebvre, line 61:
```suggestion
- Runtime agnostic
- Configure font families (subset, unicode range, weights etc)
```

proposals/0052-fonts.md
```diff
@@ -0,0 +1,369 @@
...
+- Configure font families (subset, unicode range, weights etc)
+
+# Non-Goals
+
+- Runtime API (SSR is supported tho)
+- Automatic subsetting (eg. analyzing static content)
+- Font detection
```
Princesseuh, line 68: Would be great to specify what this means, I know it personally but people might not!

proposals/0052-fonts.md
```diff
@@ -0,0 +1,369 @@
...
+  fonts: {
+    providers: [
+      fontProviders.adobe({ apiKey: process.env.ADOBE_FONTS_API_KEY }),
+      myCustomFontProvider(),
+    ],
+    defaults: {
+      provider: "adobe",
+      weights: [200, 700],
+      styles: ["italic"],
+      subsets: [
```
Princesseuh, line 106: would love an example with opentype fonts features

proposals/0052-fonts.md
```diff
@@ -0,0 +1,369 @@
...
+import { myCustomFontProvider } from "./provider";
+
+export default defineConfig({
+  fonts: {
+    providers: [
+      fontProviders.adobe({ apiKey: process.env.ADOBE_FONTS_API_KEY }),
+      myCustomFontProvider(),
+    ],
+    defaults: {
+      provider: "adobe",
```
Princesseuh, line 103: how would you know the name of the provider here? Will it be... typed (👀) automatically based on what you put in `providers`?

proposals/0052-fonts.md
```diff
@@ -0,0 +1,369 @@
...
+import { defineConfig, fontProviders } from "astro/config";
+import { myCustomFontProvider } from "./provider";
+
+export default defineConfig({
+  fonts: {
+    providers: [
+      fontProviders.adobe({ apiKey: process.env.ADOBE_FONTS_API_KEY }),
+      myCustomFontProvider(),
+    ],
+    defaults: {
```
Princesseuh, line 102: I think `defaults` make this a nicer API overall, but I wonder how required it is, do people really have that much fonts that setting defaults is required 🤔

proposals/0052-fonts.md
```diff
@@ -0,0 +1,369 @@
...
+    ],
+  },
+});
+```
+
+### Providers
+
+#### Definition
+
+A provider allows to retrieve font faces data from a font family name from a given CDN or abstraction. It's a [unifont](https://github.com/unjs/unifont) provider.
```
Princesseuh, line 137: I think some people could have a concern with us exposing a third-party "API" like that in the config, but we already do for syntax highlighting themes, so I think it's okay. Plus, the default providers unifont provide are probably enough for 99.999% of people anyway.

proposals/0052-fonts.md
```diff
@@ -0,0 +1,369 @@
...
+
+### Defaults
+
+Astro must provide sensible defaults when it comes to font weights, subsets and more. But when dealing with more custom advanced setups, it makes sense to be able to customize those defaults. They can be set in `fonts.defaults` and will be merged with Astro defaults.
+
+We need to decide what default to provide. I can see 2 paths:
+
+| Path      | Example (weight)    | Advantage             | Downside                                                                          |
+| --------- | ------------------- | --------------------- | --------------------------------------------------------------------------------- |
+| Minimal   | Only include `400`  | Lightweight           | People will probably struggle by expecting all weights to be available by default |
```
Princesseuh, line 220: I'm team just 400, Google Fonts also gives you only 400 when you unselect the "give all options" setting

proposals/0052-fonts.md
```diff
@@ -0,0 +1,369 @@
...
+---
+// layouts/Layout.astro
+import { Font } from 'astro:fonts'
+---
+<head>
+	<Font family='Inter' preload />
+	<Font family='Lato' />
+	<style>
+		h1 {
+			font-family: var(--astro-font-inter);
```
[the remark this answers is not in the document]
florian-lefebvre: Well since it's just some css, I don't think it will cause any issues? In Tailwind 3, you'd have to reference like so:
```js
const defaultTheme = require('tailwindcss/defaultTheme')

module.exports = {
  theme: {
    extend: {
      fontFamily: {
        'sans': ['var(--astro-font-inter)', ...defaultTheme.fontFamily.sans],
      },
    }
  }
}
```

proposals/0052-fonts.md
florian-lefebvre: This is not a subet right? I guess that would need to be more keys, something like the following:
```js
{
	fontKerning: "'kern' 1",
	fontVariantAlternates: "1"
	// etc...
}
```

MDN docs: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_fonts/OpenType_fonts_guide

proposals/0052-fonts.md
florian-lefebvre: Yes! Thanks to `defineConfig()` being generic now, we can do a lot of cool stuff here. Providers would be easily typed

proposals/0052-fonts.md
florian-lefebvre: I have to admit I rarely use a lot of fonts but I guess that could be the case for more complex sites. IMO people will probably want this for setting a default provider in most cases, so why not support all the other properties I suppose!

proposals/0052-fonts.md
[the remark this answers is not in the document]
miguderp: Maybe allowthe variable name to be configured in the tag, i.e. `<Font family='Lato' variable='my-custom-font' />` which generates `var(--my-custom-font)`?

This is similar to how [Next.js does it](https://nextjs.org/docs/pages/building-your-application/optimizing/fonts#with-tailwind-css), albeit they use a function syntax instead of a JSX-style component

proposals/0052-fonts.md
```diff
@@ -0,0 +1,369 @@
...
+      fontProviders.adobe({ apiKey: process.env.ADOBE_FONTS_API_KEY }),
+    ],
+    // ...
+  },
+});
+```
+
+#### Why this API?
+
+1. **Coherent API**: a few things in Astro are using this pattern, namely integrations and vite plugins. It's simple to author as a library author, easy to use as a user
```
ematipico, line 208: I disagree with this statement, this API is used **only** for integrations and vite plugins. 

For example: middleware, actions, and content collections (we could consider them an exception though) don't follow this pattern. They use a **known file** to collect users' APIs. Same thing for `@astrojs/db` (but that's integration).

I think that our configuration should inject as less runtime as possible, use good defaults, and maybe the custom providers could follow the same pattern we use with actions and middleware.

Princesseuh: Image services, which are probably the most similar thing do work this way

ematipico: Probably I missed something, please correct me if I misunderstood what you mean. Based on this docs page (https://docs.astro.build/en/reference/image-service-reference/#user-configuration), the image service uses an entry point API, which is different from the one proposed.

Princesseuh: Ah, apologies, I misunderstood what you meant. Thought this was mostly about where they're configured.

Image services do use an entrypoint, but they're configured in the Astro config, when they have settings.

florian-lefebvre: The main advantage of having objects in the config directly for providers is typing support, which imo is super important

florian-lefebvre: Alright so I guess you'd prefer such an API, am I right?
```js
defineConfig({
	fonts: {
		providers: [fontProviders.adobe({ apiKey: "..." })]
		// equal to { name: "adobe", entrypoint: "astro/assets/fonts/adobe", config: { apiKey: "..." } }
	}
})
```
I guess that's doable

ematipico (commented):
Left some comments, mostly around fallbacks

proposals/0052-fonts.md
```diff
@@ -0,0 +1,413 @@
...
+The easiest way to benefit from fallback generation is by doing the following:
+
+```js
+{
+  family: 'Roboto',
+  fallbacks: ['sans-serif']
+}
+```
+
+This will give `Roboto, 'Roboto fallback: Arial', sans-serif`. This way we can infer what local font to use. It's still possible to provide more fallbacks:
```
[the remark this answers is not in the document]
florian-lefebvre: This is what we generate

proposals/0052-fonts.md
```diff
@@ -0,0 +1,413 @@
...
+```
+
+This will give `Roboto, 'Roboto fallback: Arial', 'Times New Roman', sans-serif`.
+
+## How it works under the hood
+
+- Once the config is fully resolved, we get fonts face data using `unifont`
+- We generate fallbacks using `fontaine` and pass all the data we need through a virtual import, used by the `<Font />` component
+- We inject a vite middleware in development to download fonts as they are requested in development
+- During build, we download all fonts and put them in `outDir`
```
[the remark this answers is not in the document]
florian-lefebvre: This is outdated

proposals/0052-fonts.md
```diff
@@ -0,0 +1,413 @@
...
+
+```css
+h1 {
+  font-family: var(--astro-font-inter); /* "Inter", "Inter Fallback" */
+}
+```
+
+## Fallbacks
+
+We try to generate useful fallbacks when possible (and if the user specifies at least one):
```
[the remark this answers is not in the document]
florian-lefebvre: Sure I'll list the conditions required

proposals/0052-fonts.md
```diff
@@ -0,0 +1,471 @@
...
+      },
+    ],
+  },
+});
+```
+
+It also accepts a `as` prop, which will be used when generating CSS. This is useful in the following cases:
+
+1. You are an integration and want to make sure you don't conflict with fonts declared by the user
+2. Your font name can't be properly converted to a CSS variable name
```
ematipico, line 278: Who is "you" in these sentences? I think we should rephrase them:

```suggestion
1. Avoid conflicts between the font names declared by integrations and those declared by the end-users.
2. A font name can't be properly converted to a CSS variable name.
```

As for the second bullet point, we should at least provide an example of when a font can't be converted to a CSS variable.

florian-lefebvre: I agree with improving the wording! As for point 2, I haven't experienced it but I was thinking about non latin characters (eg. chinese): are they allowed as a css variable name?

florian-lefebvre: I found https://www.thingsaboutweb.dev/en/posts/css-variables-part-1 but I'm still unsure. I think there are cases where the variable name contains characters that will be stripped out and `as` will allow the user to have more control

florian-lefebvre: We are already using `cssesc` for view transitions, i think we could reuse it to take care of encoding

proposals/0052-fonts.md
```diff
@@ -0,0 +1,471 @@
...
+
+> Why not using fontsource?
+
+Fontsource is great! But it's not intuitive to preload, and more importantly, doesn't have all fonts. The goal is to have a more generic API for fonts (eg. you want to use a paid provider like adobe).
+
+# Goals
+
+- Specify what font to use
+- Cache fonts
+- Specify what provider to use
```
matthewp, line 57: Can you define provider somewhere?

proposals/0052-fonts.md
```diff
@@ -0,0 +1,471 @@
...
+- Load/preload font on a font basis
+- Generate fallbacks automatically
```
matthewp, line 59: is this configurable? should it be?

proposals/0052-fonts.md
matthewp: Oh sorry, I see that you do that below, nevermind.

proposals/0052-fonts.md
florian-lefebvre: It is not configurable atm. If we can generate one we will and it will take precedence over the fallbacks defined by the user. I guess we could make it configurable, what name do you think would be good? `automaticFallback: false`?

jlarmstrongiv:
# Feedback

First, I’m glad to see font optimizations coming to Astro!  That will really help with layout shifts.  This feedback is meant to improve an already great start on the implementation.

I recently implemented my own version of it, so I have a few suggestions that could improve the Fonts API, based on my own experiences.

## Remove `@capsizecss/metrics` from the default bundle

It’s massive—22.5 MB of lots of small files.  It takes forever to install if you don’t have a fast machine, fast internet, or want to spin up a fast playground.  Please don’t make it part of the default Astro installation.  By optimizing for fonts, we’ve significantly de-optimized the Astro installation for everyone who doesn’t use it.

### Solution 1

Instead, generate metrics on the fly from the font using `import { fromFile } from '@capsizecss/unpack';`

Copy over any fallback metrics you need from `@capsizecss/metrics` like `Times New Roman`, `Arial`, and `Courier New`, before publishing the fonts package.

### Solution 2

Move the Fonts API to its own package.  Yes, it’s large, but now it’s opt-in, not installed by default without an opt-out.

## Fix the issues in [`astro-capo`](https://github.com/natemoo-re/astro-capo/issues)

What does this have to do with fonts?  Well, ordering the tags in the `<head />` is important.  For example, the order of preload links and inline styles is [reversed](https://rviscomi.github.io/capo.js/user/rules/) in the [Font](https://github.com/withastro/astro/blob/eda0fb565dfdafb24d3eba6e95358efc33ed60d5/packages/astro/components/Font.astro#L4) component.

When combining multiple tags into a single component to render in the `<head />`, it become impossible to manually render them in the correct order.  And without fixing the issues in [`astro-capo`](https://github.com/natemoo-re/astro-capo/issues), it may be possible to fix the order of tags in `<head />` in user land.

## Improve fallback weights and italics

Add the bold and italic versions of fallback fonts to the fallback css.  For an example, see this [blog post](https://www.aleksandrhovhannisyan.com/blog/perfect-font-fallbacks/#multiple-weights-and-italics).

It’s particularly great for headline fallbacks and in cases where fonts never load.

florian-lefebvre:
Thanks for the feedback!
- Remove `@capsizecss/metrics`: definitely agree!
- Head ordering: I can swap the order but as soon as you use several times the `Font` component, you'll have to get `astro-capo` fixed I think
- Fallbacks for weights/italics: sure!

jlarmstrongiv:
Glad to hear!  I think the Fonts API will be very helpful for most Astro use cases.  Great work on the implementation.

---

> you'll have to get astro-capo fixed I think

I’ve looked into fixing it before, but I wasn’t familiar enough with it.  

Personally, I think it’s such an important feature that it should be added to core as a built-in component, but that’s a different discussion.

Other built-in Astro components like the View transitions `<ClientRouter />` would benefit from it too.

florian-lefebvre:
I've sent a few PRs already to address your feedback

jlarmstrongiv:
https://github.com/withastro/astro/pull/13626 At first glance, it wasn’t immediately obvious to me where you load metrics for fallback fonts like `Times New Roman`, `Arial`, and `Courier New`, which you may not be able to read from any of the configured font providers (e.g. Google Fonts).  But, you probably have that covered 🚀

https://github.com/withastro/astro/pull/13625 nice, simple fix for head ordering 👍

florian-lefebvre:
We don't read metrics for system fonts

jlarmstrongiv:
When reading the [docs](https://docs.astro.build/en/reference/experimental-flags/fonts/#fallbacks):

> If the last font in the fallbacks array is a [generic family name](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#generic-name), an [optimized fallback](https://developer.chrome.com/blog/font-fallbacks) using font metrics will be generated. To disable this optimization, set optimizedFallbacks to false.

I was under the impression that the optimized font family overrides (`ascent-override`, `descent-override`, `line-gap-override`, `size-adjust`) are generated when I use generic family names, which are mapped to system fonts [here](https://github.com/withastro/astro/blob/eda0fb565dfdafb24d3eba6e95358efc33ed60d5/packages/astro/src/assets/fonts/constants.ts#L24-L38).

That way you end up with something like:

```css
/* 
  font-family:
    Open Sans,
    "Open Sans Fallback: Arial",
    sans-serif;
*/

@font-face {
  ascent-override: 101.6519%;
  descent-override: 27.8625%;
  font-family: "Open Sans Fallback: Arial";
  font-stretch: 100%;
  font-style: normal;
  font-weight: 400;
  line-gap-override: 0%;
  size-adjust: 105.1479%;
  src: local("Arial"), local("ArialMT");
}
```
when I use the font family `Open Sans` and set the last fallback to `sans-serif`.

To get metrics for system fonts, you can hard-code or copy only the metrics files for the system fonts `@capsizecss/metrics`:
- `@capsizecss/metrics/entireMetricsCollection/arial/regular/index.mjs`
- `@capsizecss/metrics/entireMetricsCollection/timesNewRoman/regular/index.mjs`
- `@capsizecss/metrics/entireMetricsCollection/courierNew/regular/index.mjs`

Rather than installing the entire package with the metrics for all fonts.

JusticeMatthew:
Hello hello!

Just trying out the feature, and I noticed that for remote fonts, the prop is called `weights` and it accepts an array with a string like `["100 900"]` 

But for local fonts the prop is called `weight` and it just accepts the string `"100 900"` with no array

Unless there is some super complex reason for this, I feel like they should be consistent

JusticeMatthew:
Also having some issues getting local fonts to work at all

Minimal reproduction https://github.com/JusticeMatthew/astro-font-test

florian-lefebvre:
> Optimized font family overrides for system fonts would have the most impact since they are local and are already installed and are rendered immediately.

@jlarmstrongiv I don't think I understand. If a system font is being passed manually (eg. name `Arial` and fallbacks `['sans-serif']`) I don't think we should generate optimized fallbacks because those fonts are already available on the system. However, I need to check if we handle this kind of case currently

JusticeMatthew:
> 
> - When using the local provider, 1 variant = 1 font face declaration. And a font face declaration can only have 1 weight associated
> 

But the weight can be a string of multiple weights? Or is that being handled under the hood but adding the font face declarations for each one?

jlarmstrongiv:
@florian-lefebvre I’m more referring to how these packages work:

- https://nextjs.org/docs/pages/api-reference/components/font#adjustfontfallback
- https://github.com/unjs/fontaine?tab=readme-ov-file#usage

Often times, they infer a local system fallback font family (`Arial`, `Times New Roman`, or `Courier New`) based on your final css generic fallback family (`sans-serif`, `serif`, `mono`) and generate metrics for loading that local system fallback font family (renamed `Open Sans Fallback: Arial` with css metrics overrides).

That way, you may only specify `Open Sans, sans-serif` and the plugins will take care of the rest—generating a local system fallback font family with metric overrides.

Next.js focuses on fonts that are guaranteed to be installed on all machines.  Specifying Inter as a local fallback and generating overrides is certainly possible, but it’s only effective on computers that already have Inter installed, which is the minority.  Instead, optimizing local system fallback font families should be default because they are already installed, display instantly, and are supported by the majority of devices.

It can be confusing!  Let me know if that makes sense.

florian-lefebvre:
I think I understand correctly and that's what we currently have too! There are improvements in https://github.com/withastro/astro/pull/13635, can you try the preview release?

tanishqmanuja:
Feature Request - Ability to get font hash in js/ts file.

Use Case:
```ts
import type { APIRoute } from "astro";
import { URL as SITE_URL } from "@/site.config";
import raw from "./_theme.css?raw";

const css = raw.replace(
  /__DEPARTURE_MONO_SRC__/g,
  new URL("giscus/fonts/departure-mono-regular.woff2", SITE_URL).href, // this needs to be in sync with astro font api
);

export const GET: APIRoute = () =>
  new Response(css, {
    headers: { "Content-Type": "text/css" },
  });
```

Or a more generalized solution is to provide a way to customize font output location instead of default `_astro/fonts/{HASH}.{FORMAT}`

florian-lefebvre:
You're showing code there but what is your actual usecase?

> Or a more generalized solution is to provide a way to customize font output location instead of default

You can customize [`build.assets`](https://docs.astro.build/en/reference/configuration-reference/#buildassets)

tanishqmanuja:
> You're showing code there but what is your actual usecase?

Code is the use case, there is not way to make this work without extra penalty of duplicating font in public dir
image

+ I am sure CDN's don't like this as well (Cloudflare in my case, internally figures out that the file has same name, same size and other attributes it send 304: NOT MODIFIED). Now it will have to serve new asset every build, due to name change.

 > You can customize [build.assets](https://docs.astro.build/en/reference/configuration-reference/#buildassets)
 
 My idea was to output the font to a `deterministic` location not changing the entire build output dir. 
 
 ```ts
 fonts: [
      {
        provider: "local",
        id: "xyz" // or similar deterministic value
        name: "XYZ",
        cssVariable: "--font-xyz",
        variants: [
          {
            weight: 400,
            style: "normal",
            src: [`xyz.otf`],
          },
        ],
      },
 ]
 ```
 
Should output to ${ASTRO_BUILD_DIR}/fonts/${font.id}.${font.extension}
This allows for:

B) Manual JS Import

```ts
import { URL as SITE_URL } from "@/site.config";
const fontUrl = new URL("_astro/fonts/xyz.otf", SITE_URL).href
```

A) Manual CSS Import

```css
@font-face {
  font-family: "NotXYZ";
  src: url("/_astro/fonts/xyz.otf") format("opentype");
  font-display: swap;
  font-weight: 400;
  font-style: normal;
}
```

florian-lefebvre:
> Code is the use case

Code is not a use case at all, like what problem are you trying to solve? From what you say it seems your concern is about caching right?

The filename is hashed so:
1. It should not change if you didn't change your font settings
2. It should be cached between builds (because it's stored in node modules)
3. It should be cached on the client for users, as long as the cloudflare adapters is setting proper headers for any assets under the cache dir

florian-lefebvre:
Oh and I just saw the end of your message about duplicating fonts in public and in `_astro/fonts`. Because of the processing we do, the fonts file you reference should not be in `public` because they will indeed be duplicated. I'll work on a code + docs warning

tanishqmanuja:
@florian-lefebvre, In the case you want to listen to entire use case .. here it is:
- I am using `giscus` to support comments and reaction in my astro site
- To customize it you need to provide it with a custom css file lets say `theme.css`
- Why a separate css file, giscus renders inside an iframe so it could communicate with the css from my astro site
- To make the giscus theme and site theme look similar i want to use the same font
- To use the same font, we need a fixed url to use inside of the font-face definition

and here we arrive again at the same code ... which is not the use case itself :(
```ts
import type { APIRoute } from "astro";
import { URL as SITE_URL } from "@/site.config";
import raw from "./_theme.css?raw";

const font = "ASTRO_HASH" // PROBLEM HERE

const css = raw.replace(
  /__DEPARTURE_MONO_SRC__/g,
  new URL(`_astro/fonts/${font}.woff2`, SITE_URL).href,
);

export const GET: APIRoute = () =>
  new Response(css, {
    headers: { "Content-Type": "text/css" },
  });
```

florian-lefebvre:
Thanks that helps a lot to have the details! I'm okay with adding an API to get the font url. I'll need to think about its shape because for a given family you can have several font files etc

jonasgeiler:
I would really like to be able to pick only specific variants of my local font to preload. At the moment, my app preloads ALL font files, even those of legacy formats like `woff`, or fonts that are rarely used in my app.

Maybe, when specifying the `src` array, allow setting `preload` in the object, and get rid of the `preload` property on the `<Font>` component:

```ts
import { defineConfig } from "astro/config";

export default defineConfig({
    experimental: {
        fonts: [{
            provider: "local",
            name: "Custom",
            cssVariable: "--font-custom",
            variants: [
                {
                    weight: 400,
                    style: "normal",
                    src: [
                        { url:"./src/assets/fonts/MyFont-regular.woff2", preload: true },
                        { url:"./src/assets/fonts/MyFont-regular.woff" }
                    ]
                },
                {
                    weight: 700,
                    style: "normal",
                    src: [
                        { url:"./src/assets/fonts/MyFont-bold.woff2", preload: true },
                        { url:"./src/assets/fonts/MyFont-bold.woff" }
                    ]
                }
                // ...
            ]
        }]
    }
});
```

```astro
---
import { Font } from 'astro:assets';
---

<Font cssVariable="--font-custom" />
```

Which would only render the following:
```html
<!-- Ignoring the generated file hashes here... -->
<link rel="preload" href="/_astro/fonts/MyFont-regular.woff2" as="font" type="font/woff2" crossorigin="">
<link rel="preload" href="/_astro/fonts/MyFont-bold.woff2" as="font" type="font/woff2" crossorigin="">
```

Instead of:
```html
<!-- Ignoring the generated file hashes here... -->
<link rel="preload" href="/_astro/fonts/MyFont-regular.woff2" as="font" type="font/woff2" crossorigin="">
<link rel="preload" href="/_astro/fonts/MyFont-regular.woff" as="font" type="font/woff" crossorigin="">
<link rel="preload" href="/_astro/fonts/MyFont-bold.woff2" as="font" type="font/woff2" crossorigin="">
<link rel="preload" href="/_astro/fonts/MyFont-bold.woff" as="font" type="font/woff" crossorigin="">
```

This is only my personal suggestion - the exact implementation and API is obviously open for discussion.

florian-lefebvre:
Thanks for the feedback!

> my app preloads ALL font files, even those of legacy formats like woff

This should not be the case. Did you try with the latest version of Astro?

> or fonts that are rarely used in my app.

That's a good point. I quite like what you're suggesting but that would only work for the local provider. I wonder if there's a way to support it for remote providers, or to move it somewhere else? eg.

```astro
---
import { Font } from 'astro:assets';
---

<Font cssVariable="--font-custom" preload={[400,500]} />
```

It's less centralized but it works for all kinds of providers

JusticeMatthew:
> ```astro
> ---
> import { Font } from 'astro:assets';
> ---
> 
> <Font cssVariable="--font-custom" preload={[400,500]} />
> ```

Big fan of this

jonasgeiler:
> @florian-lefebvre wrote: This should not be the case. It was fixed recently, did you try with the latest version of Astro?

Yeah, I am on `astro@5.7.5`. I looked through the source code to look for the fix you mentioned and why it's not working for me - looks like the fix was only implemented for the _**non-local**_ providers.

As you can see this code block is in the `family.provider !== LOCAL_PROVIDER_NAME` branch, but there is no equivalent logic for the `family.provider === LOCAL_PROVIDER_NAME` branch:
https://github.com/withastro/astro/blob/6ed83606f9bdfdf0a35aef1318cf0e2cb4b4236e/packages/astro/src/assets/fonts/load.ts#L151-L163

StrCode:
Hello,
Just took the new fonts feature for a spin, and it's looking fantastic overall!

I did run into an issue when trying to use fonts from Fontshare. There's an issue with the URL for the Fontshare CDN, as it's consistently throwing an error when trying to load a font. You can see the error in the screenshot below:

Screenshot 2025-04-25 at 13 38 06

Minimal reproduction: https://congenial-couscous-4jpj9qg4gp42jv6v.github.dev/

florian-lefebvre:
Hi, thanks for trying it out! Can you open an issue at https://github.com/withastro/astro/issues instead? Thanks!

florian-lefebvre:
If you are using the local provider, I'd like your feedback on https://github.com/withastro/astro/pull/13640!

Pkcarreno:
I have the same request as @tanishqmanuja, my use case is to generate OG images with [Satori](https://github.com/vercel/satori)

What I usually do is to read the `.woff` file from node_modules 
```js
	const fontData = fs.readFileSync(
		'node_modules/@fontsource/ia-writer-duo/files/ia-writer-duo-latin-400-normal.woff'
	)
```

and then integrate it in Satori like this
```js
const raw = await satori(htmlContent as React.ReactNode, {
		width,
		height,
		fonts: [
			{
				name: 'IA Writer Duo',
				data: fontData,
				style: 'normal',
				weight: 400
			}
		]
	})
```

I would appreciate a way to achieve this with this feature and thus not depend on extra installations (as in this case manually install a font from fontsource)

Thanks 🙌

florian-lefebvre:
Hi there! I'm working on a way to retrieve URLs, give it a try and share your experience on the PR: https://github.com/withastro/astro/pull/13811

florian-lefebvre:
> Hi there! I'm working on a way to retrieve URLs, give it a try and share your experience on the PR: [withastro/astro#13811](https://github.com/withastro/astro/pull/13811)

cc @tanishqmanuja and @Pkcarreno

florian-lefebvre:
Hi there! I'm working on a way to have more granular preloads, give it a try and share your experience on the PR: https://github.com/withastro/astro/pull/13826

cc @jonasgeiler

JusticeMatthew:
> I think in builds we should keep them as is as it's better for caching but in dev, we could either do something like `inter-500-italic-<hash>.<ext>` or even `inter-500-italic.<ext>`

Perhaps after what we learned in [#13826](https://github.com/withastro/astro/pull/13826) it might also be good to indicate if the file is a subset? 😅

florian-lefebvre:
@JusticeMatthew see https://github.com/withastro/astro/pull/14279

robinlahtinen:
The experimental Fonts API inserts elements at the beginning of the `<head>` section, which pushes important meta tags such as the charset declaration beyond the first 1024 bytes of the HTML. As a result, Lighthouse reports: "Charset declaration is missing or occurs too late in the HTML."

delucis:
> The experimental Fonts API inserts elements at the beginning of the `<head>` section, which pushes important meta tags such as the charset declaration beyond the first 1024 bytes of the HTML. As a result, Lighthouse reports: "Charset declaration is missing or occurs too late in the HTML."

IIUC the `<Font>` component renders where it is used, so adding it to your pages below important meta tags like that should fix this.

florian-lefebvre:
Yes as Chris said you control the position in the head

RomanHauksson:
Agreed on requiring the `--`. This made it quicker for me to understand what corresponds to what.

florian-lefebvre:
Thanks for the feedback! Interesting API idea, I'll think about it. 2 questions:

1. Can you create a minimal reproduction with the satori stuff so I can look into it?
2. Do you want to send a PR to docs to update the example? File is located at https://github.com/withastro/docs/blob/main/src/content/docs/en/reference/experimental-flags/fonts.mdx

That would be super helpful

plttn:
cutting the Gordian knot here: bunny has both fonts I use as well and grabs the woff variants so I think we're golden! Thanks again!

florian-lefebvre:
Hi @Nickersoft, thanks!

> doesn't seem to be a very good story around variable fonts

The reason is we don't handle them in any special way (or if any, very little)! Handling is the responsability of each provider, in unifont. So you may way want to have a look at the fontsource provider to see how it's handled and how it could be improved: https://github.com/unjs/unifont/blob/main/src/providers/fontsource.ts.

When the feature is stabilize, we could include a section dedicated to variable fonts in docs.

> Inter also offers optical sizing (opsz) for its fonts, but there don't seem to be any way to enable it via the current API.

Currently only the google provider supports this through this option: https://github.com/unjs/unifont/blob/main/src/providers/google.ts#L16. Here is how it's used in practice: https://github.com/shipshapecode/starpod/blob/main/astro.config.mjs#L35-L46.

So if you want support for this when using the fontsource, please open an issue on the unifont repo: https://github.com/unjs/unifont/issues

Issue #1042, opened 2024-10-21
Responsive images


ascorbic:
- Accepted Date: 2024-10-21
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/1031
- Author: @ascorbic
- Champion(s): @ascorbic
- Implementation PR: n/a

# Summary

The current Astro image component offers a lot of flexibility for displaying images. It supports `densities` and `widths` props to help generate the correct `img` attributes, and the default image service supports modern formats such as AVIF and WebP. While this gives users the tools to create performant and responsive images, it does not give guidance in how to use them - and requires that they are set on all images. This proposal is for a more opinionated image component. It would offer all of the tools from the current component, and also introduce new props and config options that follow best practices by default.

# Background & Motivation

Displaying images on the web is difficult, even for the most experienced developers. Users suffer slower page loads, and poor experience as the page layout jumps around. Meanwhile sites experience poor core web vitals scores for performance, cumulative layout shift (CLS) and largest contentful paint (LCP).

The most common [`img`tag attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attributes) are well known: `src`, `alt`, `width` and `height`, there are several lesser-known attributes that are needed if an image is to have the best performance. All of these are optional according to the spec, but best practices require most of them. The most important are `srcset`, `sizes`, `loading`, `decoding` and `fetchpriority`.

These are a lot of attributes to remember and understand, though the final three have values that are usually safe to think of as dependent on just whether the image is onscreen when the page loads. Astro Image already sets `loading` and `decoding` to `lazy` and `async` by default. However `srcset` and `sizes` have no simple rules because they depend on how the image will be displayed, and can be very hard to do correctly. Images also need to be styled correctly if they are to be responsive and avoid CLS.

This proposal is inspired by the attributes generated by `@unpic/astro`, which I created, but with some changes to make it closer to the existing component behavior, and less focussed on image CDNs.

# Goals

- A single `layout` prop that, when combined with existing props, sets the attributes that will make an image responsive and following best practices.
- Config options to change the defaults for all images
- Backwards-compatible, so that existing images are unaffected unless they set the props or config options.
- Add support for optional cropping in image services

# Non-goals

- Placeholder support
- Automatic provider detection
- Art direction
- Implementing crop support for all existing image services

# Example

Responsive images will be enabled by setting the `layout` prop to `responsive`, `fixed` or `full-width`.

```astro
---
import { Image } from "astro:assets"
[...]

PR #1051, opened 2024-11-04
Responsive images

ascorbic:
# Summary

Implements opinionated best practices to Astro Image, generating srcset, sizes and styles automatically.

```astro
---
import { Image } from "astro:assets"
import rocket from "./rocket.jpg"
---
<Image src={rocket} width={800} height={600} layout="responsive" />
```

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/responsive-images/proposals/0053-responsive-images.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/1042)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/1031)

### DIFF
--- a/proposals/0053-responsive-images.md
+++ b/proposals/0053-responsive-images.md
@@ -1,0 +1,273 @@
+<!--
+  Note: You are probably looking for `stage-1--discussion-template.md`!
+  This template is reserved for anyone championing an already-approved proposal.
+
+  Community members who would like to propose an idea or feature should begin
+  by creating a GitHub Discussion. See the repo README.md for more info.
+
+  To use this template: create a new, empty file in the repo under `proposals/${ID}.md`.
+  Replace `${ID}` with the official accepted proposal ID, found in the GitHub Issue
+  of the accepted proposal.
+-->
+
+**If you have feedback and the feature is released as experimental, please leave it on the Stage 3 PR. Otherwise, comment on the Stage 2 issue (links below).**
+
+- Start Date: 2024-11-04
+- Reference Issues: <!-- related issues, otherwise leave empty -->
+- Implementation PR: <!-- leave empty -->
+- Stage 2 Issue: https://github.com/withastro/roadmap/issues/1042
+- Stage 3 PR: <!-- related roadmap PR, leave it empty if you don't have a PR yet -->
+
+# Summary
+
+The current Astro image component offers a lot of flexibility for displaying images. It supports `densities` and `widths` props to help generate the correct `img` attributes, and the default image service supports modern formats such as AVIF and WebP. While this gives users the tools to create performant and responsive images, it does not give guidance in how to use them - and requires that they are set on all images. This proposal is for a more opinionated image component. It would offer all of the tools from the current component, and also introduce new props and config options that follow best practices by default.
+
+# Example
+
+Responsive images will be enabled by setting the `layout` prop to `responsive`, `fixed` or `full-width`.
+
+```astro
+---
+import { Image } from "astro:assets"
+import rocket from "./rocket.jpg"
+---
+<Image src={rocket} width={800} height={600} layout="responsive" />
+```
+
+A new `layout` option for the `image` config will default all images to that layout. This can be overridden on each image.
+
+```js
+import { defineConfig } from "astro/config";
+
+export default defineConfig({
+  image: {
+    layout: "responsive",
+  },
+});
+```
+
+# Background & Motivation
+
+Displaying images on the web is difficult, even for the most experienced developers. Users suffer slower page loads, and poor experience as the page layout jumps around. Meanwhile sites experience poor core web vitals scores for performance, cumulative layout shift (CLS) and largest contentful paint (LCP).
+
+The most common [`img`tag attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attributes) are well known: `src`, `alt`, `width` and `height`, there are several lesser-known attributes that are needed if an image is to have the best performance. All of these are optional according to the spec, but best practices require most of them. The most important are `srcset`, `sizes`, `loading`, `decoding` and `fetchpriority`.
+
+These are a lot of attributes to remember and understand, though the final three have values that are usually safe to think of as dependent on just whether the image is onscreen when the page loads. Astro Image already sets `loading` and `decoding` to `lazy` and `async` by default. However `srcset` and `sizes` have no simple rules because they depend on how the image will be displayed, and can be very hard to do correctly. Images also need to be styled correctly if they are to be responsive and avoid CLS.
+
+# Goals
+
+- A single `layout` prop that, when combined with existing props, sets the attributes that will make an image responsive and following best practices.
+- Config options to change the defaults for all images
+- Backwards-compatible, so that existing images are unaffected unless they set the props or config options.
+- Add support for optional cropping in image services
+- Images displayed correctly, even if image service doesn't support cropping
+
+# Non-goals
+
+- Placeholder support
+- Automatic provider detection
+- Art direction
+- Implementing crop support for all existing image services
+
+# Detailed Design
+
+When a user sets the layout, either via the prop or as a default in the config, the image component will auto-generate defaults for `srcset` and `sizes`. It will also apply styles to determine the resizing behavior. This will be done according to these rules for each layout.
+
+For each of these, `<width>` is a placeholder for the value passed by the user as the image `width` prop. The **widths** value refers to the widths of the images in the `srcset`. The **sizes** value is the string value set as the `sizes` attribute.
+
+### Image `srcset`
+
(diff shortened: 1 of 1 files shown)


proposals/0053-responsive-images.md
```diff
@@ -0,0 +1,273 @@
...
+
+For this reason, we will generate a srcset with width conditions, based on the width and layout props.
+
+These are indicative implementations of the functions that generate the image widths for the `srcset`. The `breakpoints` argument is an array of possible screen resolutions, which it uses to choose candidate breakpoints. The default list is shown in the section below. In the real implementation, widths that are larger than the source image would be filtered out.
```
Princesseuh, line 84: > In the real implementation, widths that are larger than the source image would be filtered out.

How does this work for remote images without `inferSize`?

ascorbic: It would probably rely on image service options to prevent upscaling. It might request a larger size but it would return the original dimensions.

ascorbic: ...but for the built-in service I think we'd just not include super large sizes at all by default

ascorbic:
I've updated the implementation details part of the RFC based on things I've discovered while prototyping.

carlcs:
Hey! I‘m excited to see work being picked up again on the image components. I read the rfc and don’t really understand how values for the sizes attribute are created when „responsive“ layout is used. Is there JS being injected or how does the component know how large the image is show at different breakpoints?

ascorbic:
@carlcs It generates a sizes attribute based on the assumption that it's the full width of the screen when downsized. You'd need to pass your own if this is incorrect.

carlcs:
@ascorbic that‘s perfect if we can still set sizes manually! You might want to change docs a bit because this part is a bit misleading.

```
		 * The following `<Image />` component properties should not be used with responsive images as these are automatically generated:
		 * 
		 * - `densities`
		 * - `widths`
		 * - `sizes` 		 
```

flavianh:
Hi @ascorbic , I really love this feature! Is there a timeline for it to go from experimental to stable?

jasikpark:
It would be great to enable support for:

```css
img {
position: absolute;
inset: 0;
width: 100%;
height: 100%;
}
```

I've got a background image that I'm using as a wallpaper, I just want it to act like `layout="full-width"`, except I want to override the height as `100%`. `height: calc(100% * 1px);` doesn't work I don't think though...

ascorbic:
@jasikpark you should be able to do that now. You need to use a class rather than `img`, because the specificity is too low. I just tried this and it works:

```astro
---
import { Image } from "astro:assets";
import penguin from "../assets/penguin.jpg";
---
<Image src={penguin} alt="a penguin" layout="full-width" class="cover"/>
<style>
  .cover {
    height: 100%;
    position: absolute;
    inset: 0;
  }
</style>
```

yinhx3:
https://github.com/withastro/astro/issues/13666

I encountered a specificity issue when using Tailwind with responsive images. Changing the selector doesn’t resolve the problem because Tailwind’s @layer lowers its specificity.

I believe the styles applied by responsive images shouldn't exist. The sizes, width, and height attributes in HTML already provide a solid fallback for CSS and shouldn’t interfere with styling. The way responsive images impose styles contradicts the original intent of these attributes.

I previously saw Astro as a framework that stays close to native behavior and maintains predictability, which is why I chose it. However, its responsive image implementation feels overly complex, and I don’t find it appealing.

ascorbic:
@yinhx3 thanks for your feedback. One of the main reasons we have the experimental features and RFC process is to get feedback on APIs and implementations. I was planning to simplify these styles and lower the specificity. I was originally going to use `:where()`, but your comment prompted me to look into using `@layer` which looks like it will work here, and makes it play better with Tailwind 4.

connor-baer:
[Cascade layers aren't supported](https://caniuse.com/css-cascade-layers) in many of the browser versions that [Astro currently supports](https://docs.astro.build/en/install-and-setup/#browser-compatibility). Is there a way to use them in a backward-compatible way?

ascorbic:
@yinhx3 the `astro.config.mjs` options are less about the styling than about controlling the auto-generated srcsets and sizes. We may end up removing the styles, but what I don't want is to end up requiring everyone to have to add styles in order to avoid setting some defaults.

yinhx3:
@ascorbic  I agree with your perspective. Generating images of different sizes and populating the `srcset`, `sizes`, `width`, and `height` attributes is precisely what an SSG/SSR framework should handle. Global defaults for `object-fit`, `object-position`, and `width` can be set via CSS. Astro can have an opinionated CSS reset or leave it unset, allowing users to choose their own CSS reset—both approaches make sense.

ascorbic:
I am considering changing the name of the layout option `responsive`. Currently the feature is called "responsive images", with three options for `layout`: `fixed`, `full-width` and `responsive`. These are used to define which srcset and sizes to generate, as well as the default styles. It's become clear that having one of the layouts called `"responsive"` is confusing, and people are choosing it when they need `full-width`. I'm thinking that it should be renamed to `constrained`, which better reflects what it actually is for: images that are displayed at a maximum of their original size, but will downsize if the display is too narrow.

ascorbic:
This has now been updated with the changes to the layout name and simplified styles, included in https://github.com/withastro/astro/pull/13677

hyperknot:
I did some digging into responsive images, it works really well in the current version!

I opened a bug report related to duplicate / identical files being generated with different hash https://github.com/withastro/astro/issues/13819

And I'd like to propose to please consider the use case of Retina screenshots in .md files. The use case I believe is very common: some images in .md files are 1x resolution, some are 2x. How do you handle these correctly? (Without breaking into MDX).

I mean some kind of pattern needs to be used in the .md file, either filenames like `image_2x.png`, or `image.png?r=2x` or `image.png#2x`. I prefer `_2x.png` like file names.

I'm migrating from Ghost, where I made a client side JS to recognise this pattern and set the size correctly.

But with Astro's responsive images, I cannot do this, it needs to be signaled to the responsive image pipeline _before_ rendering. Can you recommend a way to do this? My only idea is to make a Remark plugin, but it's not clear how to do this. Is there any simpler way?

My proposal is basically to auto-recognise `_2x.ext` like file names and automatically handle them as Retina.

hyperknot:
Also, from the linked Gatsby page, they only use screen size based brake-points for full-width images, which makes a lot of sense. For non-full-width images, why are we doing any kind of resizing based on common screen sizes? Our "constrained" image size has nothing to do with the screen-size.

`[0.25, 0.5, 1, 2]` densities make a lot more sense. 

image

ascorbic:
@hyperknot 
I'm not keen on magic like trying to recognise filename patterns. I'd much rather there was an explicit way to pass args in markdown. 

Re the second point, it's a long time since I built it but IIRC by default Gatsby does the same as us for constrained images: it includes 1x and 2x, plus screen widths that are smaller than the 1x size. This is to allow for scaling-down on smaller screens. The difference with fullWidth is that it *only* uses the screen widths.

hyperknot:
@ascorbic You are right that it should not hardcoded filename patters. Yesterday, I dig into it and realized a very minimal, simple rehype plugin can do this task properly, no hard coding needed. So users themselves can set up whatever logic they want, they can use query strings, hash or even alt text hints.

It's as simple as this:

```
import { visit } from 'unist-util-visit'

export function rehype2xImages() {
  return (tree: any) => {
    visit(tree, 'element', (node) => {
      if (node.tagName === 'img' && node.properties.src && node.properties.src.includes('_2x.')) {
        node.properties['data-retina2x'] = true
      }
    })
    return tree
  }
}
```

Now the part which I could not figure out is how to integrate it into the rest of the pipeline. I guess outside of forking Astro, it's not possible today?

What I did was the same client side JS hack, like I used on Ghost:

```
<script>
document.querySelectorAll('img[data-retina2x="true"]').forEach(img => {
  const width = parseInt(img.getAttribute('width'));
  img.style.maxWidth = (width / 2) + 'px';
});
</script>
```

This works, but give a flash + downloads the wrong image + probably Google also doesn't like it. So it'd be really nice to have this integrated in the image pipeline.

About the magic string "data-retina2x", of course that's just a string for my use case. A better one would be "dpr" for Device Pixel Ratio. Or devicePixelRatio and then we are consistent with the [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio).

So my proposal is the following:
1. Users come up with whatever logic they want, and in a rehyde plugin they set 
```
node.properties['devicePixelRatio'] = 2.0 // or 3.0 etc.
```

2. The image pipeline recongises properties.devicePixelRatio and calculates the target width accordingly.

I think this way it'd be universal, yet very user-friendly, only requiring a tiny rehyde plugin.

hyperknot:
@ascorbic I'd be happy to start contributing to this PR. What is the best way to connect? Discord #contribute channel?

ascorbic:
@hyperknot we don't usually have more than one contributor to an RFC PR. People can give feedback, but the author of the PR is the one who would integrate any changes. I'm happy to consider specific proposals and suggestions.

hyperknot:
OK, my proposal is as simple as accepting node.properties.devicePixelRatio on Markdown images and <Image devicePixelRatio={}> on Image tags, based on what I wrote before.

matthewp (approved):
Looks great

PR #1109, opened 2025-01-24
fix typo on import loadRenderers instead of laodRenderers

Scot3004:
I created this PR because I did not find other way to do the inline suggestion on existing rfc PR

Also I feel that the discussion was already made pior to my change that is just a fix for a typo in text
image

# Summary

The rfc solves the container api

# Links

- [Rendered RFC](https://github.com/Scot3004/roadmap/blob/3dac369cdfc6e41dee2909b9920c2a85b3b76483/proposals/0048-container-api.md)


ematipico:
@Scot3004 

An inline suggestion is just a comment, you hit a particular button and then update the highlighted code. The snapshot should help:

Screenshot 2025-01-24 at 15 45 38

Issue #1151, opened 2025-04-04
Live content loaders


ascorbic:
- Accepted Date: 2025-04-04
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/1137
- Author: @ascorbic 
- Champion(s): @ascorbic 
- Implementation PR: 

# Summary

Adds support for live data to content collections. Defines a new type of content loader that fetches data at runtime rather than build time, allowing users to get the data with a similar API.

# Background & Motivation

In Astro 5, the content layer API added support for adding diverse content sources to content collections. Users can create loaders that fetch data from any source at build time, and then access it inside a page via `getEntry` and `getCollection`. The data is cached between builds, giving fast access and updates. However there is no method for updating the data store between builds, meaning any updates to the data need a full site deploy, even if the pages are rendered on-demand. This means that content collections are not suitable for pages that update frequently. Instead, today these pages tend to access the APIs directly in the frontmatter. This works, but means users don't benefit from the simple, unified API that content loaders offer. In most cases users tend to individually create loader libraries that they share between pages.

This proposal introduces a new kind of loader that fetches data from an API at runtime, rather than build time. As with other content loaders, these loaders abstract the loading logic, meaning users don't need to understand the details of how data is loaded. These loaders can be distributed as node modules, or injected by integrations. 

# Goals

- a new type of **live content loader** that is executed at runtime
- integration with `getEntry` and `getCollection`, allowing developers to use **a familiar, common API** to fetch data
- optional **integration with [route caching](https://github.com/withastro/roadmap/discussions/1131)**, allowing loaders to define cache tags, TTL etc associated with the data, which the user could decide to use
- loader-specific **query options**

# Non-goals

- server-side caching of the data. Instead it would integrate with the route cache and HTTP caches to cache the full page response
- updating the content layer data store. Live loaders return data directly and do not update the store
- support for existing loaders. They will have a different API. Developers could in theory use shared logic, but the loader API will be different

# Example

It may need to be a different content config file, depending on what compiler magic we can do.

```ts
// src/livecontent.config.ts
import { defineLiveCollection } from "astro:content";
// Possibly a different name, or maybe use the same `defineCollection`

import { storeLoader } from "@mystore/astro-loader";


const products = defineLiveCollection({
    loader: storeLoader({ field: "products", key: process.env.STORE_KEY })
})

export const collections = { products };

```

Define loaders in the site or in a shared library.

```ts
// storeloader.ts

[...]

PR #1164, opened 2025-05-03
Live content collections

ascorbic:
# Summary

Adds support for live data to content collections. Defines a new type of content loader that fetches data at runtime rather than build time, allowing users to get the data with a similar API.

```ts
import { defineCollection } from "astro:content";

const products = defineCollection({
  type: "live",
  loader: {
    name: "store-loader",
    loadCollection: async ({ filter }) => {
      // ...
      return {
        entries: products.map((product) => ({
          id: product.id,
          data: product,
        })),
      };
    },
    loadEntry: async ({ filter }) => {
      // ...
      return {
        id: filter.id,
        data: product,
      };
    },
  };
});
export const collections = { products };
```

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/main/proposals/0055-live-content-loaders.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/1151)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/1137)

### DIFF
--- a/proposals/0055-live-content-loaders.md
+++ b/proposals/0055-live-content-loaders.md
@@ -21,7 +21,7 @@
 
 # Summary
 
-Adds support for live data to content collections. Defines a new type of content loader that fetches data at runtime rather than build time, allowing users to get the data with a similar API.
+Adds support for live data to content collections. Defines a new type of content loader that fetches data at runtime rather than build time, allowing users to get the data with dedicated functions that make the runtime behavior explicit.
 
 # Example
 
@@ -47,33 +47,57 @@ export function storeLoader({
     name: "store-loader",
     loadCollection: async ({ logger, filter }) => {
       logger.info(`Loading collection from ${field}`);
-      // load from API
-      const products = await loadStoreData({ field, key, filter });
-      const entries = products.map((product) => ({
-        id: product.id,
-        data: product,
-      }));
-      return {
-        entries,
-      };
+      try {
+        // load from API
+        const products = await loadStoreData({ field, key, filter });
+        const entries = products.map((product) => ({
+          id: product.id,
+          data: product,
+        }));
+        return {
+          entries,
+        };
+      } catch (error) {
+        logger.error(`Failed to load collection: ${error.message}`);
+        return {
+          error: {
+            message: `Failed to load products: ${error.message}`,
+            code: "COLLECTION_LOAD_ERROR",
+          },
+        };
+      }
     },
     loadEntry: async ({ logger, filter }) => {
       logger.info(`Loading entry from ${field}`);
-      // load from API
-      const product = await loadStoreData({
-        field,
-        key,
-        filter,
-      });
+      try {
+        // load from API
+        const product = await loadStoreData({
+          field,
+          key,
+          filter,
+        });
 
-      if (!product) {
-        logger.error(`Product not found`);
-        return;
+        if (!product) {
+          return {
+            error: {
+              message: `Product not found`,
+              code: "ENTRY_NOT_FOUND",
+            },
+          };
+        }
+        return {
+          id: filter.id,
+          data: product,
+        };
+      } catch (error) {
+        logger.error(`Failed to load entry: ${error.message}`);
+        return {
+          error: {
+            message: `Failed to load product: ${error.message}`,
+            code: "ENTRY_LOAD_ERROR",
+          },
+        };
       }
-      return {
-        id: filter.id,
-        data: product,
-      };
     },
   };
 }
@@ -95,23 +119,30 @@ const products = defineCollection({
 export const collections = { products };
 ```
 
-The loader can be used in the same way as a normal content collection:
+The loader is accessed using new dedicated functions that make the runtime behavior explicit:
 
 ```astro
 ---
-import { getCollection, getEntry } from "astro:content";
+import { getLiveCollection, getLiveEntry } from "astro:content";
 
-// Get all entries in a collection, like other collections
-const allProducts = await getCollection("products");
+// Get all entries in a collection
+const { data: allProducts, error } = await getLiveCollection("products");
+if (error) {
+  // Handle error gracefully
+  return Astro.redirect('/500');
+}
 
 // Live collections optionally allow extra filters to be passed in, defined by the loader
-const clothes = await getCollection("products", { category: "clothes" });
+const { data: clothes } = await getLiveCollection("products", { category: "clothes" });
 
-// Get entrey by ID like other collections
-const productById = await getEntry("products", Astro.params.id);
+// Get entry by ID
+const { data: productById } = await getLiveEntry("products", Astro.params.id);
 
 // Query a single entry using the object syntax
-const productBySlug = await getEntry("products", { slug: Astro.params.slug });
+const { data: productBySlug, error: slugError } = await getLiveEntry("products", { slug: Astro.params.slug });
+if (slugError) {
+  return Astro.redirect('/404');
+}
 ---
 ```
 
@@ -123,14 +154,17 @@ This means that content collections are not suitable for pages that update frequ
 
 This proposal introduces a new kind of loader that fetches data from an API at runtime, rather than build time. As with other content loaders, these loaders abstract the loading logic, meaning users don't need to understand the details of how data is loaded. These loaders can be distributed as node modules, or injected by integrations.
 
+The API uses dedicated functions (`getLiveCollection` and `getLiveEntry`) to make it explicit that these operations perform network requests at runtime, helping developers understand the performance implications and error handling requirements.
+
 # Goals
 
 - a new type of **live content loader** that is executed at runtime
-- integration with user-facing `getEntry` and `getCollection` functions, allowing developers to use **a familiar, common API** to fetch data
+- dedicated user-facing functions `getLiveEntry` and `getLiveCollection` that make the **runtime behavior explicit**
+- **built-in error handling** with consistent error response format
 - loader-specific **query and filters**, which a loader can define and pass to the API
 - **type-safe** data and query options, defined by the loader as generic types
-- support for user-defined **Zod schemas**, executed at runtime, to validate or transform the data returned by the loader.
-- support for runtime **markdown rendering**, using a helper function provided in the loader context.
+- support for user-defined **Zod schemas**, executed at runtime, to validate or transform the data returned by the loader
+- support for runtime **markdown rendering**, using a helper function provided in the loader context
 - optional **integration with [route caching](https://github.com/withastro/roadmap/issues/1140)**, allowing loaders to define cache tags and expiry times associated with the data which are then available to the user
 
 # Non-Goals
@@ -144,13 +178,51 @@ This proposal introduces a new kind of loader that fetches data from an API at r
 
 # Detailed Design
 
-While the user-facing API is similar to the existing content loaders, the implementation is significantly different.
+The user-facing API uses dedicated functions `getLiveCollection` and `getLiveEntry` to make it clear these are runtime operations that may fail and have different performance characteristics than regular content collections.
+
+## User-facing API
+
+The new functions are exported from `astro:content` alongside the existing functions:
+
+```ts
+import {
+  getCollection,
+  getEntry,
+  getLiveCollection,
+  getLiveEntry,
+} from "astro:content";
+```
+
+These functions return a result object with either `data` or `error`, making error handling explicit and consistent:
+
+```ts
+// Success case
+const { data, error } = await getLiveCollection("products");
+if (error) {
+  // Handle error
+  console.error(error.message);
+  return Astro.redirect("/error");
+}
+// Use data safely
+data.forEach((product) => {
+  // ...
+});
+
+// With filters
+const { data: electronics } = await getLiveCollection("products", {
+  category: "electronics",
+});
+
+// Single entry
+const { data: product, error } = await getLiveEntry(
+  "products",
+  Astro.params.id
+);
+```
 
 ## Loader API
 
-A live loader is an object with two methods: `loadCollection` and `loadEntry`. For libraries that distribute a loader, the convention for these will be for users to call a function that returns a loader object, which is then passed to the `defineCollection` function. This allows the user to pass in any configuration options they need. The loader object is then passed to the `defineCollection` function.
-
-The `loadCollection` and `loadEntry` methods are called when the user calls `getCollection` or `getEntry`. They return the requested data from the function, unlike existing loaders which are responsible for storing the data in the content layer data store.
+A live loader is an object with two methods: `loadCollection` and `loadEntry`. These methods should handle errors gracefully and return either data or an error object:
 
 ```ts
 // storeloader.ts
@@ -159,20 +231,48 @@ export function storeLoader({ field, key }): LiveLoader {
   return {
     name: "store-loader",
     loadCollection: async ({ filter }) => {
-      // ...
-      return {
-        entries: products.map((product) => ({
-          id: product.id,
-          data: product,
-        })),
-      };
+      try {
+        const products = await fetchProducts(filter);
+        return {
+          entries: products.map((product) => ({
+            id: product.id,
+            data: product,
+          })),
+        };
+      } catch (error) {
+        return {
+          error: {
+            message: `Failed to load products: ${error.message}`,
+            code: "COLLECTION_LOAD_ERROR",
+            cause: error,
+          },
+        };
+      }
     },
     loadEntry: async ({ filter }) => {
-      // ...
-      return {
-        id: filter.id,
-        data: product,
-      };
+      try {
+        const product = await fetchProduct(filter);
+        if (!product) {
+          return {
+            error: {
+              message: "Product not found",
+              code: "ENTRY_NOT_FOUND",
+            },
+          };
+        }
+        return {
+          id: filter.id,
+          data: product,
+        };
+      } catch (error) {
+        return {
+          error: {
+            message: `Failed to load product: ${error.message}`,
+            code: "ENTRY_LOAD_ERROR",
+            cause: error,
+          },
+        };
+      }
     },
   };
 }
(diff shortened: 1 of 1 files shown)


ematipico:
I noticed that the RFC doesn't cover error handling. Can we cover that part? Things that could wrong:
- timeout while fetching a collection
- invalid data
- parsing error of data
- more?

Do live collections provide a mean to handle the error gracefully? If so how? If not, how users can mitigate the error? Looking forward to see this part covered by the RFC

sarah11918:
I know everyone's super excited for real-time, updating data! 🎉  Just some thoughts from me that came to mind while reading!

You've mentioned user confusion between "similar but different" APIs in the drawbacks, but I worry about using *identically-named* helper functions that do (and take? and return?) different things.

`getEntry()` and `getCollection()` are familiar to existing content collections users. (But, not familiar to someone who hasn't used existing collections!) But with different implementations under the hood, I wonder whether the similarity is actually an advantage, or just a greater chance for confusion? (Assuming I'm understanding this properly.)

From a docs/support standpoint:

> "What arguments does `getCollection()` take?" "Oh, depends which kind of collection you're querying. It can take a filtering function, or a query object." 

> "What does `getCollection()` return?" "Oh, it depends. There might be a `cacheHint` object included."

At that point, I'd probably prefer to document a `getLiveEntry()` and `getLiveCollection()` and then you can do whatever you want with those that make sense for their specific context, and not have to shoe horn them into existing functions. 😄 

I could see keeping identical naming for e.g. people being able to keep their existing querying code when swapping out an existing collection for a live content loader collection. But it doesn't seem like this is a smooth (or even envisioned?) path anyway? They aren't even configured in the same file, so treating these as a completely new and different thing seems reasonable?

I'm also assuming that creating new functions probably also means we run less risk of introducing unexpected behaviour in existing projects by changing `getCollection()` and `getEntry()` to now be "smart" and know which kind of collection they are querying?

(Feel free to ignore the rest, but I would be remiss if I didn't at least bring it up...)

There's even a world where "Content Collections" remain "at build/request time" and this is an entirely new paradigm (just to open up the world of possibilities)! These functions feel like an example of being superimposed onto our existing collections where maybe they want to be something different. And if you were allowed to operate "outside the collections/loader box", maybe there are more cases where you'd feel the freedom of flexibility? (e.g. trying to follow the established pattern of a config file for something that isn't really a config...) Maybe it's not even a "loader", but a Firehose!

(Would it also be easier to launch/promote/write about? Easier for content creators not to have to update their existing "content collections" material yet again, but instead get to talk about a "new thing" now: real-time, live, updating data?  Avoiding the need to *further* distinguish between "which version of content collections are you using?")

(Again, assuming I'm understanding everything! This just feels like some pretty significant differences that are, as you said, close, but not exactly the same. to some existing things. And from a docs / support standpoint, these are always the most challenging things to handle!)

ascorbic:
I've updated the RFC with two changes based on @ematipico and @sarah11918's feedback:

- change to use separate exports: `getLiveCollection` and `getLiveEntry`
- add explicit error handling. It now returns an object with `{ data, error }` instead of directly returning the data or throwing

stipsan:
>- support for user-defined **Zod schemas**, executed at runtime, to validate or transform the data returned by the loader

Is it possible for an Astro Integration to define generated schemas and plug it into the astro typegen pipeline?

For Sanity we're interested in providing a loader creator that can hook into our [TypeGen pipeline](https://www.sanity.io/docs/apis-and-sdks/sanity-typegen#k47494ab9257c) in order to allow a minimal API that is fully typed:
```ts
// src/sanity.types.ts (generated by TypeGen, simplified here)
export type Product = {
  _id: string
  _type: 'product'
  slug: string
  title: string
  category: string[]
}


// src/live.config.ts
import {defineLiveCollection, type ExtractFilterFromType} from '@sanity/astro'

const products = defineLiveCollection({type: 'product'});

export const collections = { products };
```

With the inference we can type it so that TS would throw on a `type` that doesn't exist in the Sanity Schema:
```ts
const products = defineLiveCollection({type: 'produc'});
                                             // ^? 'produc' doesn't exist. Did you mean 'product'?
```
A valid `type` lets us infer the return type of `defineLiveCollection({type: 'product'})` to be:
```ts
// somehow `@sanity/astro` feeds Astro this typegen
import {type InferData, type InferEntryFilter, type InferCollectionFilter} from '@sanity/astro'
import type { LiveLoader } from "astro/loaders";
import type {Product} from './sanity.types'

type ProductData = InferData<Product>
//   ^? {_id:string; _type:'product'; slug: string; title: string; category: string[]}
type ProductEntryFilter = InferEntryFilter<Product>
//   ^? {slug: string; title: string; category: string[]}
type ProductCollectionFilter = InferCollectionFilter<Product>
//   ^? {_id: string | string[]; slug: string | string[]; title: string | string[]; category: string[]}

interface SanityLoaders {
  'product': LiveLoader<ProductData, ProductEntryFilter, ProductCollectionFilter>
}

export function defineLiveCollection<DocumentType extends keyof SanityLoaders>({type: DocumentType}): SanityLoaders[DocumentType]
```


And all the usages sites would be fully typed:
```ts
const { entries: allProducts } = await getLiveCollection("products");
//               ^? {_id:string; _type:'product'; slug: string; title: string; category: string[]}

const { entries: clothes } = await getLiveCollection("products", {
  categories: ["clothes"]
  // ^? `categories` doesn't exist on `ProductCollectionFilter`, did you mean `category`?
});

const { entry: productById } = await getLiveEntry("products", Astro.params.id);
//             ^? {_id:string; _type:'product'; slug: string; title: string; category: string[]}
const { entry: productBySlug } = await getLiveEntry("products", { id: Astro.params.slug });
//                                                              ^? `id` doesn't exist on `ProductEntryFilter`, did you mean `slug`?
```

stipsan:
@ascorbic Thanks! I’ll check out that helper and give it a spin in a POC 🙌

> You can presumably trust the Sanity API to return the correct data?

Yes and no. We don’t yet enforce server-side schema validation on writes to Content Lake; it’s currently handled client-side by Sanity Studio.

However, Sanity Studio isn’t the only entity that can write to Content Lake. Content Lake accepts any valid JSON document. Users frequently automate content imports from external systems (Shopify, Salesforce, Mux, etc.) directly into Content Lake.

The only strict guarantee Content Lake provides is compliance with its query language rules ([GROQ](https://groq.dev/)). For example, if I query:
```sh
*[_type == "post"]{_type, _id, title}
```

We know upfront, per GROQ rules, this query returns either an array of objects with keys _type, _id, and title, or an empty array. _id and _type are special system properties (_type must be a string, and _id must be globally unique). From the filter, we know _type is 'post'. But title could be any valid JSON. At runtime, the typesafe declaration becomes:
```ts
type Json =
  | string
  | number
  | boolean
  | null
  | { [key: string]: Json | undefined }
  | Json[];

type QueryResult = {
  _id: string;
  _type: 'post';
  title: Json;
}[];
```

The TypeGen can analyze your sanity.config.ts from Sanity Studio to narrow down types:

```ts
type QueryResult = {
  _id: string;
  _type: 'post';
  title: string | null;
}[];
```

However, since this isn’t enforced, parser codegen becomes beneficial.

It seems achievable by using Astro’s [createCodegenDir](https://docs.astro.build/en/reference/integrations-reference/#createcodegendir) alongside generating Zod parsers when calling `injectTypes`. I’ll explore implementing Astro’s [Content Loader API](https://docs.astro.build/en/reference/content-loader-reference/) to verify this.

I imagine it’d be helpful for integration authors to have a dedicated API for schema codegen, similar to `injectTypes`, generating Zod schemas usable by both build-time and live loaders. This is particularly beneficial when dealing with live content previews during publishing workflows.

By default, our TypeGen creates types reflective of draft document states. Drafts may be incomplete or temporarily invalid—like images awaiting captions. Hence, customization options for userland are valuable. Users might be fine with handling `string | nul`l, or they might prefer coercion via parsers:

```ts
// src/sanity.types.ts (simplified example reflecting live preview states)
export type Product = {
  _id: string;
  _type: 'product';
  slug: string | null;
  title: string | null;
  category: string[] | null;
};

import { defineLiveCollection } from '@sanity/astro';

const products = defineLiveCollection({
  type: 'product',
  schema: (schema) => schema.extend({
    title: schema.shape.title.unwrap().default('Untitled'),
    category: z.preprocess((val) => (Array.isArray(val) ? val : []), schema.shape.category.unwrap()),
  }),
});

export const collections = { products };
```

We frequently see this challenge with live preview-capable applications, especially as the studio schema evolves over time.

boutell:
I'm coming at this from a slightly different perspective. For ApostropheCMS, we decided our first priority was to make on-page editing possible within Astro.

To do that, we decided to invert the control flow. While Astro is of course the front end, in a combined ApostropheCMS / Astro project there is a `[...slug].astro` route that does a lot of the lifting. That Astro route makes a call to ApostropheCMS, which responds with the information Astro needs to render that page.

And even after Astro starts rendering a page template, it still often needs to insert user-edited content at a particular point, in a way that maintains editability. So we do that by providing an `ApostropheArea` astro component, which in turn invokes various widget-specific Astro components to render different types of content widgets.

This allows ApostropheCMS development to proceed as normal, but with Astro as the rendering engine, mapping the concepts of our CMS one to one to various folders of Astro templates that produce 100% of the actual markup.

Since the CMS must know and manage the structure of all of the data, it makes sense to make those representations in one place (e.g. the CMS is the model layer), and then let Astro concern itself exclusively with presentation (Astro is the view layer).

So at least for right now, we probably wouldn't use live content loaders very much.

However, it could make sense for us to support them in the future, particularly if we find ourselves talking to customers who are less interested in ApostropheCMS "driving the bus" as it were, and more interested in static builds with the addition of *some* dynamic content from the CMS, managed via the CMS back end.

This isn't a criticism of the new API, which looks well-suited to its purpose, as long as error handling is taken into account. I bring all this up just to create awareness of different perspectives and use cases that might not map to it as expected.

Alynva:
I've read the announcement and the docs, but haven't tested it yet. Is this feature exclusive to SSR? I always find it confusing that the docs don't explicitly state which rendering methods a feature is compatible with. I mean, after understanding what the feature is and what its capabilities are, it's not hard to guess. But it would be way better if it was just stated at the start of the docs page, kinda like the "added in" field...

ascorbic:
Having spent some time working on building loaders, I think it would be good to add `lastModified` to `cacheHint`, and possibly remove `maxAge` because that's not something that loaders can normally decide.

Fakerko:
Hi, based on a thread on Discord:
https://discord.com/channels/830184174198718474/1390746735290355723

Would it be possible to return other/custom data as well? For example:

```typescript
return {
  entries: posts.map((post) => ({
    id: post.slug.current,
    data: post,
  })),
  totalPage: 10 // This is not transcribed
};
```

deslunes:
Hi ! I'm really interested in content collections and making it live extends it to more use cases. But I'm a bit concerned about on-request fetching, depending on the collection, it could end up costing me more than my budget, or hit service limits.

I don't really know if this should be part of live content collections, or a separate suggestion, but I think that it could be a great addition to be able to fetch every X amount of time, in a [CRON](https://en.wikipedia.org/wiki/Cron) way, cache the data until the new fetch happens.

I think it's a great middle ground between on request fetching and build time fetching, it's not as responsive but gives the developer enough control for the kind of data they want to implement. The content collection could be update every minute, hour, or at specific time, specific day.

JonathonRP:
> It does but there’s no advantages over the current content collections.

Interesting, so should work fine like if it was just content collection? but to get the benefits of live ssr is needed, correct?

ascorbic:
## Live Content Collections RFC Changes Summary

*Summary generated by Claude Opus*

This document summarizes the significant API changes to the Astro Live Content Collections RFC since it was opened on May 3, 2025.

---

## Timeline of Changes

### May 2025

#### Initial RFC (May 2-3)
The RFC was opened with the initial proposal for live content loaders - a new type of loader that fetches data at runtime rather than build time. The original API used existing `getCollection`/`getEntry` functions with a `type: "live"` property.

**Original design:**
```ts
const products = defineCollection({
  type: "live",
  loader: storeLoader({ ... }),
});
```

#### Error Handling Request (May 7)
@ematipico requested that the RFC cover error handling scenarios: timeouts, invalid data, parsing errors, etc.

#### API Naming Discussion (May 13)
@sarah11918 raised concerns about using identically-named helper functions (`getEntry`/`getCollection`) that behave differently for live vs. build-time collections. She suggested considering `getLiveEntry()` and `getLiveCollection()` for clearer documentation and support.

#### Cache Hint Rename (May 8)
Renamed the cache hint object for consistency.

#### Major API Overhaul (May 23)
Based on @ematipico and @sarah11918's feedback, two significant changes were made:

**1. Dedicated Functions**

*Before:* Shared `getCollection`/`getEntry` functions
```ts
const { data } = await getCollection("products");
```

*After:* Dedicated `getLiveCollection`/`getLiveEntry` functions
```ts
const { entries } = await getLiveCollection("products");
```

**2. Explicit Error Handling**

*Before:* Functions returned data directly or threw
```ts
const data = await getCollection("products"); // throws on error
```

*After:* Returns object with `{ data, error }` pattern
```ts
const { entries, error } = await getLiveCollection("products");
if (error) {
  return Astro.redirect('/500');
}
```

#### Error Object Refinements (May 23)
Multiple commits refined error handling:
- Renamed `data` to `entries`/`entry` in return types
- Switched to using proper Error objects
- Added detailed error handling patterns

#### Naming Updates (May 28)
Updated naming conventions and corrected custom error information.

---

### June 2025

#### `defineLiveCollection` Function (June 11-13)
Changed from using `defineCollection` with `type: "live"` to a dedicated `defineLiveCollection` function and introduced a separate `live.config.ts` file.

*Before:*
```ts
// src/content.config.ts
import { defineCollection } from "astro:content";

const products = defineCollection({
  type: "live",
  loader: storeLoader({ ... }),
});
```

*After:*
```ts
// src/live.config.ts
import { defineLiveCollection } from "astro:content";

const products = defineLiveCollection({
  loader: storeLoader({ ... }),
});
```

#### Removed Redundant `type: 'live'` (June 27)
Since live collections now use `defineLiveCollection`, the `type: 'live'` property became redundant and was removed. (This addressed @gingerchew's observation from the TBD stream.)

#### Cache Hints Evolution (June 22)
@ascorbic noted that `lastModified` should be added to `cacheHint`, and `maxAge` should potentially be removed since loaders can't normally decide cache duration.

*Before:*
```ts
cacheHint: {
  tags: ["products"],
  maxAge: 3600,
}
```

*After:*
```ts
cacheHint: {
  tags: ["products"],
  lastModified: new Date(product.updatedAt),
}
```

---

### October 2025

#### Type Updates (October 16)
Updated types to reflect changes in the implementation, aligning the RFC with the actual Astro codebase.

---

### January 2026

#### Collection Name in Loader Context (January 30)
Final API change: loader functions now receive the collection name in their context.

*Before:*
```ts
loadCollection: async ({ filter, logger }) => { ... }
loadEntry: async ({ filter, logger }) => { ... }
```

*After:*
```ts
loadCollection: async ({ filter, logger, collection }) => { ... }
loadEntry: async ({ filter, logger, collection }) => { ... }
```

---

## Summary of Key API Changes

| Feature | Original API (May 2025) | Final API (Jan 2026) |
|---------|------------------------|----------------------|
| Collection definition | `defineCollection({ type: "live" })` | `defineLiveCollection()` |
| Config file | `src/content.config.ts` | `src/live.config.ts` |
| Query functions | `getCollection()`/`getEntry()` | `getLiveCollection()`/`getLiveEntry()` |
| Return format | Direct data or throw | `{ entries/entry, error, cacheHint }` |
| Error handling | Try/catch around calls | Result object pattern |
| Cache hints | `{ tags, maxAge }` | `{ tags, lastModified }` |
| Loader context | `{ filter, logger }` | `{ filter, logger, collection }` |

---

## Community Feedback Incorporated

- **@ematipico**: Error handling coverage (addressed May 23)
- **@sarah11918**: Dedicated function names for clarity (addressed May 23)
- **@stipsan**: Discussion on TypeGen and Zod schema integration
- **@boutell**: Perspective on CMS-driven approaches (acknowledged, different use case)
- **@gingerchew**: Redundant `type: 'live'` property (addressed June 27)
- **@Alynva**: Request for clearer SSR/static documentation
- **@delucis**: Clarified SSR requirements in comments

---

## Current Status

As of January 30, 2026, the RFC is now marked as **ready for review** with a **call for consensus** from @withastro/tsc. The implementation PR (withastro/astro#13685) has been merged, and the feature is available experimentally.

ascorbic:
@florian-lefebvre no, just doing it now (it's a 2 line fix, plus docs)

matthewp (approved):

proposals/0055-live-content-loaders.md
```diff
@@ -0,0 +1,684 @@
...
+-->
+
+**If you have feedback and the feature is released as experimental, please leave it on the Stage 3 PR. Otherwise, comment on the Stage 2 issue (links below).**
+
+- Start Date: 2025-05-02
+- Reference Issues: <!-- related issues, otherwise leave empty -->
+- Implementation PR: https://github.com/withastro/astro/pull/13685
+- Stage 1 Discussion: https://github.com/withastro/roadmap/discussions/1137
+- Stage 2 Issue: https://github.com/withastro/roadmap/issues/1151
+- Stage 3 PR: <!-- related roadmap PR, leave it empty if you don't have a PR yet -->
```
florian-lefebvre:
```suggestion
- Stage 3 PR: https://github.com/withastro/roadmap/pull/1164
```

florian-lefebvre (approved):

ematipico (approved):

Eptagone:
Is is possible to make `loadCollection` optional? My API doesn't have a way to list all available content.

ascorbic:
@Eptagone what would you want it to do if there was no loadCollection defined? In your case it might be best to return an error or empty array.

Eptagone:
@ascorbic Nothing. I just don't need loadCollection in my case. I decided to return an error as you mentioned, thx.

Issue #1149, opened 2025-04-02
Content Security Policy (CSP)


ematipico:
- Accepted Date: 2025-04-02
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/377
- Author: @gjdev
- Champion(s): @ematipico 
- Implementation PR: 

# Summary

Add a new feature that allows Astro assets, mainly scripts and styles, to support [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP).

# Background & Motivation

When creating astro components/layouts with client side scripts, as documented here: [docs.astro.build/en/core-concepts/astro-components](https://docs.astro.build/en/core-concepts/astro-components/), the processed+bundled scripts are rendered inline in the html of the page, during a production build.
This forces the use of a CSP with script-src: 'unsafe-inline', which is called unsafe for obvious reasons.

It would be nice to have an astro config option, or a configration attribute for script tags, to instruct astro to create a javascript file for the bundled javascript, and then reference that file via a src attribute on the rendered script tag, instead of inlining the bundled javascript.

(During devlopment npm run dev the processed javascript is actually linked via a script src already, making dev builds and production builds behave differently with some content security policies).

The same could be done for stylesheets, making it configurable whether the stylesheet is inlined in the page (requiring style-src unsafe-inline), or included via a link.

# Goals

- Support at least one solution (`nonce`, hash, etc.) that would support CSP.
- Allow Astro scripts (client islands, server islands, view transitions, etc.) and Astro styles to work out of the box.
- Update adapters to support the feature, if needed.
- Make the feature opt-in, even when it's outside of experimental. 

# Non-Goals 

- Provide *multiple* solutions to fix the same problem
- Support for third-party scripts dynamically imported by users e.g. `document.createElement("script")`, `import("https://path.to/script.js")`;
- Support for third-party styles dynamically imported by users.
- Performance. Users and authors are willing to overlook possible performance drawbacks in order to increase security.

ematipico:
There are mostly two solutions that we are looking at:
- `nonce` Response header
- `hashes`


## The `nonce` header solution

The servrer must create a completely new `nonce` header at each request (it can't be predictable), and this header needs to be used inside the rendered pages. Example:
                                                                         
```js
function content(nonce) {
  return `
    <script nonce="${nonce}" src="/main.js"></script>
    <script nonce="${nonce}">console.log("hello!");</script>
    <h1>Hello world</h1> 
    `;
[...]

PR #1168, opened 2025-05-13
Content Security Policy (CSP)

ematipico:
# Summary

```js
export default defineConfig({
	security: {
		csp: true
	}
})
```

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/feat/rfc-csp/proposals/0055-csp.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/1149)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/377)

### DIFF
--- a/proposals/0055-csp.md
+++ b/proposals/0055-csp.md
@@ -450,8 +450,8 @@ This solution doesn't follow Astro core principles, where an Astro site should w
   })
   ```
 
-- The new Astro feature will be shipped with prefix `_exprimental`: `_experimentalStaticHeaders` in a patch release.
-- The new field of the hook `astro:build:generated` will be shipped with prefix `_experimental`: `_experimentalRouteToHeaders` in a patch release.
+- The new Astro feature will be shipped with the `experimental` prefix: `experimentalStaticHeaders` in a patch release.
+- The new field of the hook `astro:build:generated` will be shipped with the `experimental` prefix: `experimentalRouteToHeaders` in a patch release.
 
 After a period of trial and test with existing Astro functionalities, the experimental flag will be removed.
 


proposals/0055-csp.md
```diff
@@ -0,0 +1,251 @@
...
+
+The *main* reason for using this strategy instead of `Response` headers is because this solution allows Astro to support two important use cases:
+- SSG, especially when no adapters are involved.
+- SPA (Single Page Applications), where scripts and styles *might* be loaded dynamically using client-side routing.
+
+## Generation of hashes
+
+The hashes will be generated using the [`crypto`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto) global object, and the package [`@oslojs/encoding`](https://encoding.oslojs.dev/). Astro *already uses* this package to encrypt props for server islands.
+
+Hashes will be generated using the `SHA-256` algorithm. Users can't change the algorithm.
```
Marocco2, line 79: I would advise against hard coding SHA-256 as SHA-384 or SHA-512 are already deployed in the wild.
Following CSP spec would be a better option

https://developer.mozilla.org/en-US/docs/Glossary/Hash_function

ematipico: Are you suggesting that users provide their own function? Or are you suggesting something else?

Marocco2: > Are you suggesting that users provide their own function? Or are you suggesting something else?

I'm suggesting that CSP integration should provide SHA-256, SHA-384 and SHA-512 as options for algorithm as they are pretty much standard across web dev space.

Also, I see now that `styleHashes` and `scriptHashes` sha256 requirement is being dropped, so I'm happy about it 👍

ematipico:
@NepCono in the RFC I explained why nonce wouldn't work for Astro. Did you read that part?

NepCono:
@ematipico just read about the nonce, very understandable :-) I was just saying that the nonce solution would be more simplified than the hash. 

However still a question remaining would the out of the box implementation security > csp and strictdynamic > true that we dont have to insert a nonce on the script (not a server island:script)  which will lead to using typescript instead of javascript ?

btw; when will this be released and can I access it earlier ?

proposals/0055-csp.md
```diff
@@ -0,0 +1,251 @@
...
+
+- Start Date: 2025-05-13
+- Reference Issues: 
+- Implementation PR: 
+- Stage 2 Issue: https://github.com/withastro/roadmap/issues/1149
+- Stage 3 PR: https://github.com/withastro/roadmap/discussions/377
+
+# Summary
+
+Add a new feature that allows Astro assets, mainly scripts and styles, to support [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP).
```
florian-lefebvre, line 23: I don't think we'll handle anything else for now?
```suggestion
Add a new feature that allows Astro scripts and styles to support [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP).
```

proposals/0055-csp.md
```diff
@@ -0,0 +1,251 @@
...
+})
+```
+
+# Background & Motivation
+
+Include any useful background detail that that explains why this RFC is important.
+What are the problems that this RFC sets out to solve? Why now? Be brief!
+
+It can be useful to illustrate your RFC as a user problem in this section.
+(ex: "Users have reported that it is difficult to do X in Astro today.")
```
florian-lefebvre, line 44: This is from the default template, we can probably extract things from the original discussion

proposals/0055-csp.md
```diff
@@ -0,0 +1,251 @@
...
+- Update adapters to support the feature, if needed.
+- Make the feature opt-in, even when it's outside experimental.
+- Support the [dev server](#dev-server) (stretch)
+- View transitions (limited/opt-in support)
+
+# Non-Goals
+
+- Provide *multiple* solutions to fix the same problem
+- Support for third-party scripts dynamically imported by users e.g. `document.createElement("script")`, `import("https://path.to/script.js")`;
+- Support for third-party styles dynamically imported by users.
```
florian-lefebvre, line 59: If I undestand correctly, `injectCspAsset` would allow this

proposals/0055-csp.md
```diff
@@ -0,0 +1,251 @@
...
+Users can add an entire directive. Astro won't check if the directive is grammatically correct.
+
+```js
+Astro.insertCspPolicy(`image-src '${Astro.site}'`)
+```
+Result
+```html
+<meta 
+  http-equiv="Content-Security-Policy"
+  content="image-src 'https://example.com'; style-src 'self' 'sha256-hash1'; script-src 'self' 'sha256-hash2'"
```
florian-lefebvre, line 175: Does order matter?

ematipico: Can you be more specific with your query? The order of what :)

proposals/0055-csp.md
```diff
@@ -0,0 +1,251 @@
...
+- server islands
+- view transitions
+- external scripts (via APIs)
+- external styles (via APIs)
+- inline scripts
+- inline styles
+
+# Drawbacks
+
+The hashes solution will negatively impact the rendering engine because Astro will have to calculate the hashes **at runtime**. The most expensive solution is the support of server islands because the contents of their scripts depend on props and slots, which are only known during the rendering phase.
```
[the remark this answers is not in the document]
ematipico: No, but we can set up a codspeed benchmark

proposals/0055-csp.md
florian-lefebvre: Sorry! I meant does it matter if it's first image-src, then style-src, then script-src or can it be in any order? eg. image-src at the end

Marocco2: It won't matter

proposals/0055-csp.md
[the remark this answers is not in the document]
Marocco2: On a comparable solution, it would take around 500ms for dozens of pages in a Netlify Free tier build. This should be the base performance

proposals/0055-csp.md
ematipico: I want to rephrase it then. What I meant is that **out of the box** Astro won't create a hash for the dynamically imported script/style. Any suggestions?

ematipico:
I changed the APIs for additional directives. I believe the new API is better because counts for better DX:

[`cc6d3bf` (#1168)](https://github.com/withastro/roadmap/pull/1168/commits/cc6d3bf8136a147212cf3074decd69c9133228d0)

proposals/0055-csp.md
```diff
@@ -0,0 +1,421 @@
...
+export default defineConfig({
+    security: {
+        csp: {
+            // or SHA-384, or SHA-256 (default)
+            algorithm: "SHA-512",
+            // Additional directives added to the final content
+            directives: [
+                {
+                    type: "default-src",
+                    content: "'sef'"
```
Marocco2, line 143: Typo
```suggestion
                    content: "'self'" 
```

Marocco2 (changes requested):

proposals/0055-csp.md
```diff
@@ -0,0 +1,443 @@
...
+```
+
+Will result in something similar to the following (the result is formatted to allow better readability):
+
+```html
+<meta 
+  http-equiv="Content-Security-Policy"
+  content="
+    default-src 'self';
+    img-src 'self' 'https://images.cdn.example.com';
```
Johannes-Andersen, line 238: This would be invalid. Ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#host-source

```suggestion
    img-src 'self' https://images.cdn.example.com;
```

Screenshot 2025-06-06 at 00 19 22

kurtextrem:
Hi, a CSP meta tag currently disables the preload scanner in all Chrome versions: https://rviscomi.github.io/capo.js/user/validation/#no-meta-csp. So adding the CSP via the meta tag might make the load perf drastically worse for some sites.
The Chrome team is working on a fix: https://issues.chromium.org/issues/40273969, but with an unknown ETA.

Suggestions:
- Add a clear warning for this to the docs.
- I think it might even be worth it to consider adding support for the HTTP header meanwhile, so that at least non-static sites have an alternative way to stay performant.

florian-lefebvre:
FYI per the RFC non goals:

> Performance. Users and authors are willing to overlook possible performance drawbacks in order to increase security.

kurtextrem:
The "drawbacks" section only mentions the rendering engine. As an Astro static-build user, I'd expect my build to take a little more time when reading this; not that my website becomes slower.
Additionally, the docs don't mention this at all. IMHO this should be part of the docs.

jlarmstrongiv:
In addition to the `strictDynamic` option, it would be nice to have an `unsafeEval` option.

It is very [common](https://stackoverflow.com/a/37177219) for wasm or performance optimized libraries to depend on `eval` or `new Function`, and having the option to allow that is important.  Just off the top of my head:
- [wasm-vips](https://www.npmjs.com/package/wasm-vips) depends on `eval`
- [fast-cartesian](https://www.npmjs.com/package/fast-cartesian) depends on `new Function`

ematipico:
> The "drawbacks" section only mentions the rendering engine. As an Astro static-build user, I'd expect my build to take a little more time when reading this; not that my website becomes slower. Additionally, the docs don't mention this at all. IMHO this should be part of the docs if it's a non-goal.

This is an experimental feature and it moves really fast, so I think it's more sensible to avoid documenting things like goals/non-goals until the feature is stable.

proposals/0055-csp.md
```diff
@@ -0,0 +1,443 @@
...
+
+# Detailed Design
+
+## Solution adopted
+
+Contrary to other frameworks of the same ecosystem, our solution will rely on generating [**hashes of styles and scripts**](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP#hashes). For more information, please refer to the [alternatives](#alternatives) section.
+
+Astro provides the hashes during the rendering phase, they will be written in the `<head>` tag using the `<meta http-equiv="Content-Security-Policy">` meta tag.
+
+The *main* reason for using this strategy instead of `Response` headers is because this solution allows Astro to support two important use cases:
```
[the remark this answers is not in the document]
ematipico: That's unsettling. Isn't it possible to use nonce and hashes at the same time?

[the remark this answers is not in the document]
NepCono: What I mean, you are truly doing a good job with the Astro out of the box solution. 

However what I am telling I do not want to use the meta tags implementation because it is an issue, like @Johannes-Andersen mentioned.

I already have my CSP strict dynamic setup through middleware. Using astro.locals.nonce setup. I inject the nonce now in  my scripts however than I cannot use typescript anymore very annoying but hey that's life :-) 

Then I really wanted to use preact for the island behaviour and than I noticed that my CSP messed up life. So I start trying to make a work-around which is not working yet. I use output = server so my site is dynamic and I really don't want to loose performance. Calculating hashes doesn't work because I have no static pages. I used to have a blog the hashes integration only calculated them for the static pages. So I tried middleware to adjust the generated html and insert nonces but that is not working as it should be.

It would be nice to setup a nonce which is injected by astro itself on page generation. Or maybe someone can help me to make the solution.

But I really appreciate your efforts.

danwulff:
> I don't think it should be a blocker for release, but header support would be a good future goal, in addition to meta tags

For me specifically, I don't use on-demand rendering. I just deploy static files to CloudFlare. It would be awesome to have astro build produce calculated SHAs that I could then use to generate a headers config file (no matter where I'm hosting)

Right now I'm:

1.  using CSP 'self' values for styles via .css files and scripts via .js files
2. looping over all .html files within `/dist` and calculating the SHAs of both `<style>` elements and `style=""` attributes and `<script>` elements
3. Transforming those values to CloudFlare's `_headers` file format

It would be awesesome to see Astro support steps ^ 1 & 2, and then the last step would just be transforming those values into a config for whatever host is used.

But as mentioned previously, meta tags seem like a great first step. It would just be nice to see some support for the calculated SHA values to be used elsewhere either as a hook in the build process, or as a produced temp file that could be utilized for post-build scripting. If that makes sense.

proposals/0055-csp.md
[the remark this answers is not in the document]
matthewp: @NepCono our implementation is current based on hashes, why do you want to use nonce instead?

ematipico:
@matthewp 

> Provide the CSP mapping in one of the config hooks

What's a CSP mapping and what would it look like?

proposals/0055-csp.md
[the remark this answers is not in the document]
NepCono: Because the current hash implementation is through meta tags implementation, where somethings are not working.

Than I use a middleware implementation that I build myself to set to corresponding CSP response header.

With the current implementation of hashes being calculated everytime, it is far better and faster to use a nonce token. You only have to inject the nonce token in the scripts and your done (Astro.locals.nonce) when a user sets that problem solved. No hash calculation slowing things down, but hey that's only my opnion.

Also what @Johannes-Andersen is referring to.

In the mean time I am stuck, can you supply me with an astro integration that scans my pages (output server) so dynamic pages and or inject or hashes so I can move on ?

ematipico:
I updated the solution provided by Astro. Now it's a mix of `<meta>` tag and `Response` headers, and adapters will be able to take advantage of their "hosted solution" for static pages: [`c8843f6` (#1168)](https://github.com/withastro/roadmap/pull/1168/commits/c8843f6e4b6e58eb880e81f91ea4a608795645a0)

proposals/0055-csp.md
```diff
@@ -0,0 +1,454 @@
...
+
+## Solution adopted
+
+Contrary to other frameworks of the same ecosystem, our solution will rely on generating [**hashes of styles and scripts**](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP#hashes). For more information, please refer to the [alternatives](#alternatives) section.
+
+Astro provides the hashes during the rendering phase. These hashes will served differently based on different cases:
+- For static pages, they will be served in the `<head>` tag using the `<meta http-equiv="Content-Security-Policy">` meta tag.
+- For dynamic pages, they will be served in the `content-security-policyt` header of the `Response` type.
+
+A [new adapter feature](https://docs.astro.build/en/reference/adapter-reference/#adapter-features) will be provided, called `hostedCsp`. For adapters that support this feature, Astro **won't** serve the hashes for *static pages*`.
```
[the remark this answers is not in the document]
ematipico: I don't mind renaming. 

> The we could use it for other stuff too.

What do you mean? Do you plan to make this feature generic?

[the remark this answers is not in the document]
ascorbic: The thing we're asking is whether the adapter supports setting headers for static pages, not anythign CSP-specific. I don't htink we have any specific plans, but if we keep the feature name generic, we can use it in future if we have anything else that uses headers.

proposals/0055-csp.md
```diff
@@ -0,0 +1,454 @@
...
+
+# Detailed Design
+
+## Solution adopted
+
+Contrary to other frameworks of the same ecosystem, our solution will rely on generating [**hashes of styles and scripts**](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP#hashes). For more information, please refer to the [alternatives](#alternatives) section.
+
+Astro provides the hashes during the rendering phase. These hashes will served differently based on different cases:
+- For static pages, they will be served in the `<head>` tag using the `<meta http-equiv="Content-Security-Policy">` meta tag.
+- For dynamic pages, they will be served in the `content-security-policyt` header of the `Response` type.
```
ascorbic, line 74:
```suggestion
- For dynamic pages, they will be served in the `content-security-policy` header of the `Response`.
```

proposals/0055-csp.md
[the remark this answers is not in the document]
ematipico: Addressed here [`eb28bb3` (#1168)](https://github.com/withastro/roadmap/pull/1168/commits/eb28bb3530ef4fb0ac03848b5db9f2833fbb20d8)

proposals/0055-csp.md
```diff
@@ -0,0 +1,454 @@
...
+  http-equiv="Content-Security-Policy"
+  content="style-src 'https://styles.cdn.example.com' 'sha256-hash1' 'sha256-qewpiorfuvng3'; script-src 'self' 'sha256-hash2'"
+>
+```
+
+
+#### `Astro.insertScriptHash(string)`
+
+```js
+Astro.insertScriptHash("sha256-qewpiorfuvng3")
```
Johannes-Andersen, line 376:
```suggestion
Astro.insertScriptHash("'sha256-qewpiorfuvng3'")
```

proposals/0055-csp.md
```diff
@@ -0,0 +1,454 @@
...
+#### `Astro.setScriptResource(string)`
+
+```js
+Astro.setScriptResource("'https://scripts.cdn.example.com'")
+```
+Result
+```html
+<meta 
+  http-equiv="Content-Security-Policy"
+  content="style-src 'self' 'sha256-hash1' 'sha256-qewpiorfuvng3'; script-src 'https://scripts.cdn.example.com' 'sha256-hash2'"
```
Johannes-Andersen, line 395:
```suggestion
  content="style-src 'self' 'sha256-hash1' 'sha256-qewpiorfuvng3'; script-src https://scripts.cdn.example.com 'sha256-hash2'"
```

Johannes-Andersen (commented):
Great work on the proposal! Really exited with the new Header and adapter support 🙌

It could be a benefit to document this:
> Authors are strongly encouraged to place [meta](https://html.spec.whatwg.org/multipage/semantics.html#meta) elements as early in the document as possible, because policies in [meta](https://html.spec.whatwg.org/multipage/semantics.html#meta) elements are not applied to content which precedes them. In particular, note that resources fetched or prefetched using the Link HTTP response header field, and resources fetched or prefetched using [link](https://html.spec.whatwg.org/multipage/semantics.html#the-link-element) and [script](https://html.spec.whatwg.org/multipage/scripting.html#script) elements which precede a [meta](https://html.spec.whatwg.org/multipage/semantics.html#meta)-delivered policy will not be blocked.

– https://www.w3.org/TR/CSP3/#meta-element

Would also be valuable to make sure the Astro puts it on the top of head?
Seems from my quick testing it is added at the bottom of `<head />` atm: 
Screenshot 2025-06-11 at 14 15 21

proposals/0055-csp.md
```diff
@@ -0,0 +1,443 @@
...
+Result
+```html
+<meta 
+  http-equiv="Content-Security-Policy"
+  content="style-src 'https://styles.cdn.example.com' 'sha256-hash1' 'sha256-qewpiorfuvng3'; script-src 'self' 'sha256-hash2'"
+>
+```
+
+
+#### `Astro.insertScriptHash(string)`
```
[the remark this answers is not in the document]
ematipico: I still think this API isn't useful, as it essentially creates some unusual usage. Let's say you have some `script.js` somewhere. What would happen is the following:

- The user must import `script.js` *somewhere*, and provide its content *as a string* to the `generateHash`
- The generated hash must be saved or sent somewhere
- Then the user must provide this hash to Astro via configuration or runtime
- This very script must be added to the application 

Let's wait and see.

ematipico:
Thank you @Johannes-Andersen, I applied your suggestions. It makes more sense to have the user write their format (apostrophe VS no apostrophe). Since this is an advanced API, we can safely assume that users know what they are doing. I will have to send a PR to change the the APIs.

ematipico:
@stefanmaric Thank you for your suggestion. Unfortunately, this API wouldn't play well in Astro, where we need more control over the `script-src` and `style-src` directives. Also, in the future, we might need to have control over other directives too.

proposals/0055-csp.md
```diff
@@ -0,0 +1,443 @@
...
+```ts
+export const ALLOWED_DIRECTIVES = [
+	'base-uri',
+	'child-src',
+	'connect-src',
+	'default-src',
+	'fenced-frame-src',
+	'font-src',
+	'form-action',
+	'frame-ancestors',
```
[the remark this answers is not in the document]
ematipico: With Astro v5.9.4, we shipped an adapter feature that allows to ship static headers using the hosting platform. As for now, only the Netlify adapter supports that: https://docs.astro.build/en/reference/adapter-reference/#experimentalstaticheaders

This should allow us to ship CSP headers via `Response` headers for static pages.

HannesOberreiter:
Great implementation as far as I can tell. Two things I want to throw into the discussion:

## Naming

Currently the Runtime APIs do not really reflect in their naming that they have something to do with CSP. Maybe they need a kind of prefix, eg. `Astro.cspInsertStyleResource`

## Disable one of the two

Currently it is only possible to activate both `script-src` and `style-src` but not only one. Additionally to this, I would love a Runtime API which does completely overwrites the directives and not only add to resources/hashes (maybe an option flag for the current APIs, eg. overwrite = true). The reason on some pages I use legacy external scripts which apply a lot of inline css, and I can't be the only one with this problem.

ematipico:
> If in your Astro config, you have `directives: ["img-src 'self'"]` and you call `Astro.insertDirective("img-src whatever")`, what happens? Does it override, merge, or add (which is incorrect)?

Since it's an insertion, this API will **add an item to the existing array**. 

If that's incorrect, what do you expect from the API?

florian-lefebvre:
I would expect it to merge, like what we did for fonts basically

ematipico:
> I would expect it to merge, like what we did for fonts basically

Why though? I'm not saying it's incorrect, but `font-src` is particular because Astro needs to own `font-src`. If Astro doesn't own the other resources, why should Astro deduplicate the values?

ematipico:
> @ematipico surely it should just not send the headers in dev, not break dev

Makes sense. @Eveeifyeve feel free to file an issue, we will look at it

florian-lefebvre:
> Why though? I'm not saying it's incorrect, but font-src is particular because Astro needs to own font-src. If Astro doesn't own the other resources, why should Astro deduplicate the values that aren't owned?

Sorry I was on phone so it wasn't very clear. My point is not about deduplication but rather avoiding loosing directives. Taking my example again:

> If in your Astro config, you have `directives: ["img-src 'self'"]` and you call `Astro.insertDirective("img-src whatever")`, what happens? Does it override, merge, or add (which is incorrect)?

Here are the 3 outcomes:
- override: `img-src whatever;`. That is wrong IMO because you likely want to inherit the config from your astro config
- add (current IIRC): `img-src 'self'; img-src whatever;`. I think this is wrong because only the last occurence will be picked up
- merge (my recommendation): `img-src 'self' whatever;`. We preserve everything

What do you think?

cleancommunities:
Hello, I was pointed in this direction from the discord, in relation to an inquiry about use cases of, "style-src-attr". Repo, cleancommunities/csp-test https://discordapp.com/channels/830184174198718474/830184175176122389/1411782698720170046

ematipico:
@florian-lefebvre 

I would go for the third option, however I think we should change the APIs, the make the merging safer and less prone to error on our side e.g. 

```ts
Astro.csp.insertDirective("img-src", "whatever")
```

florian-lefebvre:
Not necessarily, we already have the logic for it with `font-src`

matthewp (approved):

proposals/0055-csp.md
```diff
@@ -0,0 +1,461 @@
...
+  To use this template: create a new, empty file in the repo under `proposals/${ID}.md`.
+  Replace `${ID}` with the official accepted proposal ID, found in the GitHub Issue
+  of the accepted proposal.
+-->
+
+**If you have feedback and the feature is released as experimental, please leave it on the Stage 3 PR. Otherwise, comment on the Stage 2 issue (links below).**
+
+- Start Date: 2025-05-13
+- Reference Issues: 
+- Implementation PR:
```
florian-lefebvre:
```suggestion
- Implementation PR: https://github.com/withastro/astro/pull/13802
```

proposals/0055-csp.md
florian-lefebvre:
```suggestion
- Automatic support for third-party styles dynamically imported by users.
```

proposals/0055-csp.md
```diff
@@ -0,0 +1,461 @@
...
+  ```js
+  export default defineConfig({
+      experimental: {
+          csp: true
+      }
+  })
+  ```
+
+- The new Astro feature will be shipped with prefix `_exprimental`: `_experimentalStaticHeaders` in a patch release.
+- The new field of the hook `astro:build:generated` will be shipped with prefix `_experimental`: `_experimentalRouteToHeaders` in a patch release.
```
florian-lefebvre: When you make a PR against `next` to stabilize CSP, we gotta remember to update these names

ematipico: `_experimentalRouteToHeaders` was added for CSP, but it's been designed to be generic. Which means that the two features aren't directly connected, and they can become stable independently. If by the time of stabilisation, `_experimentalRouteToHeaders` is still named like this, we shouldn't change it.

wildfiremedia:
Link to Rendered RFC is broken

https://github.com/withastro/roadmap/blob/feat/rfc-csp/proposals/0055-csp.md

Issue #1140, opened 2025-03-19
Route caching


ascorbic:
- Accepted Date:  2025-03-19
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/1131 https://github.com/withastro/roadmap/discussions/181
- Author: @ascorbic @matthewp 
- Champion(s): @ascorbic 
- Implementation PR: 

# Astro route caching

# Summary

Introduce a platform-agnostic route caching API for Astro SSR pages. This feature enables developers to declaratively define caching rules per route and dynamically override them in middleware or API handlers. It provides a consistent API across Node.js, Vercel, Netlify, Cloudflare, and other deployment targets, ensuring efficient caching while allowing cache invalidation by path or tag when supported. Where possible it will be implemented using CDN caches, but with fallback implemntations where this isn't supported.

# Background & Motivation

Currently, caching in Astro SSR requires manually setting `Cache-Control` or `CDN-Cache-Control` headers, which vary per platform and is not ergonomic. Many modern frameworks offer built-in support for SSR caching strategies, allowing pages to be served efficiently while reducing function invocations and improving performance.

Astro's current lack of a unified caching API means that users must implement custom caching solutions per deployment target, leading to unnecessary complexity. This proposal introduces a framework-level abstraction for caching that integrates seamlessly with Astro’s adapter system, providing an intuitive DX while using platform-specific caching mechanisms where available.

# Goals

- Provide a **platform-agnostic** caching API for Astro SSR routes and server islands.
- Support **TTL, stale-while-revalidate (SWR), and ETag-based caching**.
- **Cache in the CDN** where possible, using HTTP cache or `CacheStorage`.
- Allow **declarative caching definitions** in config, with overrides in Astro pages and server islands.
- Enable **imperative cache invalidation** via by path and tag for fine-grained control.
- Use **platform-specific caching mechanisms** (CDN, edge cache, etc.) via Astro adapters where available.
- Maintain a **consistent API and DX across deployment targets**.

# Non-goals

- Persistent cache storage in core. This is delegated to the CDN.
- `fetch` cache. This is for responses only.
- Page fragment cache, except for server islands. This is designed for caching complete responses.
- Browser cache handling, except for setting etag and respecting conditional requests.
- Cache rules for prerendered pages.

# Example

## Declarative API

Cache rules can be declared via route patterns in config:

```js
// astro.config.mjs
export default {
  cache: {
    routes: {
      '/': { maxAge: 0, swr: 60 },
[...]

PR #1245, opened 2025-10-15
Route Caching

ascorbic:
# Summary

A platform-agnostic route caching API for Astro SSR pages that enables declarative cache control using web standards. 

# Examples

## Basic route caching

```astro
---
// src/pages/products/[id].astro
import { getEntry } from 'astro:content';
const product = await getEntry('products', Astro.params.id);
Astro.cache.set({
  lastModified: product.updatedAt,
  maxAge: 300,  // Cache for 5 minutes
  swr: 3600,    // Stale-while-revalidate for 1 hour
  tags: ['products', `product:${product.id}`]
});
---
<h1>{product.data.name}</h1>
<p>{product.data.description}</p>
```

## Automatic dependency tracking

```ts
// src/pages/api/revalidate.ts
import { getLiveEntry } from "astro:content";

export const POST: APIRoute = async ({ cache, request }) => {
  const { id } = await request.json();
  const { entry } = await getLiveEntry("products", id);

  // Invalidate all pages that depend on this entry
  cache.invalidate(entry);

  return Response.json({ ok: true });
};
```

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/feat/route-caching/proposals/0056-route-caching.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/1140)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/1131)

### DIFF
--- a/proposals/0056-route-caching.md
+++ b/proposals/0056-route-caching.md
@@ -488,15 +488,25 @@ Astro will ship with built-in providers for common platforms:
 
 ### Driver Packages
 
-Driver packages can be distributed as npm packages:
+Driver packages can be distributed as npm packages. They are shown with the `@astrojs` scope for illustration, but could be published by providers themselves.
 
-- `@astrojs/cache-cloudflare-cdn` - Cloudflare as proxy in front of other origins
+- `@astrojs/cache-cloudflare-cdn` - Cloudflare as proxy in front of other origins. Distributed separately from the Cloudflare adapter, because it would usually be used with the Node adapter.
 - `@astrojs/cache-fastly` - Fastly CDN support
 - `@astrojs/cache-akamai` - Akamai edge caching
 - Custom providers for other CDNs or caching solutions
 - Custom runtime providers using external stores (Redis, Memcached, etc.)
 
-Driver packages export a cache provider that can be referenced by package name in the config and dynamically imported at runtime.
+Driver packages export a function that accepts driver-specific options and returns a cache provider configuration. This pattern provides type-safe configuration and follows Astro's established patterns (similar to the fonts API):
+
+```ts
+// @astrojs/cache-fastly
+export function cacheFastly(options: FastlyOptions) {
+  return {
+    entrypoint: new URL("./provider.js", import.meta.url),
+    options,
+  };
+}
+```
 
 ## Adapter Abstraction
 
@@ -644,28 +654,36 @@ for (const [key, value] of cache.entries()) {
 
 ### Adapter Integration
 
-Adapters provide default cache drivers via the integration API, similar to how they provide session drivers. Users can override the default driver in their config.
+Adapters provide default cache drivers via the integration API, similar to how they provide default session drivers. Users can override the default driver in their config.
 
 **Driver Loading:**
 
-Similar to sessions, cache drivers are loaded at runtime. Driver packages are referenced by their module ID, and dynamically imported. The Netlify, Vercel, Cloudflare Workers, and Node memory drivers are published as subpath exports of their respective adapter packages.
+Cache driver functions return a configuration object containing an entrypoint and options. The entrypoint points to the provider implementation that is dynamically imported at runtime. The Netlify, Vercel, Cloudflare Workers, and Node memory drivers are published as subpath exports of their respective adapter packages because they are only ever used alongside those adapters.
 
 **Adapter providing a default:**
 
 ```ts
 // Example: @astrojs/vercel adapter
+import { cacheVercel } from "@astrojs/vercel/cache";
+
 export function vercel() {
   return {
     name: "@astrojs/vercel",
     hooks: {
+      "astro:config:setup": ({ config, updateConfig }) => {
+        // Only provide default if user hasn't configured a driver
+        if (!config.cache?.driver) {
+          updateConfig({
+            cache: {
+              driver: cacheVercel(),
+            },
+          });
+        }
+      },
       "astro:config:done": ({ setAdapter }) => {
         setAdapter({
           name: "@astrojs/vercel",
           // ...other adapter properties
-          // Provide default cache driver
-          cache: {
-            driver: "@astrojs/vercel/cache", // Default
-          },
         });
       },
     },
@@ -677,26 +695,35 @@ export function vercel() {
 
 ```ts
 // astro.config.ts
+import { cacheFastly } from "@astrojs/cache-fastly";
+
 export default defineConfig({
   adapter: node(),
   cache: {
     // Override Node's default in-memory cache with Fastly
-    driver: "@astrojs/cache-fastly",
-    options: {
+    driver: cacheFastly({
       serviceId: process.env.FASTLY_SERVICE_ID,
       token: process.env.FASTLY_TOKEN,
-    },
+    }),
   },
 });
 ```
 
 **Driver packages:**
 
-Driver packages like `@astrojs/cache-fastly` export a cache provider implementation:
+Driver packages export a configuration function that accepts type-safe options and returns a provider configuration. The provider implementation is then dynamically imported at runtime:
 
 ```ts
-// @astrojs/cache-fastly/index.ts
-export default function fastlyProvider(options) {
+// fastly/index.ts
+export function cacheFastly(options: FastlyOptions) {
+  return {
+    entrypoint: new URL("./provider.js", import.meta.url),
+    options,
+  };
+}
+
+// fastly/provider.ts
+export default function fastlyProvider(options: FastlyOptions) {
   return {
     name: "fastly",
     setHeaders(cacheOptions) {
@@ -709,11 +736,9 @@ export default function fastlyProvider(options) {
 }
 ```
 
-Astro dynamically imports the package when referenced by name in the config.
-
 **Precedence:**
 
-1. User-configured `cache.driver` in `astro.config.ts` (highest priority)
+1. User-configured `cache.driver` in `astro.config.ts`
 2. Adapter-provided default via `setAdapter()`
 
 ## Dependency Tracking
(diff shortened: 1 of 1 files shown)


proposals/0056-route-caching.md
```diff
@@ -0,0 +1,1149 @@
...
+
+**User overriding the default with a driver package:**
+
+```ts
+// astro.config.ts
+export default defineConfig({
+  adapter: node(),
+  cache: {
+    // Override Node's default in-memory cache with Fastly
+    driver: "@astrojs/cache-fastly",
```
florian-lefebvre: I think I'd prefer an API like fonts, where you call a function. That would make it easier to have custom drivers:
```js
driver: cacheFastly({
	serviceId: process.env.FASTLY_SERVICE_ID,
	token: process.env.FASTLY_TOKEN,
})
```

ascorbic: Ah got it. Yeah, that could be a better pattern, as it makes it easier to type the options

florian-lefebvre: LGTM!

florian-lefebvre:
I'm not sure that's something for Astro to provide, feels like maybe it should be achievable through the driver platform (eg. cloudflare)?

wildfiremedia:
My other thought is that if we could set a max memory limit, it might help prevent the server from starving or running out of memory when a malicious or heavy request comes in. Should it only cache those with HTTP respond code 200?

https://example.com?test=1
https://example.com?test=...
https://example.com?test=1000

ascorbic:
I've been going through the RFC and made some changes to clarify things, tidy things up etc.

### `Astro.cache` as an object with methods

Changed from `Astro.cache()` (callable) to `Astro.cache.set()`. This is more consistent with other Astro global (`Astro.cookies.set()`, `Astro.session.set()`) and allows us to add `invalidate` without an unusual callable-with-methods pattern. The full `AstroCache` interface:

```ts
interface AstroCache {
  set(options: CacheOptions | CacheHint | LiveDataEntry | false): void;
  readonly tags: string[];
  invalidate(options: InvalidateOptions | LiveDataEntry): Promise<void>;
}
```

Available as `Astro.cache` in pages and `context.cache` in API routes/middleware.

### Merge semantics for multiple `cache.set()` calls

Calling `cache.set()` multiple times merges rather than replaces:

- `maxAge`, `swr`, `etag`: last-write-wins
- `lastModified`: most recent date wins
- `tags`: accumulated (union, deduplicated)
- `false`: overrides everything, disables caching

This matches the existing semantics for live collecitons, and makes it natural to compose cache hints from multiple data sources (e.g. a page that depends on both a product and an author entry). @stipsan's [Sanity use case](#issuecomment-2840780591) would usually do this via the live collections API, which allows tags to be returned as part of the loader response.

### `cache.set(false)` for opting out

If a route matches a config-level cache rule but shouldn't be cached, `Astro.cache.set(false)` explicitly disables caching.

### Unified invalidation on `Astro.cache`

Per @florian-lefebvre's [feedback](#issuecomment-2730668775), invalidation now lives on the same object: `Astro.cache.invalidate()` in pages, `context.cache.invalidate()` in API routes/middleware.

### `InvalidateOptions` simplified

Removed the `tag`/`tags` duality. Now just `tags: string | string[]`.

### API routes can set cache, not just invalidate

Added examples showing `cache.set()` in API routes and middleware, not just `cache.invalidate()`.

### Config route precedence clarified

Explicitly documented: no merging between config patterns, most specific wins, each pattern must be fully specified. Priority order:

1. `cache.set(false)` in route (disables)
2. `cache.set()` in route or middleware
3. Most specific config pattern
4. Less specific config patterns
5. No caching (default)

### No caching in dev

Explicitly documented that caching is not active during development. The API is available but calls are no-ops.

### Header stripping clarified

For header-based providers, the CDN strips `CDN-Cache-Control` per RFC 9213. For runtime providers, the framework strips `CDN-Cache-Control` and `Cache-Tag` from the outgoing response.

proposals/0056-route-caching.md
```diff
@@ -0,0 +1,1360 @@
...
+```
+
+## Define cache in config
+
+```ts
+// astro.config.ts
+export default defineConfig({
+  adapter: node(),
+  cache: {
+    routes: {
```
florian-lefebvre: This makes me think about <https://nitro.build/config#routerules> in nitro. So I'm wondering: is there any benefit to have this under `cache`? Instead that could be on the top level and we could also set `prerender: boolean` there.


So to recap, this looks good to me as is but maybe there's a way to future proof it and enable more cool things

proposals/0056-route-caching.md
```diff
@@ -0,0 +1,1360 @@
...
+    },
+  },
+});
+```
+
+# Background & Motivation
+
+Caching is essential for performant server-rendered applications, but current solutions have significant limitations:
+
+**Platform-specific implementations**: Most caching solutions (like ISR) are tightly coupled to specific hosting platforms. ISR was created for Next.js on Vercel and while other platforms have implemented versions of it, they're often limited or second-class implementations. This creates vendor lock-in and inconsistent behavior across deployments.
```
florian-lefebvre, line 193: What does that mean for <https://docs.astro.build/en/guides/integrations-guide/vercel/#isr>? Would we remove it in favor of `Astro.cache` using <https://vercel.com/changelog/tag-based-cache-invalidation-now-available-for-all-responses> under the hood?

proposals/0056-route-caching.md
```diff
@@ -0,0 +1,1360 @@
...
+- **Live collections integration**: Seamless integration with live collections by passing entries or cache hints directly to `Astro.cache.set()`
+- **Automatic dependency tracking**: Invalidate all pages that depend on changed content entries
+- **Adapter abstraction**: Adapters provide platform-specific implementations with consistent behavior
+- **Pluggable cache drivers**: Users can configure cache drivers for different CDN providers (Fastly, Cloudflare, Akamai, etc.)
+- **Node.js support**: Functional in-memory caching for Node.js deployments
+
+# Non-Goals
+
+- **Strong consistency guarantees**: Cache invalidation follows eventual consistency model
+- **Distributed caching for Node.js**: Node adapter uses in-memory cache, unsuitable for multi-instance deployments without external cache
```
florian-lefebvre, line 219: I assume you mean "by default" but it can be configured

proposals/0056-route-caching.md
```diff
@@ -0,0 +1,1360 @@
...
+- **Static/prerendered page caching**: This is for on-demand SSR routes only; static and prerendered routes are already cached by default and don't need route-level cache control
+- **Partial page caching**: This focuses on full-page route caching, not fragment or component-level caching
+- **Browser caching**: This focuses on CDN and server-side caching, not browser caching. Only `Last-Modified` and `ETag` headers are sent to browsers for conditional requests
```
florian-lefebvre, line 222: Can users access the current state of the cache (eg. after several `cache.set()` calls) so they can use this data to add browser caching headers? So not only `Astro.cache.tags`

proposals/0056-route-caching.md
```diff
@@ -0,0 +1,1360 @@
...
+
+## Cache Provider Interface
+
+Cache providers implement platform-specific caching logic while maintaining a consistent developer experience. Similar to how sessions use pluggable storage drivers, route caching uses pluggable cache providers.
+
+### Provider Interface
+
+```ts
+interface CacheProvider {
+  name: string;
```
florian-lefebvre, line 360: It's probably more useful to have the name in the `cache.provider` than the runtime implementation, so it can be used in logging etc

proposals/0056-route-caching.md
ascorbic: hmm, yeah. Good idea. routeRules would be a better name in that scneario too

proposals/0056-route-caching.md
ascorbic: I think so, or leave it deprecated and make it conflice with the new caching

proposals/0056-route-caching.md
ascorbic: I wonder if we just directly copy the Nitro pattern, with "swr: true|number is shortcut for cache: { swr: true, maxAge: number }"

florian-lefebvre: A recurring problem with supporting different shapes it doesn't work very nicely with `updateConfig()`. Otherwise no strong opinion

ascorbic:
@florian-lefebvre I have done an update that among other things moves the rules from `cache.routes` to `routeRules`

serhalp:
👋🏼 This sounds good from the Netlify side

:shipit:

One minor piece of feedback: the semantics of `invalidate()` could be more precisely specified: Is it intended to be a "hard purge" or a "soft purge"? (I assume soft.) In the case of a stale-while-revalidate cache entry, does `invalidate()` imply a background _re_validation? (I assume no.)

PringlePot:
This is a great feature, however to improve the experience of creating custom providers it would be nice to publish some helper functions. This would prevent developers from having to re-implement some of the features that the memory provider currently has, such as cache key normalization, and SWR.

ascorbic:
@withastro/tsc this is my call for consensus on the RFC

matthewp (approved):

proposals/0056-route-caching.md
```diff
@@ -0,0 +1,1406 @@
...
+  To use this template: create a new, empty file in the repo under `proposals/${ID}.md`.
+  Replace `${ID}` with the official accepted proposal ID, found in the GitHub Issue
+  of the accepted proposal.
+-->
+
+**If you have feedback and the feature is released as experimental, please leave it on the Stage 3 PR. Otherwise, comment on the Stage 2 issue (links below).**
+
+- Start Date: 2025-10-15
+- Reference Issues: https://github.com/withastro/roadmap/discussions/1131, https://github.com/withastro/roadmap/discussions/181
+- Implementation PR: <!-- leave empty -->
```
florian-lefebvre:
```suggestion
- Implementation PR: https://github.com/withastro/astro/pull/15579
```

florian-lefebvre (approved):

ematipico (approved):

ascorbic:
This proposal has reached consensus.

Issue #1335, opened 2026-03-31
Configurable loggers / json logging


ematipico:
- Accepted Date: 2026-03-31
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/908
- Author: @rburgst
- Champion(s): @ematipico
- Implementation PR: N/A

# Summary

Add support for configurable log handlers in Astro, allowing users to replace the default console output with custom logging implementations (e.g. structured JSON).

# Background & Motivation

Users deploying Astro SSR to production need structured logging for integration with log aggregation services like Kibana, Logstash, CloudWatch, and Grafana/Loki. Agents running builds and dev locally work better with JSON. Today, Astro's logger is hardcoded – `BaseApp` uses `consoleLogDestination` which cannot be replaced, even by adapter authors. Error logs in rendering are split across multiple lines and formatted for human reading, making them unparseable by log aggregation tools.

This is a common requirement for any production SSR deployment and has been widely requested for a long time (this is currently the most upvoted feature request).

# Goals

- Allow users to configure a custom log handler that receives all Astro log output
- Support structured logging (JSON) for production deployments out of the box
- Provide a `--json` CLI flag for zero-config JSON output (e.g. `astro build --json`, `astro dev --json`)
- Follow the established `{ entrypoint, config }` factory pattern used by session drivers, font providers etc
- Work across all scopes: build, dev, and SSR runtime
- Ensure Vite logs flow through the custom handler (they already route through Astro's logger)
- Ship under `experimental.logger` initially

# Non-Goals 

- Request-scoped child loggers with correlation IDs
- Replacing the `debug` package integration – separate concern
- Separate handlers per scope (build vs server) – single handler with platform detection if needed

matthewp:
Do we think people want custom loggers or do they really just want a JSON logger? It would of course be much easier for us to just add JSON output than come up with a customizable API.

ematipico:
> Do we think people want custom loggers or do they really just want a JSON logger? It would of course be much easier for us to just add JSON output than come up with a customizable API.

I believe we must provide a custom logger or at least a destination that they can control in SSR. A destination that they can control would allow users to use redaction rules, like [`pino-noir`](https://github.com/pinojs/pino-noir) does

ematipico:
Feature advanced to Stage 3  https://github.com/withastro/roadmap/pull/1339

PR #1339, opened 2026-04-08
Configurable loggers / json logging

ematipico:
# Summary

```ts
// astro.config.mjs
import { defineConfig, logHandlers } from "astro/config";

export default defineConfig({
  experimental: {
    logger: logHandlers.json({ pretty: true }),
  },
});
```

```shell
astro dev --json
```

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/logger/proposals/0059-custom-logger.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/1335)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/908)

### DIFF
--- a/proposals/0059-custom-logger.md
+++ b/proposals/0059-custom-logger.md
@@ -146,22 +146,20 @@ export default defineConfig({
 });
 ```
 
-Astro will export a `defineDestination` type-safe function where users can create their custom destination:
+Library authors can use the `AstroLoggerDestination` interface if they need to create custom loggers:
 
 ```ts
 // astro-pino-logger/src/handler.ts
-import { defineDestination } from "astro/logger";
-import type { AstroLogMessage } from "astro";
+import type { AstroLoggerMessage, AstroLoggerDestination } from "astro";
 
 interface PinoConfig {
   destination?: string;
   level?: string;
 }
 
-export default defineDestination<PinoConfig>((config) => {
-  const pino = require("pino")(config);
+const pinoLogger = (config: PinoConfig = {}): AstroLoggerDestination => {
   return {
-    write(event: AstroLogMessage) {
+    write(event: AstroLoggerMessage) {
       pino[event.level]({ label: event.label }, event.message);
     },
     flush() {
@@ -172,7 +170,9 @@ export default defineDestination<PinoConfig>((config) => {
       pino.destination?.end();
     },
   };
-});
+}
+
+export default pinoLogger;
 ```
 
 The `AstroLoggerDestination` interface will be defined as follows:
--- a/proposals/0059-custom-logger.md
+++ b/proposals/0059-custom-logger.md
@@ -17,7 +17,7 @@ comment on the Stage 2 issue (links below).**
 - Reference Issues: <!-- related issues, otherwise leave empty -->
 - Implementation PR: <!-- leave empty -->
 - Stage 2 Issue: https://github.com/withastro/roadmap/issues/1335
-- Stage 3 PR: <!-- leave empty -->
+- Stage 3 PR: https://github.com/withastro/roadmap/pull/1339
 
 # Summary
 
--- a/proposals/0059-custom-logger.md
+++ b/proposals/0059-custom-logger.md
@@ -1,0 +1,324 @@
+<!--
+  Note: You are probably looking for `stage-1--discussion-template.md`!
+  This template is reserved for anyone championing an already-approved proposal.
+
+  Community members who would like to propose an idea or feature should begin
+  by creating a GitHub Discussion. See the repo README.md for more info.
+
+  To use this template: create a new, empty file in the repo under `proposals/${ID}.md`.
+  Replace `${ID}` with the official accepted proposal ID, found in the GitHub Issue
+  of the accepted proposal.
+-->
+
+**If you have feedback and the feature is released as experimental, please leave it on the Stage 3 PR. Otherwise,
+comment on the Stage 2 issue (links below).**
+
+- Start Date: 2026-04-07
+- Reference Issues: <!-- related issues, otherwise leave empty -->
+- Implementation PR: <!-- leave empty -->
+- Stage 2 Issue: https://github.com/withastro/roadmap/issues/1335
+- Stage 3 PR: <!-- leave empty -->
+
+# Summary
+
+Add support for configurable log handlers in Astro, allowing users to replace the default console output with custom
+logging implementations (e.g. structured JSON).
+
+# Example
+
+Using a built-in handler to get JSON logs:
+
+```js
+// astro.config.mjs
+import { defineConfig, logHandlers } from "astro/config";
+
+export default defineConfig({
+  experimental: {
+    logger: logHandlers.json({ pretty: false }),
+  },
+});
+```
+
+Using a third-party handler:
+
+```js
+// astro.config.mjs
+import { defineConfig } from "astro/config";
+import { pino } from "astro-pino-logger";
+
+export default defineConfig({
+  experimental: {
+    logger: pino({ destination: "/var/log/app.log" }),
+  },
+});
+```
+
+# Background & Motivation
+
+Users deploying Astro SSR to production need structured logging for integration with log aggregation services like
+Kibana, Logstash, CloudWatch, and Grafana/Loki. Agents running builds and dev locally work better with JSON. Today,
+Astro's logger is hardcoded – `BaseApp` uses `consoleLogDestination` which cannot be replaced, even by adapter authors.
+Error logs in rendering are split across multiple lines and formatted for human reading, making them unparseable by log
+aggregation tools.
+
+This is a common requirement for any production SSR deployment and has been widely requested for a long time (this is
+currently the most upvoted feature request).
+
+# Goals
+
+- Allow users to configure a custom log handler that receives all Astro log output
+- Support structured logging (JSON) for production deployments out of the box
+- Provide a `--json` CLI flag for zero-config JSON output (e.g. `astro build --json`, `astro dev --json`, `astro preview --json`)
+- Follow the established `{ entrypoint, config }` factory pattern used by session drivers, font providers etc
+- Work across all scopes: build, dev, and SSR runtime
+- Ensure Vite logs flow through the custom handler (they already route through Astro's logger)
+
+# Non-Goals
+
+- Request-scoped child loggers with correlation IDs
+- Replacing the `debug` package integration – separate concern
+- Separate handlers per scope (build vs server) – single handler with platform detection if needed
+- Redaction capabilities, library authors are responsible for that
+
+# Detailed Design
+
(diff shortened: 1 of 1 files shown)


proposals/0059-custom-logger.md
```diff
@@ -0,0 +1,324 @@
...
+- Allow third-party logging systems to use their `flush` and `close` APIs.
+- SSR, where the file descriptors of the destinations must be closed. Particularly:
+  - `flush` is needed in serverless environments (Vercel, Netlify, Cloudflare, etc.) so that logs are written to the
+    destination at the end of a rendering request, without closing the file descriptor.
+  - `close` is needed in server environments (Node.js, Bun, Deno, etc.) so that logs are written to the destination
+    when a `SIGTERM`/`SIGINT` signal is sent by the server, and the file descriptor can be closed safely.
+
+## Runtime
+
+The `Astro` global and `APIContext` type will expose a new `log` field that can be used as follows:
```
florian-lefebvre, line 202: Should it be a "global" virtual module instead? What are the pros and cons?

proposals/0059-custom-logger.md
```diff
@@ -0,0 +1,324 @@
...
+
+```js
+import pino from "astro-pino-logger";
+
+export default defineConfig({
+  logger: pino({ destination: "/var/log/app.log" }),
+});
+```
+
+Astro will export a `defineDestination` type-safe function where users can create their custom destination:
```
florian-lefebvre: I think it would make more sense to just have type. For all these kinds of APIs we chose to use a type instead, eg.:

```ts
export default function(config: PinoConfig): LogDestination {
	return {
		// ...
	}
}
```

proposals/0059-custom-logger.md
```diff
@@ -0,0 +1,324 @@
...
+-->
+
+**If you have feedback and the feature is released as experimental, please leave it on the Stage 3 PR. Otherwise,
+comment on the Stage 2 issue (links below).**
+
+- Start Date: 2026-04-07
+- Reference Issues: <!-- related issues, otherwise leave empty -->
+- Implementation PR: <!-- leave empty -->
+- Stage 2 Issue: https://github.com/withastro/roadmap/issues/1335
+- Stage 3 PR: <!-- leave empty -->
```
florian-lefebvre:
```suggestion
- Stage 3 PR: https://github.com/withastro/roadmap/pull/1339
```

proposals/0059-custom-logger.md
ematipico: As explained down belong, logging is almost purely runtime, so I don't see the value of a virtual module. Plus, I really don't like `import { info} from "astro:logger"`

florian-lefebvre: For the record, I'm asking this because having it on the context makes it restricted to the request. Just wanna make sure it's intentional. I think having it as a virtual module or the context doesn't make much difference in terms of implementation so there's no real blocker

ematipico: It is intentional. Read the part of the `flush` function, below

proposals/0059-custom-logger.md
```diff
@@ -0,0 +1,324 @@
...
+
+# Adoption strategy
+
+- The feature ships behind `experimental.logger`.
+- Alternatively, users can use JSON logging via `--experimental-json`.
+- After some testing time, the experimental flag will be removed, and the new feature will be shipped in a minor
+
+# Unresolved Questions
+
+- `flush` and `close` are asynchronous functions, and in the presence of multiple handlers we might need to await them. Should we call them without await? I wouldn't want to delay rendering. How do we catch errors if they fail?
```
florian-lefebvre, line 364: I imagine an adapter could provide something like `waitUntil` which doesn't block the request, otherwise block the request

ematipico: Possibly, but probably we can forget of the flush function. We will revisit once `waitUntil` lands in the next minor

proposals/0059-custom-logger.md
```diff
@@ -0,0 +1,367 @@
...
+
+export default defineConfig({
+  logger: logHandlers.console({ 
+    pretty: true,
+    level: "warn"
+  }),
+});
+```
+
+### `logHandlers.node(config?)`
```
matthewp, line 283: From the user perspective, is there any real difference between the console and node loggers? They do the same, right, one just only works in Node, in which case, why have it?

proposals/0059-custom-logger.md
```diff
@@ -0,0 +1,367 @@
...
+
+export default defineConfig({
+  logger: logHandlers.node({ 
+    pretty: true,
+    level: "warn"
+  }),
+});
+```
+
+### `logHandlers.compose(...LogHandlerConfig[])`
```
matthewp, line 298: The implementation PR had this as taking an array, not args, which is right?

proposals/0059-custom-logger.md
matthewp: I agree here, the `defineDestination` doesn't seem to add enough value for this niche use-case.

proposals/0059-custom-logger.md
```diff
@@ -0,0 +1,367 @@
...
+      pino.destination?.end();
+    },
+  };
+});
+```
+
+The `LoggerDestination` interface will be defined as follows:
+
+```ts
+export interface LoggerDestination<T> {
```
matthewp: What is `T` here? From the example it seems like it's always the `AstroLogMessage`?

proposals/0059-custom-logger.md
ematipico: They don't do the same thing.

Node.js uses stdout/stderr to print the messages.

Console prints to `console` via `info`/`warn`/`error`

proposals/0059-custom-logger.md
ematipico: The example was incorrect. I will fix it

proposals/0059-custom-logger.md
matthewp: Console always prints to stdout/stderr though. Do you know if there's an advantage to having both? Or did you do it that way because we already had both in the codebase (which IIRC is just historical and not purposefully designed).

proposals/0059-custom-logger.md
```diff
@@ -0,0 +1,367 @@
...
+Where the configuration has the following shape:
+
+```ts
+export interface LogHandlerConfig {
+  entrypoint: string | URL;
+  config?: Record<string, any>;
+}
+```
+
+Internally, Astro will use Vite to resolve the entrypoint. We will probably use the same instance of the configuration (
```
matthewp: What is the reason for Astro loading the logger? For example with Vite you can provide a [customLogger](https://vite.dev/config/shared-options#customlogger) but it's just a set of methods.

I assume we do it this way because we use the logger at runtime. Is that true?

proposals/0059-custom-logger.md
ematipico: > Console always prints to stdout/stderr though

Do you mean that the runtime (Node.js, workderd, etc.) does that? If so, then it's an implementation detail. 

> Or did you do it that way because we already had both in the codebase (which IIRC is just historical and not purposefully designed).

Yes we do. CLI and Pipeline use two different loggers. CLI uses a node.js logger, while the pipeline uses the console logger. Additionally, I didn't exclude the idea that a user could create a **client** logger, so I wanted to keep things open ended for now.

proposals/0059-custom-logger.md
ematipico: Reworded. It doesn't use vite anymore, as we agreed before during our meeting

ematipico:
Call for consensus is now open.

ascorbic:
👍🏻  looks good

matthewp:
👍 sounds good to me

proposals/0059-custom-logger.md
```diff
@@ -0,0 +1,365 @@
...
+  Replace `${ID}` with the official accepted proposal ID, found in the GitHub Issue
+  of the accepted proposal.
+-->
+
+**If you have feedback and the feature is released as experimental, please leave it on the Stage 3 PR. Otherwise,
+comment on the Stage 2 issue (links below).**
+
+- Start Date: 2026-04-07
+- Reference Issues: <!-- related issues, otherwise leave empty -->
+- Implementation PR: <!-- leave empty -->
```
florian-lefebvre:
```suggestion
- Implementation PR: https://github.com/withastro/astro/pull/16477
```

florian-lefebvre (approved):

ematipico:
@matthewp and @ascorbic please *approve* the PR :)

matthewp (approved):

delucis:
Just used a user query to dogfood this API — really handy way for people to customise stuff!

Here’s the custom logger I made in case it’s of interest: https://stackblitz.com/edit/github-cf392aa7?file=astro.config.mjs,custom-logger.js
And the Discord thread: https://discord.com/channels/830184174198718474/1509285901791006942

Some quick feedback/ideas based on trying it:

- It would be nice to have a `defineLogger()` utility similar to our other type helpers. That would simplify some of the docs examples a bit potentially by not requiring as much manual typing.
- It feels a shame that the options to a logger aren’t safely typed — definitely tricky because of the entrypoint pattern though.
- The `"SKIP_FORMAT"` label might need to be documented or redesigned? Feels unexpected to have the label be used for something that isn’t actually a display label and likely for users to end up logging it instead of filtering it out like Astro’s implementations do.
- I ended up copying Astro’s `getEventPrefix()` logic to be able to replicate Astro’s behaviour — I basically wanted a logger that was 1:1 to Astro’s default behaviour but had custom control over skipping some messages. Don’t know if that suggests there are more utilities that could be useful to users.

Take all that with a grain of salt as just the experience for a specific kind of logger. But thought it would be helpful to share.

florian-lefebvre:
We decided against having `defineLogger()` because all similar APIs just use types FYI.

For the options type safety, that's good feedback. I think the RFC and the [docs on custom loggers](https://docs.astro.build/en/reference/experimental-flags/logger/#custom-loggers) should be updated to match the official pattern:

```ts
function pinoLogger(config: Options): LoggerHandlerConfig {
	return {
		entrypoint: '...',
		config
	}
}
```

We have a similar situation for session drivers. You don't have to use this pattern but we only document this one because it's better DX

delucis:
Oh, and just remembered another thing that tripped me up getting started:

The docs show adding a custom logger from an npm package entry point

```js
      entrypoint: "@org/custom-logger",
```

But most often people will be adding a local file to their project when building a custom logger, I’d guess. Or at least would do so initially before bundling it up to be published. It took me quite a bit of trial and error to find a way to specify a local file that worked in all contexts, ending up with:

```js
      entrypoint: fileURLToPath(new URL('./custom-logger.js', import.meta.url)),
```

That is definitely not super friendly — for most our APIs, e.g. like `injectRoute()`, we support passing a relative path like `'./custom-logger.js'`. Would be good to support that here too.

Issue #1342, opened 2026-04-16
Advanced Routing


matthewp:
- Accepted Date: 2026-04-16
- Reference Issues/Discussions: https://github.com/withastro/roadmap/discussions/1320
- Author: @matthewp 
- Champion(s): @matthewp
- Implementation PR: 

# Summary

This proposal introduces optional `src/app.ts` support as Astro's request entrypoint, giving users full control over the request pipeline using a standard fetch-based API. Astro's rendering and features are composed within this entrypoint. A Hono app can also be exported directly since it shares the same `{ fetch }` shape, and Hono middleware support is provided as an add-on.

# Background & Motivation

Today, Astro controls the entire request pipeline internally. Users can hook into it with middleware, but they can't control when things run relative to each other or relative to Astro itself. In practice this means:

- No explicit request ordering. If you need auth to run before Astro rendering, and logging to run after, you're relying on implicit middleware ordering that Astro controls. There's no way to say "run this, then Astro, then that."
- Features are tightly coupled to Astro internals. Middleware, Actions, rewrites, and i18n are all baked into Astro's request handling. They can't be used independently, reordered, or replaced — they run when and how Astro decides.
- Hard to integrate non-Astro request logic. If you want to mount an API from another library, add rate limiting, or run platform-specific logic before Astro sees the request, you're working against the framework rather than with it.

`src/app.ts` gives users an explicit entrypoint where they control the full request lifecycle. Astro features like Actions or redirects can be extracted into standalone packages that users compose themselves.

# Goals

- Introduce optional `src/app.ts` as a first-class request entrypoint.
- Use a standard `fetch`-based API as the contract, allowing users to control the request pipeline with plain `Request`/`Response` objects. Since a Hono app exposes the same `{ fetch }` shape, exporting a Hono app works naturally.
- Provide Hono middleware as an add-on (`astro/hono`) for users who want to use Astro features within a Hono middleware chain.
- Keep current behavior unchanged when `src/app.ts` is not present.
- Move existing Astro request features like Astro middleware, rewrites, Actions, into entrypoint modules (`astro/rewrites` for example) `src/app.ts`.

# Non-Goals

- Replacing Astro's adapters with third-party framework adapters. This could be a long-term goal to make Astro more maintainable but is a lot more work than the `src/app.ts` idea discussed here.

# Open questions

- How should state be shared across Astro features in `src/app.ts`? Options include a mutable state object (like `AstroRequest` above), storing state on the request object, or something else.
[...]

PR #1344, opened 2026-04-21
Advanced Routing

matthewp:
# Summary

Provide a `src/app.ts` to allow full control over Astro rendering pipeline.

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/advanced-routing-stage-3/proposals/0056-advanced-routing.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/1342)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/1320)

### DIFF
--- a/proposals/0056-advanced-routing.md
+++ b/proposals/0056-advanced-routing.md
@@ -166,7 +166,7 @@ The user creates the `FetchState` when using the `astro/fetch` API. The expected
 - `request` - The current `Request` object.
 - `response` - The `Response` produced by handlers, if any.
 - `routeData` - The matched route information.
-- `pathname` - The resolved pathname for the request.
+- `pathname` - The resolved pathname for the request. During a static build this may differ from the URL pathname — for example, it may include a `.html` extension to match the output file.
 - `locals` - The `App.Locals` object, available to middleware and endpoints via `ctx.locals` / `Astro.locals`.
 - `cookies` - The `AstroCookies` instance for reading/writing cookies.
 
@@ -209,9 +209,9 @@ interface ContextProvider<T> {
 ```
 The API on `FetchState`:
 
-- `provide<T>(key, provider)` — Registers a provider. The `create` factory is deferred until the first time accessed.
-- `resolve<T>(key)` — Lazily calls `create()`, caches the result, and returns it. Returns `undefined` if no provider was registered.
-- `finalizeAll()` — Runs all registered `finalize` callbacks for providers that were actually resolved. Returns synchronously when nothing needs finalizing.
+- `provide<T>(key, provider): void` — Registers a provider. Synchronous; the `create` factory is deferred until the first `resolve` call.
+- `resolve<T>(key): T | undefined` — Lazily calls `create()`, caches the result, and returns it. Synchronous. Returns `undefined` if no provider was registered.
+- `finalizeAll(): Promise<void> | void` — Runs all registered `finalize` callbacks for providers that were actually resolved. Returns `void` synchronously when nothing needs finalizing; returns a `Promise<void>` when any `finalize` callback is asynchronous. Callers should `await` the result.
 
 For example, the `sessions()` handler registers a provider that lazily creates an `AstroSession` and persists it on finalize:
 
@@ -227,6 +227,20 @@ state.provide<AstroSession>('session', {
 ```
 
 From the user's perspective they only need to add the `session()` handler/middleware and then they get `ctx.session` / `Astro.session` as values. This could also be used by 3rd party integrations.
+
+### Typing Context Providers
+
+Custom context providers are typed using the same pattern as `locals`: module augmentation. A library or user registers a provider with a string key, and the corresponding type is declared by extending an interface in `env.d.ts`:
+
+```ts
+declare namespace App {
+  interface Providers {
+    oauth: import('./lib/oauth').OAuthSession;
+  }
+}
+```
+
+With this declaration, `ctx.oauth` and `Astro.oauth` are properly typed. Built-in providers like `session` are pre-declared by Astro's own type definitions, so users don't need to augment those themselves.
 
 ## Feature Handlers
 
@@ -285,10 +299,11 @@ function middleware(
 ): Promise<Response>
 ```
 
-The `next` callback is where you put the inner handlers (typically `pages`):
+The `next` callback is where you put the inner handlers (typically `pages`). The `Response` returned by `next` is also set on `state.response`, so both the return value and the state property reference the same object:
 
 ```ts
 const response = await middleware(state, (s) => pages(s));
+// response === state.response
 ```
 
 ### `pages(state)`
@@ -318,6 +333,12 @@ function cache(
   state: FetchState,
   next: () => Promise<Response>,
 ): Promise<Response>
+```
+
+Unlike `middleware`, the `next` callback here does not receive `state` because `cache` only wraps the response — it does not need to forward state to an inner handler. The caller already has `state` in scope and can close over it:
+
+```ts
+const response = await cache(state, () => middleware(state, (s) => pages(s)));
 ```
 
 ### `astro(state)`
@@ -369,6 +390,8 @@ export default {
 };
 ```
 
+Note that `finalizeAll()` must be called by the user when composing handlers manually. Context providers like `sessions()` register lazy `finalize` callbacks (e.g. to persist session data), and the user controls when those run. This is intentional — the low-level API gives the user full control over the request lifecycle, including cleanup. The combined `astro()` handler calls `finalizeAll()` internally, so users who don't need fine-grained control don't have to think about it.
+
 The API is flexible, so it's possible to slot business logic anywhere. Add auth before `pages`, add logging around `middleware`, skip `i18n` entirely, or replace `pages` with your own rendering — it's all just function calls.
 
 ## Platform Entrypoints
@@ -381,7 +404,11 @@ The API is flexible, so it's possible to slot business logic anywhere. Add auth 
 
 `src/app.ts` does not replace platform entrypoints. Users who need platform-specific APIs (e.g. Durable Objects, queues) still use their platform's entrypoint for those concerns.
 
+Some platforms also provide edge middleware (e.g. Vercel Edge Middleware, Netlify Edge Functions) that runs before the request reaches the adapter. This layer is outside Astro's control and is not affected by this proposal — it continues to work as it does today.
+
 ## Hono API
+
+[Hono](https://hono.dev/) was chosen because it is built entirely on Web Standard APIs (`Request`/`Response`/`fetch`), which aligns with the fetch-handler model at the core of this proposal. It has the largest middleware ecosystem among fetch-native frameworks, giving users access to auth, CORS, rate limiting, and many other capabilities without Astro needing to build them. Hono is also lightweight — it is a middleware router in pure JavaScript with no server runtime of its own, so it adds no overhead beyond what the user's middleware does.
 
 A Hono-specific API of middleware will be provided as `astro/hono`. Hono is used here purely for its middleware composition capability — no server is started and no ports are bound. When a Hono app is exported from `src/app.ts`, Astro calls its `fetch` method directly with the incoming `Request`, the same way it would call any other fetch handler. Hono acts as a middleware router in pure JavaScript; all actual HTTP serving is handled by the adapter/platform layer above.
 
--- a/proposals/0056-advanced-routing.md
+++ b/proposals/0056-advanced-routing.md
@@ -195,7 +195,7 @@ export default {
 }
 ```
 
-Note that using the `astro/hono` API the creation of FetchState is not necessary, since Hono has its own Context object; handlers can get/create a FetchState by inspecting HonoContext for a certain key (likely `astro.fetchState`).
+Note that using the `astro/hono` API the creation of FetchState is not necessary, since Hono has its own Context object; handlers get/create a FetchState lazily via the `getFetchState(context)` function exported from `astro/hono`. This function is part of the public API, so custom Hono middleware and third-party packages can use it to access the same shared `FetchState` that the built-in Astro middleware wrappers use.
 
 ### Context Providers
 
@@ -414,7 +414,24 @@ Some platforms also provide edge middleware (e.g. Vercel Edge Middleware, Netlif
 
 A Hono-specific API of middleware will be provided as `astro/hono`. Hono is used here purely for its middleware composition capability — no server is started and no ports are bound. When a Hono app is exported from `src/fetch.ts`, Astro calls its `fetch` method directly with the incoming `Request`, the same way it would call any other fetch handler. Hono acts as a middleware router in pure JavaScript; all actual HTTP serving is handled by the adapter/platform layer above.
 
-The `astro/hono` exports are thin wrappers around the lower-level `astro/fetch` feature handlers. They store `FetchState` on Hono's context object and delegate directly to the underlying handlers.
+The `astro/hono` exports are thin wrappers around the lower-level `astro/fetch` feature handlers. They store `FetchState` on Hono's context object and delegate directly to the underlying handlers. The `getFetchState(context)` function is also exported, allowing custom Hono middleware and third-party packages to access the per-request `FetchState`:
+
+```ts
+import { Hono } from 'hono';
+import { getFetchState, pages } from 'astro/hono';
+
+const app = new Hono();
+
+app.use(async (context, next) => {
+  const state = getFetchState(context);
+  state.locals.user = await authenticate(context.req.raw);
+  await next();
+});
+
+app.use(pages());
+
+export default app;
+```
 
 User-facing API will look like:
 
--- a/proposals/0056-advanced-routing.md
+++ b/proposals/0056-advanced-routing.md
@@ -195,18 +195,146 @@ From the user's perspective they only need to add the `session()` handler/middle
 
 ## Feature Handlers
 
-As much as possible, the aim is to group distinct "features" of Astro into individual handlers. In the `astro/hono` API these will be provided as separate middleware. The following features are expected to be provided:
+Each Astro feature is exposed as a plain function from `astro/fetch`. These functions fall into three categories based on their return type:
 
-- trailingSlash - Controls enforcement of Astro `trailingSlash` config.
-- i18n
-- sessions
-- redirects
-- cache providers
-- pages folder - This is rendering pages and calling endpoints.
-- user middleware - This runs the user's `src/middleware.ts` if present. `src/middleware.ts` is still supported when `src/app.ts` is used; it becomes one of the composable handlers in the pipeline rather than being automatically injected.
-- actions
+- **Early-exit handlers** return `Response | undefined`. If they return a `Response`, the request is handled and the caller should stop. If `undefined`, the caller continues to the next handler.
+- **Wrapping handlers** take a `next` callback and always return a `Response`. They run logic before and/or after the inner handler.
+- **Post-processors** take an existing `Response` and return a (possibly modified) `Response`.
 
-Additionally there will be an `astro()` handler that combines all of the features into a single API. Expectation is that most people who don't want fine-grained control will use this API.
+All handlers take a `FetchState` as their first argument. Here is each handler, what it does, and its signature:
+
+### `trailingSlash(state)`
+
+Enforces the `trailingSlash` config option (`'always'`, `'never'`, or `'ignore'`). If the request pathname doesn't match the configured policy, returns a `301` redirect to the corrected URL. Otherwise returns `undefined`. Call it first, before any other handler.
+
+```ts
+function trailingSlash(state: FetchState): Response | undefined
+```
+
+### `redirects(state)`
+
+Checks if the matched route is a redirect configured via `astro.config`. If so, returns the redirect `Response`. Otherwise returns `undefined`.
+
+```ts
+function redirects(state: FetchState): Promise<Response> | undefined
+```
+
+### `actions(state)`
+
+Handles Astro Action requests. For RPC actions (JSON API calls via `actions.myAction()`), executes the action and returns a `Response` containing the result. For form actions (progressive enhancement via `<form>`), executes the action and stores the result on the state so the page can read it via `Astro.getActionResult()`, then returns `undefined` to let page rendering continue. For non-action requests, returns `undefined`.
+
+```ts
+function actions(state: FetchState): Promise<Response | undefined> | undefined
+```
+
+### `sessions(state)`
+
+Registers the session provider on the state. The session object is created lazily — only when user code accesses `ctx.session` or `Astro.session`. No-op if sessions are not configured in `astro.config`.
+
+Call this early, before middleware runs. After the response is produced, call `state.finalizeAll()` in a `finally` block to persist any session mutations.
+
+```ts
+function sessions(state: FetchState): Promise<void> | void
+```
+
+Returns `undefined` when sessions are not configured.
+
+### `middleware(state, next)`
+
+Runs the user's `src/middleware.ts` (if present) around the `next` callback. The user's `onRequest` handler receives the usual `APIContext` and a `next` function. If no user middleware is configured, calls `next` directly.
+
+```ts
+function middleware(
+  state: FetchState,
+  next: (state: FetchState) => Promise<Response>,
+): Promise<Response>
+```
+
+The `next` callback is where you put the inner handlers (typically `pages`):
+
+```ts
+const response = await middleware(state, (s) => pages(s));
+```
+
+### `pages(state)`
+
+Renders the matched page or endpoint and returns the response. This is the core rendering handler — it loads the route's component module, executes it, and builds the response. Returns 404/500 error pages when appropriate.
+
+```ts
+function pages(state: FetchState): Promise<Response>
+```
+
+### `i18n(state, response)`
+
+Post-processes a response against the app's i18n configuration. Handles locale redirects for the root URL, 404 responses for invalid locales, and fallback routing. Returns the response unmodified if i18n is not configured or the strategy is `'manual'`.
+
+```ts
+function i18n(state: FetchState, response: Response): Promise<Response>
+```
+
+Call this after the response has been produced (typically by `middleware` + `pages`).
+
+### `cache(state, next)`
+
+Wraps a render callback with cache provider logic. Handles runtime caching (providers that implement `onRequest`), CDN-based providers (headers only), and the no-cache case transparently. Cache headers are applied and stripped internally.
+
+```ts
+function cache(
+  state: FetchState,
+  next: () => Promise<Response>,
+): Promise<Response>
+```
+
+### `astro(state)`
+
+The combined handler that runs all features in the default order. Equivalent to composing every handler above in the standard sequence: trailing slash → redirects → sessions → actions → cache → middleware → pages → i18n, with error handling and finalization.
+
+```ts
+function astro(state: FetchState): Promise<Response>
+```
+
+Most users who don't need fine-grained control should use this.
+
+### Composing handlers
+
+Here's how the individual handlers compose into a full pipeline:
+
+```ts
+import {
+  FetchState, trailingSlash, redirects,
+  sessions, actions, middleware, pages, i18n
+} from 'astro/fetch';
+
+export default {
+  async fetch(request: Request) {
+    const state = new FetchState(request);
+
+    // Early exits — return immediately if they handle the request.
+    const slash = trailingSlash(state);
+    if (slash) return slash;
+
+    const redirect = redirects(state);
+    if (redirect) return redirect;
+
+    // Register session provider (lazy, no-op if unconfigured).
+    sessions(state);
+
+    const action = await actions(state);
+    if (action) return action;
+
+    try {
+      // Middleware wraps page rendering; i18n post-processes.
+      const response = await middleware(state, () => pages(state));
+      return await i18n(state, response);
+    } finally {
+      // Persist sessions, etc.
+      await state.finalizeAll();
+    }
+  }
+};
+```
+
+The power of this API is that you can slot your own logic anywhere. Add auth before `pages`, add logging around `middleware`, skip `i18n` entirely, or replace `pages` with your own rendering — it's all just function calls.
 
 ## Platform Entrypoints
 
--- a/proposals/0056-advanced-routing.md
+++ b/proposals/0056-advanced-routing.md
@@ -195,7 +195,7 @@ export default {
 }
 ```
 
-Note that using the `astro/hono` API the creation of `FetchState` is not necessary, since Hono has its own Context object; the `astro/hono` handler wrappers automatically create and store a `FetchState` on the Hono context under the `"fetchState"` key. You can access it directly via `context.get("fetchState")`.
+Note that using the `astro/hono` API the creation of FetchState is not necessary, since Hono has its own Context object; handlers can get/create a FetchState by inspecting HonoContext for a certain key (likely `astro.fetchState`).
 
 ### Context Providers
 
--- a/proposals/0056-advanced-routing.md
+++ b/proposals/0056-advanced-routing.md
@@ -135,7 +135,16 @@ class RedirectsHandler {
 }
 ```
 
-- **Post-processors** run after the main response has been produced and may mutate it (e.g. i18n rerouting unhandled requests). These expose a method like `finalize()`.
+- **Post-processors** take an existing `Response` and return a modified one. They can be used at any point, but typically run after page rendering. For example, the i18n handler post-processes the response from `pages()` to handle locale redirects and fallbacks:
+
+```ts
+class I18nHandler {
+  finalize(state: FetchState, response: Response): Response {
+    // ...
+  }
+}
+```
+
 - **Helpers** provide checks or utilities that other handlers use, without directly producing or modifying responses.
 
 This proposal doesn't aim to enforce a single interface across all handlers, as the methods vary depending on what each feature needs to do.
--- a/proposals/0056-advanced-routing.md
+++ b/proposals/0056-advanced-routing.md
@@ -1,0 +1,270 @@
+- Start Date: 2026-04-21
+- Reference Issues: https://github.com/withastro/roadmap/discussions/1320
+- Implementation PR: https://github.com/withastro/astro/pull/16366
+- Stage 2 Issue: https://github.com/withastro/roadmap/issues/1342
+- Stage 3 PR: <!-- leave empty -->
+
+# Summary
+
+Provides greater control over the request lifecycle in Astro as well as the ability to use server frameworks that use the Request/Response APIs such as [Hono](https://hono.dev/), allowing the user to inject logic between Astro features such as i18n, redirects, rewrites, sessions, and even the pages directory.
+
+# Example
+
+The minimal API for this proposal is a `src/app.ts` file with the following shape:
+
+```ts
+export default {
+  fetch(request: Request) {
+    return new Response('ok', {
+      status: 200
+    });
+  }
+}
+```
+
+In order to be useful, we also provide ways to call into Astro's request handling features, one through a low-level module `astro/fetch` with plain functions that handle requests, and through a higher-level `astro/hono` which provides middleware for use with Hono apps.
+
+__astro/fetch__
+
+This is the low-level API that includes request handlers for various Astro features. The user has the ability to call them in any order.
+
+```ts
+import { FetchState, redirects, pages } from 'astro/fetch';
+
+export default {
+  fetch(request: Request) {
+    const state = new FetchState(request);
+
+    // Check redirects first
+    const redirect = redirects(state);
+    if (redirect) return redirect;
+
+    // Render pages
+    return pages(state);
+  }
+}
+```
+
+For users who don't need fine-grained control, a combined `astro()` handler runs all features in the default order:
+
+```ts
+import { FetchState, astro } from 'astro/fetch';
+
+export default {
+  fetch(request: Request) {
+    const state = new FetchState(request);
+    return astro(state);
+  }
+}
+```
+
+__astro/hono__
+
+This is the higher-level API targeting Hono specifically. Features are provided as Astro middleware and can be mixed and matched.
+
+```ts
+import { astro } from 'astro/hono';
+import { Hono } from 'hono';
+
+const app = new Hono();
+
+app.use(astro());
+
+export default app;
+```
+
+# Background & Motivation
+
+Astro started as a static site generator with simple file-based routing. As use-cases for Astro grew we needed to add new ways to configure routing; redirects through Astro config, middleware through a special `src/middleware.ts` file, Actions through a special file, i18n through config (or middleware).
+
+Despite this, users run into limitations with the current approach:
+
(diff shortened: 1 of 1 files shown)


proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+The minimal API for this proposal is a `src/app.ts` file with the following shape:
+
+```ts
+export default {
+  fetch(request: Request) {
+    return new Response('ok', {
+      status: 200
+    });
+  }
+}
```
ematipico, line 22: I think it's fine to keep the example like this, however I think we should expose a `defineFetch` (or similar) utility like we do for actions (`defineAction`), middleware (`defineMiddleware`) and so on. This will improve type-safety. For examples, things might not be clear:
- `fetch` is `async` by design
- `fetch` must return a `Response`

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+- **Hard to integrate non-Astro request logic.** If you want to mount an API from another library, add rate limiting, or run platform-specific logic before Astro sees the request, you're working against the framework rather than with it.
+- **Not all requests reach user code.** Requests that do not match a RouteData object never get passed through Astro's pipeline or reach user middleware, leading users to add logic in front of Astro through their platform's own APIs.
+
+This proposal aims to consolidate a single pipeline for Astro request handling where all requests go into, and the user has complete control.
+
+# Goals
+
+- Allow developers to have complete control over requests coming into their Astro application.
+- Provide an API that is completely compatible with Fetch handlers, such as Hono, so users can gain the benefit of those ecosystems.
+- Break individual Astro features into small APIs that can be composed however the user wishes.
```
ematipico, line 93: I think it's important to state if the goal is to have this feature working for SSG, SSR, both, or some of them.

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+- Removing any features; this is just a different way to control what order features run. Features such as injectRoute will still exist, even if it would be possible to implement similar functionality through Hono middleware.
+- Provide ways for integrations to inject code into the pipeline. The user has complete control over the request pipeline in this proposal; in fact they have to include the right middleware to even render pages.
+
+# Detailed Design
+
+Currently most of the logic to resolve what gets called during a request is contained within the App class. This class doubles as the external API used by adapters to render pages.
+
+The bulk of the changes for this proposal will be to extract the logic out of the App class and into feature-specific handler classes. Each feature is organized as a class, but the API of each handler varies depending on what it does. There are a few common patterns:
+
+- **Request interceptors** check the request and may return a response early (e.g. redirects). These have a `handle(state: FetchState)` method that returns a `Response` or `undefined`:
```
ematipico, line 128: Two things need to be addressed imho:
- Calling it _request_ interceptor might be misleading because `handle` accepts a `FetchState`, not a `Request`. 
- It's important to highlight if `state` is a copy or not, and if `state` is meant to be mutable in an imperative way

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+
+```ts
+class RedirectsHandler {
+  handle(state: FetchState): Response | undefined {
+    // ...
+  }
+}
+```
+
+- **Post-processors** run after the main response has been produced and may mutate it (e.g. i18n rerouting unhandled requests). These expose a method like `finalize()`.
```
ematipico: nit: Add code snippet for the finalize function, like you did for `handle`. Also, explain what "main response" means from an architectural point of view. Where is it coming from? What creates it? etc.

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+- Stage 2 Issue: https://github.com/withastro/roadmap/issues/1342
+- Stage 3 PR: https://github.com/withastro/roadmap/pull/1344
+
+# Summary
+
+Provides greater control over the request lifecycle in Astro as well as the ability to use server frameworks that use the Request/Response APIs such as [Hono](https://hono.dev/), allowing the user to inject logic between Astro features such as i18n, redirects, rewrites, sessions, and even the pages directory.
+
+# Example
+
+The minimal API for this proposal is a `src/app.ts` file with the following shape:
```
ematipico: At this point I would call it `src/fetch.ts`. Everything orbits around a public-facing fetch APIs and concepts. The only time we mention "app", is when we mention the `App` class, which is and stays an internal implementation. 

Also, highlight the fact that the file can be any supported extension `.js`, `.mjs`, etc.

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+- **Helpers** provide checks or utilities that other handlers use, without directly producing or modifying responses.
+
+This proposal doesn't aim to enforce a single interface across all handlers, as the methods vary depending on what each feature needs to do.
+
+The user-facing API in `astro/fetch` wraps these classes as plain functions (e.g. `redirects(state)`, `pages(state)`). The `astro/hono` API wraps them further as Hono middleware. See the [Feature Handlers](#feature-handlers) and [Hono API](#hono-api) sections below for details.
+
+Internally, handler classes require configuration from Astro's SSR manifest (route table, i18n settings, etc.). To keep the user-facing API simple, the `astro/fetch` module imports the manifest via a Vite virtual module (`virtual:astro:manifest`) at build time and passes it into the handler constructors. This means users just call `redirects(state)` rather than needing to import and wire up the manifest themselves. The handler classes themselves remain pure and accept the manifest as a constructor argument, which makes them directly unit-testable without the virtual module.
```
ematipico, line 154: This is a very important detail that it might shape the APIs. All APIs depend on a vite virtual module, which means that they can't be used anywhere outside of `src/app.ts`, and require vite. Given that, I do wonder if this should be a virtual module `astro:fetch` at this point.

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+
+## FetchState
+
+Every request has additional state associated with it. Some of it is specific to the adapter such as the `clientAddress`, while other state is loaded during the request lifecycle, such as the `APIContext` object that is created to pass to user middleware.
```
ematipico:
```suggestion
Every request has additional state associated with it. Some of it is specific to the adapter such as the `clientAddress`, while other state is loaded during the request lifecycle, such as the `APIContext` object that is created to pass to user middleware or endpoint.
```

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+
+The `FetchState` object is the representation of this state and needs to be created by the user when using the `astro/fetch` API. The expected public surface includes:
+
+- `request` - The current `Request` object.
+- `response` - The `Response` produced by handlers, if any.
+- `routeData` - The matched route information.
+- `pathname` - The resolved pathname for the request.
```
ematipico: Important to highlight that `pathname` may differ in build e.g. `.html` extension

proposals/0056-advanced-routing.md
ematipico: Also explain if `state` is meant to be mutable or not.

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+
+Internally, handler classes require configuration from Astro's SSR manifest (route table, i18n settings, etc.). To keep the user-facing API simple, the `astro/fetch` module imports the manifest via a Vite virtual module (`virtual:astro:manifest`) at build time and passes it into the handler constructors. This means users just call `redirects(state)` rather than needing to import and wire up the manifest themselves. The handler classes themselves remain pure and accept the manifest as a constructor argument, which makes them directly unit-testable without the virtual module.
+
+## FetchState
+
+Every request has additional state associated with it. Some of it is specific to the adapter such as the `clientAddress`, while other state is loaded during the request lifecycle, such as the `APIContext` object that is created to pass to user middleware.
+
+The `FetchState` object is the representation of this state and needs to be created by the user when using the `astro/fetch` API. The expected public surface includes:
```
ematipico: I think I saw `locals` too during the API bash.

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+
+export default {
+  fetch(request: Request) {
+    const state = new FetchState(request);
+
+    // Run the redirects
+    const response = redirects(state);
+    if(response) {
+      return response;
+    }
```
ematipico, line 190: So, you mentioned that `state` might have a `response`. Given that `redirects` returned a `response`, I do wonder which one should be the source of truth.

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+Internally, handler classes require configuration from Astro's SSR manifest (route table, i18n settings, etc.). To keep the user-facing API simple, the `astro/fetch` module imports the manifest via a Vite virtual module (`virtual:astro:manifest`) at build time and passes it into the handler constructors. This means users just call `redirects(state)` rather than needing to import and wire up the manifest themselves. The handler classes themselves remain pure and accept the manifest as a constructor argument, which makes them directly unit-testable without the virtual module.
+
+## FetchState
+
+Every request has additional state associated with it. Some of it is specific to the adapter such as the `clientAddress`, while other state is loaded during the request lifecycle, such as the `APIContext` object that is created to pass to user middleware.
+
+The `FetchState` object is the representation of this state and needs to be created by the user when using the `astro/fetch` API. The expected public surface includes:
+
+- `request` - The current `Request` object.
+- `response` - The `Response` produced by handlers, if any.
```
ematipico: It's important to mention where a `response` is coming from.

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+  create: () => T;
+  /** Optional cleanup / persist callback. */
+  finalize?: (value: T) => Promise<void> | void;
+}
+```
+The API on `FetchState`:
+
+- `provide<T>(key, provider)` — Registers a provider. The `create` factory is deferred until the first time accessed.
+- `resolve<T>(key)` — Lazily calls `create()`, caches the result, and returns it. Returns `undefined` if no provider was registered.
+- `finalizeAll()` — Runs all registered `finalize` callbacks for providers that were actually resolved. Returns synchronously when nothing needs finalizing.
```
ematipico: Please state if these functions are `async` by design or not

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+  create() {
+    return new AstroSession({ cookies, config, ... });
+  },
+  finalize(session) {
+    return session.persist();
+  },
+});
+```
+
+From the user's perspective they only need to add the `session()` handler/middleware and then they get `ctx.session` / `Astro.session` as values. This could also be used by 3rd party integrations.
```
ematipico, line 231: How do types work here? Let's suppose I create a fetch handler **in my project**, which returns `new OAuthSession()` from the `create` function, with key `oauth`, how can I make sure that `ctx.oauth` is properly typed when implementing my endpoint/middleware?  

For example, for `locals` users need to extend the types themselves. How does it work in this case?

proposals/0056-advanced-routing.md
ematipico: How do users have access to the manifest/configuration?

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+
+### `i18n(state, response)`
+
+Post-processes a response against the app's i18n configuration. Handles locale redirects for the root URL, 404 responses for invalid locales, and fallback routing. Returns the response unmodified if i18n is not configured or the strategy is `'manual'`.
+
+```ts
+function i18n(state: FetchState, response: Response): Promise<Response>
+```
+
+Call this after the response has been produced (typically by `middleware` + `pages`).
```
ematipico:
```suggestion
This API requires a `Response`, you must call it after a fetch handler that returns a `Response` object, for example `middleware` or/and `pages`.
```

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+Returns `undefined` when sessions are not configured.
+
+### `middleware(state, next)`
+
+Runs the user's `src/middleware.ts` (if present) around the `next` callback. The user's `onRequest` handler receives the usual `APIContext` and a `next` function. If no user middleware is configured, calls `next` directly.
+
+```ts
+function middleware(
+  state: FetchState,
+  next: (state: FetchState) => Promise<Response>,
```
ematipico, line 300: Is this one of the cases where `state.response` is populated? If that's the case, is `response` the same response returned by the `next` function?

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+Call this after the response has been produced (typically by `middleware` + `pages`).
+
+### `cache(state, next)`
+
+Wraps a render callback with cache provider logic. Handles runtime caching (providers that implement `onRequest`), CDN-based providers (headers only), and the no-cache case transparently. Cache headers are applied and stripped internally.
+
+```ts
+function cache(
+  state: FetchState,
+  next: () => Promise<Response>,
```
ematipico, line 336: Why doesn't this callback accept `state`?

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+    if (action) return action;
+
+    try {
+      // Middleware wraps page rendering; i18n post-processes.
+      const response = await middleware(state, () => pages(state));
+      return await i18n(state, response);
+    } finally {
+      // Persist sessions, etc.
+      await state.finalizeAll();
+    }
```
ematipico, line 390: I think it's a bit inconsistent that `middleware` and/or `i18n` requires a try/catch, while the other functions don't, but I understand we need some error handling in place. 

The problem I see is that `finalizeAll` wasn't mentioned as something that users need to call, because `provide` and `resolve` are called either by Astro, or the library authors. 

In this case `finalizeAll` is called **by the end-user**, which feels inconsistent with the design of the context providers.

I see a few options:
- error handling should be controlled by the library authors
- error handling should be controlled by Astro
- users control error handling but low-level APIs are still called by us
- another solution that is more consistent with the design

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+  }
+};
+```
+
+The power of this API is that you can slot your own logic anywhere. Add auth before `pages`, add logging around `middleware`, skip `i18n` entirely, or replace `pages` with your own rendering — it's all just function calls.
```
ematipico:
```suggestion
The API is flexible, so it's possible to slot business logic anywhere. Add auth before `pages`, add logging around `middleware`, skip `i18n` entirely, or replace `pages` with your own rendering — it's all just function calls.
```

While I understand that this API is powerful, this comes with downsides. I suggested a more objective term, and removed the use of "you".

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+
+## Platform Entrypoints
+
+`src/app.ts` runs within Astro's request handling, after the adapter and any platform-specific entrypoint (e.g. Cloudflare's `worker.ts`). The layering is:
+
+1. **Adapter / platform entrypoint** (e.g. `worker.ts`) - Platform-specific logic, receives the raw platform request.
+2. **`src/app.ts`** - User's fetch handler, receives a standard `Request`.
+3. **Feature handlers** - Astro features like redirects, pages, etc.
```
ematipico, line 405: We need to take edge middleware into account too.

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+
+`src/app.ts` does not replace platform entrypoints. Users who need platform-specific APIs (e.g. Durable Objects, queues) still use their platform's entrypoint for those concerns.
+
+## Hono API
+
```
ematipico, line 412: I think it's important that this RFC answers the following questions:
- why was `hono` chosen?
- why having a direct `astro/hono` specifier, instead of `@astrojs/hono` package? - Considering that Astro won't depend on `hono`, there's no reason for having a separate package.

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,431 @@
...
+```ts
+import { redirects, actions, pages } from 'astro/hono';
+import { Hono } from 'hono';
+
+const app = new Hono();
+app.use(redirects());
+app.use(actions());
+app.use(pages());
+
+export default app;
```
ematipico, line 447: I would like to have a more in-depth paragraph of the functions exported from `astro/hono`. From the API bash, you mentioned that these functions are small wrappers around the functions exported from `astro/fetch`. If that's still the case, a paragraph should address the following concerns/questions:
- Since `src/app.ts` runs for prerendered pages too, does that mean we spawn an Hono server while rendering them? 
- If `redirects` is a small wrapper of `redirects` coming from `astro/fetch`, does that mean that it inherits the same restrictions? E.g. vite dependent, can't be used outside Astro **and** vite.

ematipico (commented):
It's a good proposal as a starter! Here's the summary of my review, which we should address (hopefully before experimental release): 
- It's possible we take this for granted, but the proposal doesn't mention how this new APIs function with SSR/prerendered pages. I assume it works like the middleware, but I might be wrong. I think it's worth write it down.
- I noticed some inconsistency or lack of information in the design of some APIs, for example `state.response` and context providers. I left some comments around that area.
- Error handling is completely absent, and I think we should have a paragraph that explain how would that work. E.g. there's a runtime error in `fetch` during the build, what happens? 
- While there's some hype around hono, the RFC doesn't say **why** Hono and why `astro/hono` instead of `@astrojs/hono`.

proposals/0056-advanced-routing.md
matthewp: I added an explanation of the shape and why it's that way here: [239213b](https://github.com/withastro/roadmap/pull/1344/commits/239213b02740d28f9edcb0e395d5b7f28918cdc1)

Added a use of `satisfies Fetchable` as that aligns with how other platforms do type safety.

proposals/0056-advanced-routing.md
matthewp: Yeah good point, I added [21edb9b](https://github.com/withastro/roadmap/pull/1344/commits/21edb9bf96931f7e66e3a08f2b9704f7149c7be4) to address this, the reason for `astro/fetch` is the idea that people will want to import the handlers to create higher-level handlers; for example Starlight might import all of them to do its own opinionated rendering logic. For that you want a real module if you're going to publish to node_modules/ while depending on this API.

proposals/0056-advanced-routing.md
matthewp: Added a new goal here to clarify it works the same in SSG and SSR.

proposals/0056-advanced-routing.md
matthewp: **Request interceptors** seems approach here because it's meant to describe the pattern and what these handlers do is intercept the request and return responses if the request matches what it is that they handle.

Remember that this is an explanation of the internal code, not public documentation. 

Also, I updated the text to explain that FetchState is mutable here: 6346abb

proposals/0056-advanced-routing.md
matthewp: Added this here: b2d9e9e

proposals/0056-advanced-routing.md
matthewp: Added that here: be415c7

proposals/0056-advanced-routing.md
matthewp: I'm going to see if we really need `response` on FetchState and if not, remove it.

proposals/0056-advanced-routing.md
ematipico: Right! Didn't think of that. I suppose we would need to document that users need vitest to test their fetch handler

millette:
Is https://elysiajs.com/ support planned or general 3rd party hooks to support it or are we *limited* to `astro/fetch` and `astro/hono` ?

florian-lefebvre:
You should be able to use `astro/fetch` to implement an elysia middleware. I don't think Astro will ship one for this in particular, given it's aimed at Bun (and it is still not officially supported)

princemuel:
Can we use other languages e.g golang, rust, elixir etc to implement the request pipeline?

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,472 @@
...
+import { handler } from 'astro/fetch';
+
+export class MyDurableObject { ... }
+
+export default {
+  fetch: handler
+}
+```
+
+This proposal does not currently address this use case, but it may be something we want to support as part of or alongside this work.
```
alexanderniebuhr, line 516: I would like to bring some more light into this. Right now it looks a bit too complicated for users who use platform-specific entrypoints and advanced-routing the same time, as well as us to maintain both. If I understand everything correctly a request flow looks like this

```
-> raw request -> worker.ts -> handler(packages/integrations/cloudflare/src/utils/handler.ts) -> transformed request -> app.ts ->
```

Which would lead to the following code:

1. `src/worker.ts`

```ts
import handler from "@astrojs/cloudflare/entrypoints/server";

export default {
	// raw request from the worker
	async fetch(request, env, ctx) {
		return handler.fetch(request, env, ctx);
	},
} satisfies ExportedHandler<Env>;
```

2. `src/app.ts`

```ts
import { FetchState, astro } from 'astro/fetch';

export default {
  fetch(request: Request) {
    // transformed request which comes from the handler in @astrojs/cloudflare/entrypoints/server
    const state = new FetchState(request);
    return astro(state);
  }
}
```

Decision Matrix (x means the file is required, if both columns have ?, one **or** the other can be used.):


| use-cases | needs worker.ts | needs app.ts |
|--------------|-----------------|--------------|
| platform-specific features e.g. durable objects | x | |
| custom routing order | | x |
| platform-specific features e.g. durable objects & custom routing order | x            | x       |
| custom fetch, e.g. `/api/*`  | ? | ? |
| access bindings directly  | ? | ? |
| use hono | ? | ? |

As you can see from the above matrix, there is not really any value for users to use advanced-routing other when they need to use a custom pipeline order. If they want to add custom handling based on a url `/api/*` they can use `worker.ts`. If they want to add or use hono, they can use `worker.ts`. If they want to access bindings directly they can use `workers.ts`. For all these use-cases they can also use `app.ts`, but they should prefer `worker.ts` since it's faster and allows for future usage of durable objects, without a new file. They only reason I can think of why they should use `app.ts` is if they need to change the order of astro's pipeline.

This in my opinion has two downsides, it makes this harder to document and explain for us when to use which, and we still need to keep and maintain `packages/integrations/cloudflare/src/utils/handler.ts`. In my opinion and in an ideally perfect world, `worker.ts` get's obsolete, and `app.ts` supports platform-specific entrypoints, including raw request handling, top-level exports e.g. Durable Objects, ideally so we can remove the `packages/integrations/cloudflare/src/utils/handler.ts` maintenance and keep documentation simple and lean. Users should have one file to reach for customizing everything, including platform-specific entrypoints as long as they follow the request/response design as well as astro pipeline.

florian-lefebvre: If the advanced routing entrypoint was configurable, it could point to the same file as wrangler's entrypoint. I wonder if that could work?

alexanderniebuhr: Interesting idea, however reading the RFC it sounds like the incoming `Request` inside the `fetch` handler of a provider-specific entrypoint is not the same as the one `app.ts` expects 🤔

matthewp: I stopped calling this "application entrypoint" because that's not accurate and can't be accurate. Astro can't get in front of adapters/platforms; those will always run first since they own the platform. If you create a worker.ts then that's runs first, regardless of what Astro does, that's `wrangler` who makes that decision.

But to address your concerns here, I think we can document that `astro/fetch` does work inside of `worker.ts` so you should be able to use that instead of `src/app.ts`. I think that does make sense. In which case we probably can deprecate `@astrojs/cloudflare/entrypoints/server` and tell people to just use `import { astro  } from 'astro/fetch';` instead since there's not much code between the two anyways.

matthewp:
@princemuel No, this doesn't change the fact that Astro is a JavaScript framework.

proposals/0056-advanced-routing.md
matthewp: @florian-lefebvre No that doesn't work, the Cloudflare vite plugin loads `worker.ts`, Astro doesn't. The layer is always going to be:

(host runtime) -> (host entrypoint) -> (adapter) -> astro/app -> src/app.ts

We have no control over the first 3 things in the chain. That's why I moved away from calling this an "entrypoint". It's an entrypoint inside of _Astro's_ request handling (or as close as it can be).

alexanderniebuhr: If `astro/fetch` works inside of `worker.ts` that is great and we should definetly document it that way. In that case the decision for users would be the following in my opionon: 

If they use a custom entrypoint they should not use `src/app.ts`, if they don't use a custom entrypoint they can use `src/app.ts`


> we probably can deprecate @astrojs/cloudflare/entrypoints/server

That is my hope, but is that true though? The [handler](https://github.com/withastro/astro/blob/c0db63dd9659317e503dfce2d8eb50d6d67be34d/packages/integrations/cloudflare/src/utils/handler.ts#L37) from `@astrojs/cloudflare/entrypoints/server` does do some special stuff with `SESSION` & `ASSETS` binding and custom headers though, would that also work with `astro/fetch`? We would need to find a way to get that logic inside the functions of `astro/fetch`?

florian-lefebvre: Is `app.ts` ever relevant for purely static sites? If not, that makes me wonder if this should be an adapter feature instead of core feature. Core could still provide `astro/(fetch|hono)` but it would only be used in adapter managed entrypoints

matthewp: Yes, redirects and i18n for example both work in static sites. Hono itself has a bunch as well, like sitemaps, and some development focused middleware.

Need to be careful to over-index on Cloudflare here, the Node adapter has no application entrypoint. Nor does the Vercel adapter.

matthewp: > That is my hope, but is that true though? The [handler](https://github.com/withastro/astro/blob/c0db63dd9659317e503dfce2d8eb50d6d67be34d/packages/integrations/cloudflare/src/utils/handler.ts#L37) from @astrojs/cloudflare/entrypoints/server does do some special stuff with SESSION & ASSETS binding and custom headers though, would that also work with astro/fetch? We would need to find a way to get that logic inside the functions of astro/fetch?

I'll look into this, but presumably we can replace this stuff with handlers, so if you're using Cloudflare you could do (as one example):

```ts
import { astro, FetchState } from 'astro/fetch';
import { cf } from '@astrojs/cloudflare/fetch';

export default {
  fetch(request: Request) {
    const state = new FetchState(request);
    await cf(state);
    return astro(state);
  }
}
```

proposals/0056-advanced-routing.md
matthewp: The user needs to call it because they created the FetchState instance. If you use `astro()` you don't because it does that internally. But if you use the lower-level APIs then we can't call it because we wouldn't know when/where to do so, any handler can be called in any order, only the user knows when the request is finished.

If I update the RFC to explain this, does that work?

proposals/0056-advanced-routing.md
matthewp: Added a note that pathname may include a `.html` extension during static builds.

proposals/0056-advanced-routing.md
matthewp: Updated the signatures with explicit return types. `provide` and `resolve` are synchronous, `finalizeAll` returns `Promise<void> | void`.

proposals/0056-advanced-routing.md
matthewp: Same pattern as `locals`. Added a "Typing Context Providers" section showing how users extend `App.Providers` in `env.d.ts` to type custom keys like `ctx.oauth`. Built-in providers like `session` are pre-declared by Astro.

proposals/0056-advanced-routing.md
matthewp: Yes, `middleware` sets `state.response` to the Response returned by `next`, and they're the same object. Added a clarification with a `response === state.response` comment in the code example.

proposals/0056-advanced-routing.md
matthewp: `cache` only wraps the response, so it doesn't need to forward state into `next`. The caller already has `state` in scope and can close over it. Added a note explaining the difference from `middleware` with an example.

proposals/0056-advanced-routing.md
matthewp: Added a note that platform edge middleware (Vercel, Netlify, etc.) runs before the adapter and is outside Astro's control. This proposal doesn't change how that layer works.

proposals/0056-advanced-routing.md
matthewp: Added a paragraph at the top of the Hono API section explaining the choice: fetch-native, largest middleware ecosystem among fetch frameworks, and lightweight with no server runtime. The `astro/hono` vs `@astrojs/hono` question is covered in the Alternatives section.

proposals/0056-advanced-routing.md
matthewp: Here's a PR that adds these: https://github.com/withastro/astro/pull/16729

So you could skip creating a `src/app.ts` and put this in your worker instead, basically equivalent:

```ts
import { astro, FetchState } from 'astro/fetch';
import { cf } from '@astrojs/cloudflare/fetch';

export default {
  async fetch(request: Request, env: Env, ctx: ExecutionContext) {
    const state = new FetchState(request);
    const asset = await cf(state, env, ctx);
    if (asset) return asset;
    return astro(state);
  }
}
```

proposals/0056-advanced-routing.md
ematipico: That's not what I meant; my question runs deeper. I'll explain. Up until now, Astro **was in charge** of the middleware, completely. When edge middleware is enabled, the entire file and business logic are placed on the edge, meaning the middleware **no longer runs within the Astro rendering pipeline**.

However, with this feature, the user is now in charge of where the middleware runs. So if the user decides to use `middleware` from `astro/fetch` or `astro/hono`, and then enables Netlify edge middleware, what happens? Does `middleware` become a no-op? Do we have two middlewares? 

It could be interesting to add a test in our adapters and see what the result is.

proposals/0056-advanced-routing.md
alexanderniebuhr: Looks nice. That would probably replace `import handler from "@astrojs/cloudflare/entrypoints/server";` in the future, which would be nice :)

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,499 @@
...
+
+The `src/app.ts` module will be opt-in. Without this file Astro will behave as it normally does.
+
+Initially this feature will be gated by an `experimental.advancedRouting` flag. In the next major version of Astro there will be no flag, meaning that `src/app.ts` becomes a special-file that users can't use for other purposes.
+
+To accommodate projects which already have a `app.ts` file, this will be configurable:
+
+```
+export default defineConfig({
+  fetchFile: 'fetch.ts'
```
florian-lefebvre: I think it should be called `fetchEntrypoint` to match other APIs, and allow:
- relative paths as strings
- package specifiers
- URLs

Also I don't know if `fetchX` is the right name but I don't have a better idea and I don't want to derail this comment ;)

matthewp: I don't think it should allow package specifiers. The primary goal of this RFC is to allow user to have control over routing. So the intent is for this to be a file in the project. It's an explicit non-goal of allowing integrations to "take over" this feature and I think allowing package specifiers would encourage that. The way for 3rd parties to participate is to export handlers/middleware that the user can user.

For this reason I want to keep it as `fetchFile`. It's similar to middleware, actions, and content config, all of which are name files (and not configurable at all).

matthewp:
I'm doing a call for consensus on this feature. As usually, there is a 3-day waiting period for people to raise concerns that this RFC should not be accepted, and then it will be assumed to pass.

cc @ascorbic @ematipico @florian-lefebvre from TSC.

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,497 @@
...
+The user creates the `FetchState` when using the `astro/fetch` API. The expected public surface includes:
+
+- `request` - The current `Request` object.
+- `response` - The `Response` produced by handlers, if any.
+- `routeData` - The matched route information.
+- `pathname` - The resolved pathname for the request. During a static build this may differ from the URL pathname — for example, it may include a `.html` extension to match the output file.
+- `locals` - The `App.Locals` object, available to middleware and endpoints via `ctx.locals` / `Astro.locals`.
+- `cookies` - The `AstroCookies` instance for reading/writing cookies.
+
+> **Note:** The exact shape of `FetchState` is still a work in progress and may change during the review and implementation process.
```
florian-lefebvre: Did it change?

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,497 @@
...
+      return response;
+    }
+
+    // Continue...
+    return pages(state);
+  }
+}
+```
+
+Note that using the `astro/hono` API the creation of FetchState is not necessary, since Hono has its own Context object; handlers can get/create a FetchState by inspecting HonoContext for a certain key (likely `astro.fetchState`).
```
florian-lefebvre: Is the key an implementation detail or should it be clearly documented?

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,497 @@
...
+
+### Context Providers
```
florian-lefebvre, line 200: I think this is fine for now but I don't really like the fact 3rd party handlers can extend the Astro global, there are `locals` for this exact usecase. So maybe the end game here is to remove this and move to locals, even for session and cache.

Again, not something to tackle now to avoid disruptions but to keep in mind for later.

florian-lefebvre, line 200: I also want to note that this means eg. `Astro.session` may not be defined anymore if the user doesn't include it as a middleware, which can break integrations assuming it exists. So I think this could be considered breaking. What do you think?

proposals/0056-advanced-routing.md
```diff
@@ -0,0 +1,497 @@
...
+
+```ts
+declare namespace App {
+  interface Providers {
+    oauth: import('./lib/oauth').OAuthSession;
+  }
+}
+```
+
+With this declaration, `ctx.oauth` and `Astro.oauth` are properly typed. Built-in providers like `session` are pre-declared by Astro's own type definitions, so users don't need to augment those themselves.
```
florian-lefebvre, line 245: That means `Astro.session` will be typed even if the corresponding middleware isn't set up. Is there anything that can be done to improve this?

proposals/0056-advanced-routing.md
matthewp: That wouldn't be breaking, the user didn't have the option of not including it before. It's still available by default, so `Astro.session` is defined.

matthewp: The idea of doing it this way was to make Astro more extensible, and eventually move things like Actions to a separate package, but still have the same API it has today. But as that is not happening now, I'd be fine with marking this API as internal and not supporting it for 3rd parties yet.

proposals/0056-advanced-routing.md
matthewp: I don't think so... unfortunately types need to be static. I don't think we could even use injectTypes here since we don't know its not used until much later. I think ensuring we have good runtime errors is probably the best we can do.

proposals/0056-advanced-routing.md
florian-lefebvre: In that case I think it'd be better to keep it internal

florian-lefebvre: I think it is breaking but maybe I misunderstand it:
- Before: `Astro.session` always defined
- After: If `fetchFile` is specified and the `session()` middleware is not used, then `Astro.session` is undefined. But the types say it is always defined

Is that correct?

(This is very much related to https://github.com/withastro/roadmap/pull/1344#discussion_r3362715698 probably)

ematipico (approved):

proposals/0056-advanced-routing.md
matthewp: Breaking is when you don't change your code and something stops working. You have to change your code to get something that was previously working to work again. The only thing breaking in this RFC is that `src/fetch.ts` becomes a reserved file name. Everything else is opt-in.

matthewp: Done in https://github.com/withastro/astro/pull/16980

matthewp: Actually maybe we are talking about different things. I'm thinking in terms of semver `major` breaking changes. Are you just pointing out that this breaks assumptions of integrations? 

If so, I could add a note to a section on Caveats, etc.

matthewp: I'm pretty sure that's what you're saying here. In that case I think what I'd rather do is just provide stubs so that Astro.session continues to exist even if the handler is not added.

florian-lefebvre: Yes I'm talking from the POV of the integrations. I think section in caveats + mention in docs + stub would be great if you don't mind!

matthewp: So the situation is not what we thought, `Astro.session` was always possibly undefined. It changed from an optional prop in 5.x: https://github.com/withastro/astro/blob/astro%405.18.0/packages/astro/src/types/public/context.ts

to `AstroSession | undefined` now: https://github.com/withastro/astro/blob/main/packages/astro/src/types/public/context.ts

So the only difference now is that there isn't a getter with a clear error message. I'll add that back. But there isn't a breaking change from the integration author perspective, they always needed to check for it existing.

matthewp: https://github.com/withastro/astro/pull/16982

proposals/0056-advanced-routing.md
matthewp: Updated the list in 489e0be.

proposals/0056-advanced-routing.md
matthewp: It's public, I'll open a PR to document, meanwhile here's the RFC change: 5520551

matthewp: Ah, you know what, it's not as easy as just documenting the key. it's created lazily on first access using the internal `getFetchState` function. We *could* expose this, but I'd rather wait until someone asks for it.

florian-lefebvre: If I understand correctly, if there's no public API for it that means a package that wants to have handlers like the official ones can only implement some for use in a simple Fetchable, not in an Hono middleware. I think it should be a documented public API so both flavors have the same capabilities

florian-lefebvre (changes requested):
I think https://github.com/withastro/roadmap/pull/1344#discussion_r3371525056 should be addressed before I approve

ascorbic (approved):

proposals/0056-advanced-routing.md
matthewp: Updated the RFC:

- Implementation: https://github.com/withastro/astro/pull/16998
- Docs: https://github.com/withastro/docs/pull/14021

florian-lefebvre (approved):

Issue #1358, opened 2026-05-08
astro dev --background


matthewp:
- Accepted Date: 2026-05-08
- Reference Issues/Discussions: #1308
- Author: @ascorbic
- Champion(s): @ascorbic @matthewp
- Implementation PR: 

# Summary

Add built-in background process management to `astro dev`, enabling AI coding agents to reliably start, monitor, query and stop the dev server.

# Background & Motivation

AI coding agents are increasingly the primary interface developers use to build Astro sites. These agents need to interact with the dev server constantly: starting it, checking if edits succeeded, reading errors and stopping it when done.

Today, this is painful. The dev server is designed for humans watching a terminal. Agents struggle with:

- **Long-running processes.** Most agents shell out, wait for exit, and read output. A dev server never exits. Agents either hang, or start it and lose track of it.
- **Readiness detection.** There is no reliable signal for "the server is ready to accept requests". Agents resort to polling with `curl`, sleeping for arbitrary durations, or parsing ANSI-coloured output for strings like "Local:".
- **HMR feedback.** After editing a file, agents have no way to know whether the HMR update succeeded or failed without making an HTTP request and hoping the error page is parseable. This is the single biggest failure mode observed in agent workflows.
- **Error parsing.** Build and runtime errors are formatted for human readability with colours, box-drawing characters and contextual snippets. Agents cannot reliably extract the file, line number and error type from this output.
- **Process lifecycle.** Agents frequently start duplicate dev servers, lose track of running instances, or leave zombie processes behind. There is no way to check if a server is already running for the current project.

My [`bgproc`](https://github.com/ascorbic/bgproc) project demonstrates that a thin process management layer with JSON output dramatically improves the agent experience with dev servers. Experiments by @FredKSchott demonstrated that bgproc reduced the time takes by an agent to triage an issue using `astro dev` by 81%, saved 13 tool calls and thousands of tokens. However, as a generic external wrapper, tools like bgproc are less discoverable and cannot provide framework-specific signals like HMR status, typed error codes or build state, and it relies upon heuristics to find ports that have been opened. Building this capability directly into Astro unlocks a significantly better experience.

# Goals

- Enable agents to start the dev server as a background process and receive structured JSON confirmation when it is ready, without polling or heuristics.
- Provide a machine-readable feedback loop for file changes: did the HMR update succeed or fail, and if it failed, what was the error?
[...]

PR #1362, opened 2026-05-11
Background dev server

matthewp:
# Summary

Background process management for `astro dev`, enabling AI coding agents to reliably start, monitor, query and stop the dev server.

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/rfc-0057-dev-background/proposals/0057-dev-background.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/1358)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/1308)

### DIFF
--- a/proposals/0057-dev-background.md
+++ b/proposals/0057-dev-background.md
@@ -267,14 +267,6 @@ The background server redirects both stdout and stderr to `.astro/dev.log`. The 
 
 The log file is truncated when a new background server starts.
 
-### Log rotation
-
-To prevent unbounded log growth, the log file is capped at a maximum size (default: 10 MB). When the file exceeds the limit, the oldest content is truncated to bring it back under the cap. The limit can be overridden with `--max-log-size`:
-
-```sh
-$ astro dev --background --max-log-size 50mb
-```
-
 The `logs` command reads this file and writes its contents to stdout.
 
 ## Integration with `src/app.ts` (Advanced Routing)
--- a/proposals/0057-dev-background.md
+++ b/proposals/0057-dev-background.md
@@ -1,5 +1,3 @@
-**If you have feedback and the feature is released as experimental, please leave it on the Stage 3 PR. Otherwise, comment on the Stage 2 issue (links below).**
-
 - Start Date: 2026-05-11
 - Reference Issues: https://github.com/withastro/roadmap/discussions/1308
 - Implementation PR: https://github.com/withastro/astro/pull/16610
--- a/proposals/0057-dev-background.md
+++ b/proposals/0057-dev-background.md
@@ -269,6 +269,21 @@ The log file is truncated when a new background server starts.
 
 The `logs` command reads this file and writes its contents to stdout.
 
+## Telemetry
+
+When an AI coding agent is detected, four optional fields are added to the existing `ASTRO_CLI_SESSION_STARTED` telemetry event:
+
+| Field | Type | Description |
+|---|---|---|
+| `isAgentic` | `boolean` | Whether the CLI session was run by an AI coding agent. |
+| `agentId` | `string` | Identifier of the detected agent (e.g. `"cursor-agent"`, `"claude-code"`). |
+| `agentName` | `string` | Display name of the detected agent (e.g. `"Cursor Agent"`, `"Claude Code"`). |
+| `agentType` | `string` | Type of agentic environment: `"agent"`, `"interactive"`, or `"hybrid"`. |
+
+These fields are collected for all CLI commands (not just `astro dev`) so the Astro team can understand which agents are commonly used and prioritize improvements. No new telemetry events are introduced. The fields are populated using `detectAgenticEnvironment()` from `am-i-vibing` and follow the same anonymous telemetry pipeline, respecting `astro telemetry disable`.
+
+The [astro.build/telemetry](https://astro.build/telemetry/) page must be updated to document these new fields before release.
+
 ## Integration with `src/app.ts` (Advanced Routing)
 
 If the project uses `src/app.ts` ([RFC 0056](https://github.com/withastro/roadmap/pull/1344)), the background server and health endpoint work identically. The `/_astro/status` middleware is registered at the Vite layer, before the user's fetch handler, so it is always available regardless of the user's routing configuration.
--- a/proposals/0057-dev-background.md
+++ b/proposals/0057-dev-background.md
@@ -328,8 +328,7 @@ An earlier design had all CLI commands print raw JSON to stdout (e.g. `{"pid": 1
 # Adoption strategy
 
 - **Fully opt-in.** Without `--background`, `astro dev` behaves exactly as it does today, unless an AI agent is detected, in which case background mode is enabled automatically.
-- **Not a breaking change.** The new subcommands and flags are additions to the CLI. The `/_astro/status` endpoint is a new route in the dev server that does not conflict with user routes (it uses the `/_astro/` prefix already reserved by Astro). The lock file adds duplicate server detection to foreground mode, but this is a guard against a common mistake, not a behavior change.
-- **Experimental phase.** Initially gated behind `--experimental-*` prefixed flags to gather feedback before stabilizing.
+- **Not a breaking change.** The new subcommands and flags are additions to the CLI. The `/_astro/status` endpoint is a new route in the dev server only — it does not exist in production builds. It does not conflict with user routes (it uses the `/_astro/` prefix already reserved by Astro). The lock file adds duplicate server detection to foreground mode, but this is a guard against a common mistake, not a behavior change.
 - **Agent instruction templates.** Once shipped, we can provide recommended agent instructions (e.g. for `.claude/commands/`, `.opencode/`, etc.) that use `astro dev` and rely on automatic agent detection for the background behavior.
 - **Documentation.** A new guide covering agent workflows with the background dev server, including common patterns for starting, checking status, and stopping the server.
 
--- a/proposals/0057-dev-background.md
+++ b/proposals/0057-dev-background.md
@@ -51,11 +51,8 @@ $ astro dev --background
 
 ### Automatic agent detection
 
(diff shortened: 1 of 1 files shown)


matthewp:
I'm doing a call for consensus on this change.

proposals/0057-dev-background.md
```diff
@@ -0,0 +1,334 @@
...
+
+## Logging
+
+The background server redirects both stdout and stderr to `.astro/dev.log`. This file contains the standard Astro dev server output, the same content that would appear in the terminal if run in foreground mode.
+
+The log file is truncated when a new background server starts.
+
+### Log rotation
+
+To prevent unbounded log growth, the log file is capped at a maximum size (default: 10 MB). When the file exceeds the limit, the oldest content is truncated to bring it back under the cap. The limit can be overridden with `--max-log-size`:
```
florian-lefebvre: I'm wondering how the default was chosen. Asking because 10 MB potentially seems like a lot for an agent context? Did you check how much tokens that can represent on average?

EDIT: since it is truncated on a new start, it probably doesn't matter that much

proposals/0057-dev-background.md
```diff
@@ -0,0 +1,334 @@
...
+
+# Adoption strategy
+
+- **Fully opt-in.** Without `--background`, `astro dev` behaves exactly as it does today, unless an AI agent is detected, in which case background mode is enabled automatically.
+- **Not a breaking change.** The new subcommands and flags are additions to the CLI. The `/_astro/status` endpoint is a new route in the dev server that does not conflict with user routes (it uses the `/_astro/` prefix already reserved by Astro). The lock file adds duplicate server detection to foreground mode, but this is a guard against a common mistake, not a behavior change.
+- **Experimental phase.** Initially gated behind `--experimental-*` prefixed flags to gather feedback before stabilizing.
+- **Agent instruction templates.** Once shipped, we can provide recommended agent instructions (e.g. for `.claude/commands/`, `.opencode/`, etc.) that use `astro dev` and rely on automatic agent detection for the background behavior.
+- **Documentation.** A new guide covering agent workflows with the background dev server, including common patterns for starting, checking status, and stopping the server.
+
+# Unresolved Questions
```
florian-lefebvre, line 335: Should unresolved questions be resolved before Stage 3 being merged?

proposals/0057-dev-background.md
```diff
@@ -0,0 +1,334 @@
...
+### Signal Handling
+
+The background server handles:
+
+- `SIGTERM`: Graceful shutdown. The dev server's `stop()` method is called, which removes the lock file and closes the HTTP server.
+- `SIGKILL`: Cannot be caught; the lock file will be stale. Detected and cleaned up by the next `status`, `stop`, or `astro dev` invocation.
+
+## Logging
+
+The background server redirects both stdout and stderr to `.astro/dev.log`. This file contains the standard Astro dev server output, the same content that would appear in the terminal if run in foreground mode.
```
florian-lefebvre: Something that is not clear to me is what the format of the log file is. Is that JSON, JSONL? Only the output format of the CLI is mentionned

proposals/0057-dev-background.md
matthewp: yeah when you start the dev server it truncates the file and starts fresh so you don't fill up the context with old logs.

10MB is just an arbitrary number that felt like more than you'd really need, but not so large that it causes any problems.

proposals/0057-dev-background.md
matthewp: Not necessarily, they represent things you've thought about but don't have an answer for. They are not for essential things, but are intended to let the reader know about things that are *not* covered. So a reader could/should consider them in their review.

proposals/0057-dev-background.md
matthewp: It uses JSON when run from within an agent, but otherwise follows the logger you're using. I'll update the text.

florian-lefebvre: Alright then my feedback is it should probably always use the same format, probably json. Wdyt?

matthewp: I updated this in https://github.com/withastro/roadmap/pull/1362/changes/cb6f47ef1b243e5ac99521ffde4fb7364637c9c1

proposals/0057-dev-background.md
```diff
@@ -0,0 +1,335 @@
+- Start Date: 2026-05-11
+- Reference Issues: https://github.com/withastro/roadmap/discussions/1308
+- Implementation PR: https://github.com/withastro/astro/pull/16610
+- Stage 2 Issue: https://github.com/withastro/roadmap/issues/1358
+- Stage 3 PR: https://github.com/withastro/roadmap/pull/1362
+
+# Summary
```
ematipico, line 7: I think having telemetry here is important to call out. We will have to update the website too.

proposals/0057-dev-background.md
```diff
@@ -0,0 +1,335 @@
...
+[astro] No dev server is running.
+```
+
+## Health endpoint
+
+All running dev server instances (foreground and background) expose a status endpoint:
+
+```sh
+$ curl -s http://localhost:4321/_astro/status
+{"ok": true}
```
ematipico, line 112: We should also show an example of unhealthy payload.

proposals/0057-dev-background.md
```diff
@@ -0,0 +1,335 @@
...
+# Summary
+
+Add built-in background process management to `astro dev`, enabling AI coding agents to reliably start, monitor, query and stop the dev server.
+
+# Example
+
+## Starting the dev server in the background
+
+```sh
+$ astro dev --background
```
ematipico, line 16: As per PR, these are now subcommands, not flags anymore.

proposals/0057-dev-background.md
```diff
@@ -0,0 +1,335 @@
...
+When `astro dev` starts in foreground mode, it checks for an existing lock file. If a live server is already running, it throws an error with the existing server's URL and PID, and suggests using `kill <pid>` or `astro dev --force` to replace it. This prevents agents (and humans) from accidentally starting duplicate dev servers.
+
+## Health Endpoint: `/_astro/status`
+
+All running dev server instances (foreground and background) expose a JSON status endpoint at `/_astro/status`. This is implemented as a Vite plugin (`astro:dev-status`) that registers a middleware handler.
+
+### Response Shape
+
+```json
+{"ok": true}
```
ematipico, line 226: Is `false` for when the endpoint isn't healthy?

proposals/0057-dev-background.md
```diff
@@ -0,0 +1,335 @@
...
+- **Custom logger:** Whatever format the user's custom logger implementation writes to stdout/stderr.
+
+The log file is truncated when a new background server starts.
+
+### Log rotation
+
+To prevent unbounded log growth, the log file is capped at a maximum size (default: 10 MB). When the file exceeds the limit, the oldest content is truncated to bring it back under the cap. The limit can be overridden with `--max-log-size`:
+
+```sh
+$ astro dev --background --max-log-size 50mb
```
ematipico: nit: not a big fan of words because then you'll have to do a mini parser... I would go for plain numbers. But no strong opinions, it's up to you, not a blocker at all

proposals/0057-dev-background.md
```diff
@@ -0,0 +1,335 @@
...
+
+## Raw JSON output
+
+An earlier design had all CLI commands print raw JSON to stdout (e.g. `{"pid": 12345, "url": "..."}`). This was changed to use Astro's logger so that human users see the normal Astro-styled output. The output format depends on the user's configured logger, so if a JSON logger is configured, these commands will naturally produce JSON. The JSON formatter functions are retained internally for programmatic use and testing.
+
+# Adoption strategy
+
+- **Fully opt-in.** Without `--background`, `astro dev` behaves exactly as it does today, unless an AI agent is detected, in which case background mode is enabled automatically.
+- **Not a breaking change.** The new subcommands and flags are additions to the CLI. The `/_astro/status` endpoint is a new route in the dev server that does not conflict with user routes (it uses the `/_astro/` prefix already reserved by Astro). The lock file adds duplicate server detection to foreground mode, but this is a guard against a common mistake, not a behavior change.
```
ematipico: Just for the sake of completeness: we should say that this endpoint doesn't exist in production. You never know!

proposals/0057-dev-background.md
```diff
@@ -0,0 +1,335 @@
...
+- **Experimental phase.** Initially gated behind `--experimental-*` prefixed flags to gather feedback before stabilizing.
```
ematipico: nit: not the case anymore

ematipico (commented):
A couple of things we should address imho:
- telemetry
- log rotation missing from PR 
- `--background` -> `background` (it's not a subcommand, given the PR)

matthewp:
> --background -> background (it's not a subcommand, given the PR)

Went back and forth on this and landed on what the stage 1 said, which was that `--background` is a flag on dev, but the other commands (stop, status, etc) are subcommands on dev. Updated the implementation PR to reflect.

proposals/0057-dev-background.md
matthewp: I added a sedction on telemetry.

proposals/0057-dev-background.md
matthewp: There isn't one :) This is normal in health endpoints though, if they are unhealthy then they won't respond. If the server is up, it's healthy. There's no concept of an unhealthy Astro dev server.

This endpoint just exists as a fallback in case normal fetches don't respond on a running background server. The agent will query this to determine if the process is dead, then restart it.

proposals/0057-dev-background.md
matthewp: Updated this, `--background` is still a flag, the others are subcommands.

proposals/0057-dev-background.md
matthewp: Got rid of the flag because it would mean adding a lot of complexity around log rotation, not something that dev servers normally do themselves anyways.

proposals/0057-dev-background.md
matthewp: added a note on this.

florian-lefebvre (approved):

ematipico (approved):

ascorbic (approved):
Approved!

Issue #1428, opened 2026-08-26
Support more validators


florian-lefebvre:
- Accepted Date: 2026-08-26
- Reference Issues/Discussions: #1272, #1116
- Author: @florian-lefebvre
- Champion(s): @florian-lefebvre
- Implementation PR: 

# Summary

Astro only supports zod as a validator. This proposal aims to decouple Astro from zod, allowing to use more validators at the same time.

# Background & Motivation

AFAIK Astro has been using zod for a long time, both internally and in its public APIS. This proposal only talks about the public APIs.

We've seen, while working on Astro 6, that being too tied to zod was a huge pain. It makes upgrading much harder and prevents users from using what they want.

I think we should make Astro agnostic in terms of schema validation. One problem is we can't only use Standard Schema (#1116) because we do things not covered by this spec. Affected areas are:

- The `astro/zod` module which re-exports astro's `zod` version
- Schemas in content collections
- Schemas in actions

# Goals

- Allow users to use other validators than zod
- Be backward compatible
- Ensure Astro's internal validator does not leak in the runtime

# Non-Goals

- Change Astro's internal validator for something else than zod

# Example

## `astro/zod`

If in the end we support several validation libraries, I think we should get rid of it. We could:

- Deprecate it in v7.x
- Ask users to install `zod` in their projects instead
- Update docs to show installing `zod`
- Remove in v8.0

## Content collections

The schema would need to implement https://standardschema.dev/schema and optionally https://standardschema.dev/json-schema.

Our current API is incompatible with supporting several validation libraries. We'd need to rethink `image()` and `reference()` entirely to allow this. An example of what it could look like:

```diff
import {
  defineCollection,
  reference,
+  image,
} from 'astro:content';
import { z } from 'zod';

const blogCollection = defineCollection({
-  schema: ({ image }) => z.object({
+  schema: (ctx) => z.object({
    title: z.string(),
-    cover: image(),
+    cover: z.string().transform(src => image(ctx, { src })),
    coverAlt: z.string(),
-	author: reference('authors'),
+	author: z.string().transform(id => reference('authors', id))
  }),
});

export const collections = {
  // ...
  blog: blogCollection,
};
```

This would solve 2 other complaints: schema transforms and validation on images in content collections (both lost with the content layer IIRC).

## Actions

JSON actions would just need to accept https://standardschema.dev/schema. Form actions are more complicated.

[...]

PR #1438, opened 2026-09-16
Support more validators

florian-lefebvre:
# Summary

Decouple Astro's public APIs from Zod: content collection schemas and action inputs accept any [Standard Schema](https://standardschema.dev) validator, and the Zod-only helpers they were built around (`image()`, `reference()`, form `input`, `astro/zod`) get validator-agnostic replacements.

```ts
// src/content.config.ts
import { defineCollection } from 'astro:content';
import { glob } from 'astro/loaders';
import * as v from 'valibot';

  loader: glob({ pattern: '**/*.md', base: './src/blog' }),
  schema: v.object({
    title: v.string(),
    draft: v.optional(v.boolean(), false),
  }),
});

export const collections = { blog };
```

`image()` and `reference()` become ordinary functions you call from inside a schema transform, so they work with any validator and their results can be validated further:

```diff
  import { defineCollection, reference } from 'astro:content';
+ import { image } from 'astro/content/image';
  import { z } from 'zod';

  const blog = defineCollection({
    loader: glob({ pattern: '**/*.md', base: './src/blog' }),
-   schema: ({ image }) => z.object({
+   schema: (context) => z.object({
-     cover: image(),
+     cover: z.string().transform((src) => image(context, { src })),
-     author: reference('authors'),
+     author: z.string().transform((id) => reference('authors', id)),
    }),
  });
```

Everything ships as a minor and stays backward compatible. The deprecations are removed in Astro 8.

# Links
- [Rendered RFC](https://github.com/withastro/roadmap/blob/rfc/validators/proposals/0063-validators.md)
- [Stage 2 Proposal](https://github.com/withastro/roadmap/issues/1428)
- [Stage 1 Proposal](https://github.com/withastro/roadmap/discussions/1272) / [Replace Zod with Standard Schema](https://github.com/withastro/roadmap/discussions/1116)
- [Implementation PR](https://github.com/withastro/astro/pull/17892)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


matthewp:
So, we had those list of things that were Zod-specific, like image(), and I can see that these things are covered in the RFC, but what i don't see is an explanation for how they are no longer Zod-specific? It reads as though they are not, but how is that possible?

florian-lefebvre:
@matthewp I added a section that explains this point in more details. TLDR: current APIs are deprecated (zod schema factory) and we have new ones that fit the new model (helper function that users can use in their schema)

